PR #115, opened 2021-09-03
Remove idemix

ale-linux:
We remove the local implementation of idemix and use the one
provided at https://github.com/IBM/idemix

Signed-off-by: Alessandro Sorniotti <aso@zurich.ibm.com>


ale-linux:
As recommended by @adecaro: we need to be sure that this doesn't break the token SDK. @KElkhiyaoui could you please have a look and see whether this breaks the token SDK? If so we can work together on a successful integration. Thx

PR #197, opened 2021-12-11
Make IsFinal work for real

yacovm:
The way how we implement broadcasting a transaction and then waiting for it to commit is inherently flawed:

1. We send it to the orderer
2. We connect to the orderer and start listening for blocks starting from the last block.

After (1) and before (2) the transaction might already enter a block and then (2) will miss it.
Moreover, the way (2) is implemented is connecting to the peer's delivery service and it just doesn't scale well when many transactions should be sent and waited for.

Implemented a pub-sub mechanism that is piggybacking on the already existing background task that pulls blocks and parses them. 

### DIFF
--- a/integration/fabric/iou/views/approver.go
+++ b/integration/fabric/iou/views/approver.go
@@ -6,6 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 package views
 
 import (
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
 	"github.com/pkg/errors"
 
 	"github.com/hyperledger-labs/fabric-smart-client/integration/fabric/iou/states"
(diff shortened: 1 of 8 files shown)


adecaro:
```The way how we implement broadcasting a transaction and then waiting for it to commit is inherently flawed:

We send it to the orderer
We connect to the orderer and start listening for blocks starting from the last block.```

This is no true!

adecaro:
The implementation of `IsFinal` follows already what you described: https://github.com/hyperledger-labs/fabric-smart-client/blob/e9102907a854c540a199590b92bae620a1791602/platform/fabric/core/generic/committer/committer.go#L119

yacovm:
The implementation of IsFinal that runs is [this](https://github.com/hyperledger-labs/fabric-smart-client/blob/e9102907a854c540a199590b92bae620a1791602/platform/fabric/core/generic/finality/fabric.go#L58-L100) one, not the one you pointed at

adecaro:
That's not true, look at https://github.com/hyperledger-labs/fabric-smart-client/blob/e9102907a854c540a199590b92bae620a1791602/platform/fabric/core/generic/channel.go#L125

The code you refer to is used by the committer when a transaction is marked as UNKNOWN or have dependencies.
- https://github.com/hyperledger-labs/fabric-smart-client/blob/e9102907a854c540a199590b92bae620a1791602/platform/fabric/core/generic/channel.go#L95
- https://github.com/hyperledger-labs/fabric-smart-client/blob/e9102907a854c540a199590b92bae620a1791602/platform/fabric/core/generic/committer/committer.go#L119

PR #204, opened 2021-12-15
Do not use QSCC to query block, fetch it from peer instead

yacovm:

### DIFF
--- a/platform/fabric/core/generic/delivery.go
+++ b/platform/fabric/core/generic/delivery.go
@@ -9,11 +9,15 @@ package generic
 import (
 	"context"
 
+	"github.com/hyperledger/fabric-protos-go/common"
+	pb "github.com/hyperledger/fabric-protos-go/peer"
+	"github.com/hyperledger/fabric/protoutil"
+
 	delivery2 "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/delivery"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
-	"github.com/hyperledger/fabric-protos-go/common"
-	"github.com/hyperledger/fabric/protoutil"
 )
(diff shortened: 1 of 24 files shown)


platform/fabric/core/generic/delivery.go
```diff
@@ -24,52 +22,32 @@ func (c *channel) Scan(ctx context.Context, txID string, callback driver.Deliver
...
+					return false, err
 				}
 
-				// The block is too big, download each transaction as needed
-				logger.Warnf("block [%d] too big to be downloaded, it contains [%d] txs",
-					filteredBlock.Number,
-					len(filteredBlock.FilteredTransactions))
-				block = nil
-			}
+				if common.HeaderType(chHdr.Type) != common.HeaderType_ENDORSER_TRANSACTION {
```
adecaro, line 32: if the transaction is not valid, it should be discarded

platform/fabric/core/generic/delivery/delivery.go
```diff
@@ -119,10 +119,14 @@ func (d *delivery) Run() error {
 			}
 
 			switch r := resp.Type.(type) {
-			case *pb.DeliverResponse_FilteredBlock:
-				logger.Debugf("delivery service [%s:%s], commit block [%d]", address, d.channel, r.FilteredBlock.Number)
+			case *pb.DeliverResponse_Block:
+				if r.Block == nil || r.Block.Data == nil || r.Block.Header == nil || r.Block.Metadata == nil {
+					return errors.Errorf("invalid block received")
```
adecaro, line 124: should we try to reconnect instead of interrupting delivery?

adecaro (changes requested):
Please, have a look at my comments.

adecaro (approved):
LGTM

PR #241, opened 2022-03-02
Add event service

mbrandenburger:
The event service enables service-to-service communication via pub/sub


platform/view/services/events/service.go
```diff
@@ -0,0 +1,53 @@
...
+func (s *Service) GetSubscriber() Subscriber {
+	return s.EventSystem
+}
+
+func (s *Service) GetPublisher() Publisher {
+	return s.EventSystem
+}
+
+func getService(sp view.ServiceProvider) (EventService, error) {
+	s, err := sp.GetService(&Service{})
```
adecaro, line 32: what about making &Service{} a global var that is initiazialed once and for all? This will avoid instantiating an empty object each time a getService is invoked.

adecaro (approved):
Just a minor, if you want to address.

platform/view/services/events/service.go
mbrandenburger: Pushed a fixup addressing your comment! Thanks Angelo

PR #242, opened 2022-03-02
larger keys in rw sets

adecaro:
Fabric does not set a limit on the length of the keys in a RWSet.
This PR is to match the same constraint.

### DIFF
--- a/platform/view/services/db/keys/utils.go
+++ b/platform/view/services/db/keys/utils.go
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,20}$")
-	keyRegexp = regexp.MustCompile("^[a-zA-Z0-9._\u0000=" + string(utf8.MaxRune) + "+/-]{1,200}$")
+	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,50}$")
+	keyRegexp = regexp.MustCompile("^[a-zA-Z0-9._\u0000=" + string(utf8.MaxRune) + "+/-]{1,4096}$")
 )
 
 const NamespaceSeparator = "\u0000"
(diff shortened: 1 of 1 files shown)


platform/view/services/db/keys/utils.go
```diff
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,20}$")
-	keyRegexp = regexp.MustCompile("^[a-zA-Z0-9._\u0000=" + string(utf8.MaxRune) + "+/-]{1,200}$")
+	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,50}$")
+	keyRegexp = regexp.MustCompile("^[a-zA-Z0-9._\u0000=" + string(utf8.MaxRune) + "+/-]{1,}$")
```
mbrandenburger, line 18: instead of `{1,}` you could also just have a `+`.

platform/view/services/db/keys/utils.go
```diff
@@ -14,8 +14,8 @@ import (
 )
 
 var (
-	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,20}$")
-	keyRegexp = regexp.MustCompile("^[a-zA-Z0-9._\u0000=" + string(utf8.MaxRune) + "+/-]{1,200}$")
+	nsRegexp  = regexp.MustCompile("^[a-zA-Z0-9._-]{1,50}$")
```
mbrandenburger, line 17: why still limiting the namespace to 50 char?

mbrandenburger (approved):
LGTM

platform/view/services/db/keys/utils.go
adecaro: oh, I didn't know.

platform/view/services/db/keys/utils.go
adecaro: Good point, I don't think there is a rule here. The limit there is not hurting for now :)

Issue #243, opened 2022-03-12
add ability to configure idemix cash size


adecaro:
This issue requires the following:
- Introduce `cacheSize` field for MSP. This field applies to idemix to pre-compute pseudonyms before they are used. 
- Remove useless log lines in the kvs and idemix pseudonyms generation

PR #245, opened 2022-03-14
add ability to configure idemix cash size #243

adecaro:

### DIFF
--- a/platform/fabric/core/provider.go
+++ b/platform/fabric/core/provider.go
@@ -149,13 +149,14 @@ func (p *fnsProvider) newFNS(network string) (driver.FabricNetworkService, error
 	}
 
 	// Local MSP Manager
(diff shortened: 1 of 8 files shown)


platform/fabric/core/provider.go
```diff
@@ -149,13 +149,14 @@ func (p *fnsProvider) newFNS(network string) (driver.FabricNetworkService, error
 	}
 
 	// Local MSP Manager
+	cacheSize := 500
```
mbrandenburger, line 152: is this some default cache size? can this value still change via config - is this happening in `msp/service.go` Line 405-408?

If so, I would rename this variable here to defaultCacheSize with a comment that this value may be overritten by using a config file

adecaro: very good, let me do that.

PR #247, opened 2022-03-15
fix p2p unit-tests flake #246

adecaro:

### DIFF
--- a/platform/view/services/comm/p2p_test.go
+++ b/platform/view/services/comm/p2p_test.go
@@ -8,6 +8,7 @@ package comm
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


platform/view/services/comm/p2p_test.go
```diff
@@ -105,7 +113,16 @@ func setupTwoNodesFromFiles(t *testing.T) (*P2PNode, *P2PNode, string, string) {
...
-	bootstrapNode, node := setupTwoNodes(t, bootstrapNodeID, bootstrapNodeEndpoint, nodeID, nodeEndpoint, &PrivateKeyFromFile{bootstrapNodeSK}, &PrivateKeyFromFile{nodeSK})
+	// sometimes setupTwoNodes can fail due to timeouts, give it a second chance
+	var err error
+	var bootstrapNode, node *P2PNode
+	for i := 0; i < 3; i++ {
+		bootstrapNode, node, err = setupTwoNodes(t, bootstrapNodeID, bootstrapNodeEndpoint, nodeID, nodeEndpoint, &PrivateKeyFromFile{bootstrapNodeSK}, &PrivateKeyFromFile{nodeSK})
+		if err == nil {
+			break
+		}
+	}
```
mbrandenburger, line 124: I suggest a dedicated Retry function, which can be used in many places.

This is cleaner than just running the code in the loop adding a comment that we want to retry ... 

For instance:
```go
func retry(attempts int, sleep time.Duration, f func() error) (err error) {
	for i := 0; i < attempts; i++ {
		if i > 0 {
			time.Sleep(sleep)
			sleep *= 2
		}

		err = f()
		if err == nil {
			return nil
		}
	}
	return fmt.Errorf("no luck after %d attempts: last error: %v", attempts, err)
}
```

```go
var bootstrapNode, node *P2PNode
retry(3, 0 *time.Seconds, func() error {
  bootstrapNode, node, err = setupTwoNodes(t, bootstrapNodeID, bootstrapNodeEndpoint, nodeID, nodeEndpoint, &PrivateKeyFromFile{bootstrapNodeSK}, &PrivateKeyFromFile{nodeSK})
  if err != nil {
    return err 
  }
})
```

adecaro: beautiful, let me do that

adecaro: done

platform/view/services/comm/p2p_test.go
```diff
@@ -260,3 +274,18 @@ func SessionsForMPCTestRound(t *testing.T, bootstrapNode, node *P2PNode, bootstr
...
+			sleep *= 2
+		}
+
+		err = f()
+		if err == nil {
+			return nil
+		}
+	}
+	return fmt.Errorf("no luck after %d attempts: last error: %v", attempts, err)
+}
```
mbrandenburger, line 291: I suggest to put this function somewhere else, so we can easily reuse it.

adecaro: okay

adecaro: done :)

mbrandenburger (approved):
Thx Angelo. LGTM now

PR #250, opened 2022-03-17
remove debug.Stack logging where not needed

adecaro:


mbrandenburger (commented):
This is good. What is your preferred way of showing the stack trace? There are a few more places in the code where we log the stack trace on warnings. I would prefer to show the stack trace only if something crashes.

adecaro:
done ;)

PR #253, opened 2022-03-23
Attach SGX devices to FPC container

mbrandenburger:
DCAP based attestation requires to attach the /dev/sgx/provision device
to the FPC docker container. This commit allows to attach multiple
devices to the container. The user is responsible to detect the devices
and provide them via the WithSGXDevicesPaths option.


adecaro (approved):
LGTM

PR #262, opened 2022-04-05
Remove wrong rlock release

mbrandenburger:
This PR addresses a bug in the KVS, where a RLock release is called on a unlocked lock.


adecaro (approved):
LGTM

PR #270, opened 2022-05-03
Fix sig deserializer data race

mffrench:

### DIFF
--- a/platform/view/core/sig/deserializer.go
+++ b/platform/view/core/sig/deserializer.go
@@ -9,6 +9,7 @@ package sig
 import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap/zapcore"
+	"sync"
 
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/driver"
 )
@@ -20,8 +21,9 @@ type Deserializer interface {
 }
 
 type deserializer struct {
-	sp            driver.ServiceProvider
-	deserializers []Deserializer
+	sp                 driver.ServiceProvider
+	deserializersMutex sync.RWMutex
+	deserializers      []Deserializer
 }
 
 func NewMultiplexDeserializer(sp driver.ServiceProvider) (*deserializer, error) {
@@ -32,12 +34,16 @@ func NewMultiplexDeserializer(sp driver.ServiceProvider) (*deserializer, error) 
 }
 
 func (d *deserializer) AddDeserializer(newD Deserializer) {
+	d.deserializersMutex.Lock()
 	d.deserializers = append(d.deserializers, newD)
+	d.deserializersMutex.Unlock()
 }
 
 func (d *deserializer) DeserializeVerifier(raw []byte) (driver.Verifier, error) {
 	var errs []error
-	for _, des := range d.deserializers {
+
+	copyDeserial := d.threadSafeCopyDeserializers()
+	for _, des := range copyDeserial {
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
 			logger.Debugf("trying deserialization with [%v]", des)
 		}
@@ -60,7 +66,9 @@ func (d *deserializer) DeserializeVerifier(raw []byte) (driver.Verifier, error) 
 
 func (d *deserializer) DeserializeSigner(raw []byte) (driver.Signer, error) {
 	var errs []error
-	for _, des := range d.deserializers {
+
+	copyDeserial := d.threadSafeCopyDeserializers()
+	for _, des := range copyDeserial {
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
 			logger.Debugf("trying signer deserialization with [%s]", des)
 		}
@@ -83,7 +91,9 @@ func (d *deserializer) DeserializeSigner(raw []byte) (driver.Signer, error) {
 
 func (d *deserializer) Info(raw []byte, auditInfo []byte) (string, error) {
 	var errs []error
-	for _, des := range d.deserializers {
+
+	copyDeserial := d.threadSafeCopyDeserializers()
+	for _, des := range copyDeserial {
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
 			logger.Debugf("trying info deserialization with [%v]", des)
 		}
@@ -102,4 +112,12 @@ func (d *deserializer) Info(raw []byte, auditInfo []byte) (string, error) {
 	}
 
 	return "", errors.Errorf("failed info deserialization [%v]", errs)
+}
+
+func (d *deserializer) threadSafeCopyDeserializers() []Deserializer {
+	d.deserializersMutex.RLock()
+	res := make([]Deserializer, len(d.deserializers))
+	copy(res, d.deserializers)
+	d.deserializersMutex.RUnlock()
+	return res
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /.idea/
 .DS_Store
 .build
+*.iml


platform/view/core/sig/deserializer.go
```diff
@@ -32,12 +34,22 @@ func NewMultiplexDeserializer(sp driver.ServiceProvider) (*deserializer, error)
...
 
 func (d *deserializer) AddDeserializer(newD Deserializer) {
+	d.deserializersMutex.Lock()
 	d.deserializers = append(d.deserializers, newD)
+	d.deserializersMutex.Unlock()
 }
 
 func (d *deserializer) DeserializeVerifier(raw []byte) (driver.Verifier, error) {
 	var errs []error
+
```
adecaro, line 44: why don't we use a read lock here? I expect this function to be called many times. A read lock would be faster given that `AddDeserializer` in comparison is called fewer times.

mffrench: 👍

platform/view/core/sig/deserializer.go
```diff
@@ -103,3 +113,13 @@ func (d *deserializer) Info(raw []byte, auditInfo []byte) (string, error) {
 
 	return "", errors.Errorf("failed info deserialization [%v]", errs)
 }
+
+func (d *deserializer) threadSafeCopyDeserializers() []Deserializer {
```
adecaro, line 117: Doesn't this waste memory? We call DeserializeVerifier very often? Would it be better to read lock at each function without copying?

adecaro (changes requested):
Still a doubt

platform/view/core/sig/deserializer.go
yacovm: I think the most performant solution uses an **atomic.Value**:

**AddDeserializer**: Appends the slice _under a lock._ 

Any other method, only reads the slice, so it just loads the **atomic.Value**.

adecaro: why is this faster than a mutex? Do you have numbers?

mffrench: I defined this func to reduce locking time under assumption golang garbage collector will do the work on the unused allocated struct... 

@adecaro @yacovm : will look on the atomic value method to get an idea. Let me know your final decision

That being said, that's not the better solution I like, I would prefer have a dedicated thread safe slice tooling : https://dnaeon.github.io/concurrent-maps-and-slices-in-go/

yacovm: How else are you going to store your deserializers?

adecaro: sorry, I missed the last point. Then, I'm okay either way.
To recap, no copying and either RW mutex or atomic value (as per Yacov suggestion)

yacovm: [This](https://github.com/yacovm/fabric-smart-client/commit/e5e6f7b7ca4126b7bdc92afc7f6d47640d97dc15) is what I had in mind.

adecaro: @yacovm , shall we then stick to non-cloning?

yacovm: Do what you want, it doesn't matter to me

mffrench: @yacovm : I will test/validate your proposal as it seems to me the best. I updated this PR with it.

mffrench: @adecaro @yacovm : my tests seems to be happy with that fix, however fsc CI is not : 

```
/home/runner/work/fabric-smart-client/fabric-smart-client/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/core/sig/deserializer.go
/home/runner/work/fabric-smart-client/fabric-smart-client/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/core/sig/deserializer.go
Error: /bin/sh: 1: test: /home/runner/work/fabric-smart-client/fabric-smart-client/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/core/sig/deserializer.go: unexpected operator
make: *** [Makefile:3: checks] Error 1
Fix formatting issues
```

I'm not sure to understand the issue there ... Will dig a bit. Let share some hints if any and I will fix.

mffrench: FYI: CI retry with fmt on `platform/view/core/sig/deserializer.go`

mffrench: Our asset exchange integration test with atomic.Value solution seems to be a bit slower (?!) than the one with copy solution.

With atomic.Value solution, I get `pseudonym is invalid` on claim (meaning we claim after timeout defined as 120sec) between 50% and 75%. It goes well 100% time with copy solution. 

@yacovm : I'm thinking that could comes from a lock time higher in atomic.Value solution than in the copy solution. Would that be OK if I reduce that lock time like this (my understanding is we don't need to protect Load and Store func from datarace) : 

```
func (d *deserializer) AddDeserializer(newD Deserializer) {
	var deserializers []Deserializer
	prev := d.deserializers.Load()
	if prev != nil {
		deserializers = prev.([]Deserializer)
	}

        d.lock.Lock()
	deserializers = append(deserializers, newD)
	d.lock.Unlock()
	d.deserializers.Store(deserializers)
}
```

mffrench: So while applying the fix I saw the link you provided @yacovm was not the final one and so the final function I'm testing is that one : 

```
func (d *deserializer) AddDeserializer(newD Deserializer) {
	var deserializers []Deserializer
	prev := d.deserializers.Load()
	if prev != nil {
		deserializers = prev.([]Deserializer)
	}

	d.lock.Lock()
	res := make([]Deserializer, len(deserializers)+1)
	copy(res, deserializers)
	res[len(res)-1] = newD
	d.lock.Unlock()

	d.deserializers.Store(res)
}
```

Let me know if you see any issue with that.

Note that the copy I see in that `AddDeserializer` version answer my question about the potential datarace on the slice.

yacovm: > Looking that :
> 
>     * atomic.Value solution still require a copy to fix data race on the slice

The thing is - we should only copy when we mutate, but not when we read. 

>     * atomic.Value solution make fails our interop asset exchange test with a 120sec timeout cause claim is done after the timeout.


If it fails the test consistently then there is probably something wrong with it. Can you point out at what is wrong with it?

adecaro:
LGTM

platform/view/core/sig/deserializer.go
mffrench: `The thing is - we should only copy when we mutate, but not when we read.`
I would be interested to know why ? In the end seems to me good to avoid return back slice pointer which could be modified after the read - which may lead to other datarace...

`If it fails the test consistently then there is probably something wrong with it. Can you point out at what is wrong with it?`
There is no problem with the test itself but we have issue with overall performance since our last migration which requires me to increase our asset exchange/transfer from 30sec to 120sec to make them work (claim is done after the timeout and then our balance checks fails). When testing with atomic.Value solution instead of sync.Mutex solution I introduce another latency which make the claims comes after 120sec 50%.

mffrench: `The thing is - we should only copy when we mutate, but not when we read.`
I think in that scenario I understand you would prefer to copy on write and not on read : to reduce read latency ? That is a point I was thinking when I try to understand why I see `atomic.Value` solution slower from my test perspective : are we executing more `AddDeserialized` than read operation ? 

-> I will investigate that point later with the dev branch I kept on my fork

mffrench: @adecaro : do we `AddDeserializer` each time we create idemix pseudo id ?

yacovm: > `The thing is - we should only copy when we mutate, but not when we read.` I would be interested to know why ? In the end seems to me good to avoid return back slice pointer which could be modified after the read - which may lead to other datarace...

If a deserializer can be mutated we have much bigger problems. It should be immutable. 

> 
> `If it fails the test consistently then there is probably something wrong with it. Can you point out at what is wrong with it?` There is no problem with the test itself but we have issue with overall performance since our last migration which requires me to increase our asset exchange/transfer from 30sec to 120sec to make them work (claim is done after the timeout and then our balance checks fails). When testing with atomic.Value solution instead of sync.Mutex solution I introduce another latency which make the claims comes after 120sec 50%.

I meant something wrong with the solution I proposed, not in the test. 

> When testing with atomic.Value solution instead of sync.Mutex solution I introduce another latency which make the claims comes after 120sec 50%.

But why?

Issue #273, opened 2022-05-05
increase namespace length limit in rw sets


saoussen:
update limit in func ValidateNs from 50 to 64

PR #274, opened 2022-05-05
increase namespace length limit in rw sets #273

saoussen:
Signed-off-by: Saoussen MAROUANI <saoussen.marouani@ibm.com>


adecaro (approved):
LGTM

PR #280, opened 2022-05-11
fabric-sdk: add keep alive config support #279

adecaro:

### DIFF
--- a/integration/nwo/fabric/topology/core_template.go
+++ b/integration/nwo/fabric/topology/core_template.go
@@ -282,6 +282,14 @@ fabric:
         files:
         - {{ .PeerLocalTLSDir Peer }}/ca.crt
       rootCertFile: {{ .CACertsBundlePath }}
+      keepalive:
+       client:
+         interval: 60s
+         timeout: 600s
+       server:
+         interval: 60s
+         timeout: 600s
+         minInterval: 60s 
     orderers: {{ range Orderers }}
       - address: {{ OrdererAddress . "Listen" }}
         connectionTimeout: 10s
--- a/platform/fabric/core/generic/config/config.go
+++ b/platform/fabric/core/generic/config/config.go
@@ -87,6 +87,26 @@ func (c *Config) TLSRootCertFile() string {
 	return c.configService.GetString("fabric." + c.prefix + "tls.rootCertFile")
 }
 
+func (c *Config) KeepAliveClientInterval() time.Duration {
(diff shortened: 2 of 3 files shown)


platform/fabric/core/generic/config/config.go
```diff
@@ -87,6 +87,26 @@ func (c *Config) TLSRootCertFile() string {
...
+	return c.configService.GetDuration("fabric." + c.prefix + "tls.keepalive.server.interval")
+}
+
+func (c *Config) KeepAliveServerTimeout() time.Duration {
+	return c.configService.GetDuration("fabric." + c.prefix + "tls.keepalive.server.timeout")
+}
+
+func (c *Config) KeepAliveServerMinInterval() time.Duration {
+	return c.configService.GetDuration("fabric." + c.prefix + "tls.keepalive.server.minInterval")
+}
```
mbrandenburger, line 108: What would happen if the configuration does not contain the entry? Would `c.configService.GetDuration` just panic or do we have default values somewhere? Maybe directly from `core_template.go`?

integration/nwo/fabric/topology/core_template.go
```diff
@@ -282,6 +282,14 @@ fabric:
...
         - {{ .PeerLocalTLSDir Peer }}/ca.crt
       rootCertFile: {{ .CACertsBundlePath }}
+      keepalive:
+       client:
+         interval: 60s
+         timeout: 600s
+       server:
+         interval: 60s
+         timeout: 600s
+         minInterval: 60s
```
mbrandenburger, line 292: Just thinking ... how would a FSC user be able to change these settings without touch `core_template.go`?

adecaro: just changing the configuration file, no?

platform/fabric/core/generic/config/config.go
adecaro: good point, let's test :)

integration/nwo/fabric/topology/core_template.go
adecaro: or we need explicit support on the topology. Let's wait for a first request on this 😅

platform/fabric/core/generic/config/config.go
adecaro: actually, it just return the default value the type of variable can get. Duration gets 0, being an int64.

Issue #276, opened 2022-05-09
Datarace in FSC DB memory driver


mffrench:
Hello,

here's a new datarace in the FSC DB memory driver...

```
==================
WARNING: DATA RACE
Read at 0x00c00228e390 by goroutine 1512:
  github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/memory.(*database).mapForNamespaceForReading()
      /go/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/memory/db.go:140 +0x77
  github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/memory.(*database).GetStateRangeScanIterator()
      /go/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/memory/db.go:158 +0x72
  github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/unversioned.(*Unversioned).GetStateRangeScanIterator()
      /go/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/services/db/driver/unversioned/unversioned.go:55 +0xa6
  github.com/hyperledger-labs/fabric-smart-client/platform/view/services/kvs.(*KVS).GetByPartialCompositeID()
      /go/src/github.com/hyperledger-labs/fabric-smart-client/platform/view/services/kvs/kvs.go:211 +0x1ea
  github.ibm.com/fabric-security-research/fabric-token-sdk-plus/token/services/interop/transfer.(*kvsStore).GetByPartialCompositeID()
      /go/src/github.ibm.com/fabric-security-research/fabric-token-sdk-plus/token/services/interop/transfer/store.go:52 +0xe4
  github.ibm.com/fabric-security-research/fabric-token-sdk-plus/token/services/interop/transfer.(*pledgeVault).PledgeByTokenID()
      /go/src/github.ibm.com/fabric-security-research/fabric-token-sdk-plus/token/services/interop/transfer/store.go:91 +0x132
  git.cloud.ibm.com/preons/go.fabflow/pkg/broker.(*FFBroker).interopTransferClaim()
      /go/src/git.cloud.ibm.com/preons/go.fabflow/pkg/broker/token_interop_transfer_claim.go:92 +0x259
  git.cloud.ibm.com/preons/go.fabflow/pkg/broker.(*FFBroker).receivePledgeRequest()
[...]

PR #282, opened 2022-05-24
fix db memory data race

mffrench:
see ticket https://github.com/hyperledger-labs/fabric-smart-client/issues/276

### DIFF
--- a/platform/view/services/db/driver/memory/db.go
+++ b/platform/view/services/db/driver/memory/db.go
@@ -137,10 +137,14 @@ func (db *database) Discard() error {
 }
 
 func (db *database) mapForNamespaceForReading(ns string, add bool) map[string]*versionedValue {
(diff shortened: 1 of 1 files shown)


adecaro:
@mffrench , please can you add a unit-test to check that the race condition is gone? Thanks :)

mffrench:
@adecaro : would love to add such tests but need to rush on our app stack now :/ Currently only test for me to validate we don't get that datarace again is on our app stack...

Regarding tests validating no more datarace, we should handle all func available from db interface, run them in appropriate way in several go routine and test it against memory driver or other one and ...

adecaro:
what I don't get is why there was a race condition before. Can you explain the rational of your changes, please?

adecaro (commented):
LGTM for now, but need to be revised soon.

PR #289, opened 2022-06-07
Upgrade libp2p

mbrandenburger:
- upgrade libp2p v0.20.1
- integrate api changes
- fix issue with streams mutex


platform/view/services/comm/builder.go
```diff
@@ -33,16 +33,18 @@ func newHost(ListenAddress string, keyDispenser PrivateKeyDispenser) (*P2PNode,
...
-	)
+		libp2p.ForceReachabilityPublic(),
+	}
+
+	host, err := libp2p.New(opts...)
 	if err != nil {
 		return nil, err
 	}
 
+	ctx := context.Background()
```
adecaro, line 47: Shall we pass the context to give the caller of the function to have a chance to interrupt the process? If the passed context is nil, then context.Background() is used.

platform/view/services/comm/builder.go
```diff
@@ -160,7 +162,10 @@ func (p *P2PNode) startFinder() {
 }
 
 func (p *P2PNode) start() {
-	discovery.Advertise(context.Background(), p.finder, rendezVousString)
+	_, err := p.finder.Advertise(context.Background(), rendezVousString)
```
adecaro, line 165: Shall we pass the context to give the caller of the function to have a chance to interrupt the process? If the passed context is nil, then context.Background() is used.

platform/view/services/comm/p2p.go
```diff
@@ -242,24 +243,29 @@ func (s *streamHandler) handleIncoming() {
...
+			for i, thisSH := range s.node.streams[remotePeerID] {
 				if thisSH == s {
-					s.node.streams[s.stream.Conn().RemotePeer()] = append(s.node.streams[s.stream.Conn().RemotePeer()][:i], s.node.streams[s.stream.Conn().RemotePeer()][i+1:]...)
+					s.node.streams[remotePeerID] = append(s.node.streams[remotePeerID][:i], s.node.streams[remotePeerID][i+1:]...)
 					s.wg.Done()
+					s.node.streamsMutex.Unlock()
 					return
 				}
 			}
+			s.node.streamsMutex.Unlock()
```
adecaro, line 266: this unlock is not invoked is line 263 (the return) is reached, no?

adecaro (changes requested):

platform/view/services/comm/builder.go
mbrandenburger: good idea - let's cover this in a separate PR. There are a few improvements we could add to the comm services.

platform/view/services/comm/p2p.go
mbrandenburger: correct, if we return before, this line is not reached. At this point the Unlock is kind of pointless as we panic anyway.

However, as @mffrench proposed to reduce panics whenever possible, the `panic` should replaces with better conflict resolution.

platform/view/services/comm/builder.go
adecaro: Okay!

platform/view/services/comm/p2p.go
adecaro: Sorry, my fault. I didn't see the lock before return. All good.

PR #291, opened 2022-06-08
CI go1.17

adecaro:


mbrandenburger (approved):
LGTM! 

Another approach we could consider for future updates is to get the go version from the `go.mod`. Also, we might want to run with our "supported target" go version and maybe also the most recent go version. But for now let's move forward with this.

PR #292, opened 2022-06-08
Improve identity cache default

mffrench:

### DIFF
--- a/platform/fabric/core/generic/config/config.go
+++ b/platform/fabric/core/generic/config/config.go
@@ -15,6 +15,8 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc"
 )
 
+const DefaultMSPCacheSize = 3
+
 // configService models a configuration registry
 type configService interface {
 	// GetString returns the value associated with the key as a string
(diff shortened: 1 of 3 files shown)


platform/fabric/core/generic/config/config.go
```diff
@@ -198,14 +198,16 @@ func (c *Config) GetPath(key string) string {
 	return c.configService.GetPath("fabric." + c.prefix + key)
 }
 
-func (c *Config) MSPCacheSize(defaultValue int) int {
+const DefaultCacheSize = 3
```
adecaro, line 201: Please, move this const to the top of the file.
Shall we call it DefaultMSPCacheSize?
Please, add also a comment describing the constant.

platform/fabric/core/generic/msp/cache.go
```diff
@@ -35,7 +35,6 @@ func NewIdentityCache(backed IdentityCacheBackendFunc, size int) *IdentityCache
 		timeout: time.Millisecond * 100,
 	}
 	go ci.run()
-
```
adecaro, line 38: do we need to remove this file?

adecaro (changes requested):

platform/fabric/core/generic/msp/cache.go
mffrench: you mean this line I guess... don't think so ;)

platform/fabric/core/generic/config/config.go
mffrench: OK ! making the change

adecaro (approved):

Issue #293, opened 2022-06-16
extract IPs from discovery


adecaro:
Currently, the fabric-sdk uses discovery to retrieve the public keys of the endorsers. Then, the fabric-sdk tries to lookup the corresponding IPs looking at the configured `resolver`. This puts additional burden at configuration time.

This github issue asks to use discovery to retrieve also the IPs.

PR #294, opened 2022-06-16
extract IPs from discovery #293

adecaro:
This enhancement is tested here: https://github.com/hyperledger-labs/fabric-token-sdk/pull/260

### DIFF
--- a/integration/nwo/fabric/network/resolver.go
+++ b/integration/nwo/fabric/network/resolver.go
@@ -63,7 +63,6 @@ func (n *Network) GenerateResolverMap() {
 				//P2PPort:    fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[P2PPort]),
 			}
 			path = n.PeerLocalMSPIdentityCert(peer)
-			continue
 		}
 
 		var aliases []string
--- a/integration/nwo/fabric/network/resolver.go
+++ b/integration/nwo/fabric/network/resolver.go
@@ -63,6 +63,9 @@ func (n *Network) GenerateResolverMap() {
 				//P2PPort:    fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[P2PPort]),
 			}
 			path = n.PeerLocalMSPIdentityCert(peer)
+
+			// don't add fabric peers, they will be discovered
+			continue
 		}
 
 		var aliases []string
--- a/integration/nwo/fabric/network/resolver.go
+++ b/integration/nwo/fabric/network/resolver.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package network
 
 import (
-	"fmt"
 	"path/filepath"
 
 	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api"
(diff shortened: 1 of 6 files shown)


integration/nwo/fabric/network/resolver.go
```diff
@@ -57,12 +56,15 @@ func (n *Network) GenerateResolverMap() {
...
-				//P2PPort:    fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[P2PPort]),
-			}
-			path = n.PeerLocalMSPIdentityCert(peer)
+			//addresses = map[api.PortName]string{
+			//	//ViewPort:   fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[ListenPort]),
+			//	ListenPort: fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[ListenPort]),
+			//	//P2PPort:    fmt.Sprintf("127.0.0.1:%d", n.Context.PortsByPeerID(n.Prefix, peer.ID())[P2PPort]),
+			//}
+			//path = n.PeerLocalMSPIdentityCert(peer)
+
```
mbrandenburger, line 65: can this be removed?

adecaro: actually no because the FPC Echo fails. So, I will restore it for now until I found a way to discover the info of the peer endpoints hosting the enclave.

platform/fabric/core/generic/chaincode/discovery.go
```diff
@@ -93,9 +93,29 @@ func (d *Discovery) Call() ([]view.Identity, error) {
...
-		endorserIdentities = append(endorserIdentities, e.Identity)
+	configResult, err := cr.Config()
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed getting config for [%s]", key)
+	}
+
+	var discoveredEndorsers []driver.DiscoveredPeer
+	for _, peer := range endorsers {
+		// extract peer info
+		member := peer.AliveMessage.GetAliveMsg().Membership
```
yacovm, line 104: GetAliveMessage can be nil

adecaro: Under which conditions?

yacovm: Well you receive this from a peer, what if the peer is faulty and doesn't put the message there? 
Yes, continue to other peers.

platform/fabric/core/generic/chaincode/discovery.go
```diff
@@ -127,24 +147,18 @@ func (d *Discovery) send() (discovery.Response, error) {
...
-	//if len(d.ImplicitCollections) > 0 {
-	//	for _, collection := range d.ImplicitCollections {
-	//		collectionNames = append(collectionNames, fmt.Sprintf("_implicit_org_%s", collection))
-	//	}
-	//}
-
+	// New discovery request for:
+	// - endorsers,
+	// - config, and
+	// - local peers (for alive message)
```
yacovm, line 153: why do you use local peer query? You need to be a local admin to issue it by default, otherwise you get access denied. 

You can do a peer membership query that is channel scoped instead.

platform/fabric/core/generic/chaincode/discovery.go
adecaro: all right

platform/fabric/core/generic/chaincode/discovery.go
adecaro: Let me make sure I understand. Here I need to get the endorsers of a given chaincode and I also want their IP addresses. What's the minimum I need?

yacovm: The endorser query gives you the IP as well. You don't need anything else. 
If you wish the TLS root CAs you also need the config query which you use.

adecaro: perfect, I'll update the code

yacovm (approved):

PR #300, opened 2022-06-27
Enforce GOPATH

mbrandenburger:
- Update readme with link to gopath doc
- Enforce GOPATH is set when using integration test suite


mbrandenburger:
In the long term we should get rid of the GOPATH requirement. However, this is an attempt to enforce GOPATH is set.

adecaro (approved):
LGTM

PR #308, opened 2022-07-07
Refactor nwo docker support

mbrandenburger:
This PR changes the use of docker in the NWO. The NWO uses docker to launch some components, such as Orion, Weaver, Monitoring. This refactoring has two goals: 
First, remove the use of docker (use of docker client) during platform `init` function and move the corresponding logic to `postRun`. This allows a docker-free use of the fsccli artifactgen. 
Second, reduce redundant and cluttered code. Docker related code is now located in a `docker.go` file per platform.


integration/nwo/common/docker/utils.go
```diff
@@ -0,0 +1,185 @@
...
+			return nil, err
+		}
+
+		singleInstance = &Docker{dockerClient}
+	}
+
+	return singleInstance, nil
+}
+
+func (d *Docker) CheckImagesExist(requiredImages ...string) error {
```
adecaro, line 45: Please, add a comment to any new function introduced.

integration/nwo/common/docker/utils.go
```diff
@@ -0,0 +1,185 @@
...
+type Docker struct {
+	Client *docker.Client
+}
+
+var singleInstance *Docker
+var lock = &sync.Mutex{}
+
+func GetInstance() (*Docker, error) {
+
+	if singleInstance == nil {
```
adecaro, line 30: this check should happen when the lock has been acquired already, otherwise you can have more than one thread after the if.

integration/nwo/fabric/network/network.go
```diff
@@ -236,60 +233,7 @@ func (n *Network) PostRun(load bool) {
...
-				err := n.DockerClient.RemoveImage(i.ID)
-				Expect(err).NotTo(HaveOccurred())
-				break
-			}
-		}
-	}
-
-	err = n.DockerClient.RemoveNetwork(nw.ID)
-	Expect(err).NotTo(HaveOccurred())
+	// DO nothing
```
adecaro, line 236: it looks like that n.DockerClient is still defined but not used anymore, shall we remove?

integration/nwo/fabric/platform.go
```diff
@@ -111,32 +101,23 @@ type Platform struct {
...
-		},
-	)
-	Expect(err).NotTo(HaveOccurred())
 
-	network := network.New(
+	n := network.New(
 		context,
 		t.(*topology.Topology),
-		dockerClient,
+		nil, // not needed anymore
```
adecaro, line 111: shall we remove it?

integration/nwo/orion/platform.go
```diff
@@ -220,164 +185,29 @@ func (p *Platform) Members() []grouper.Member {
...
-				NetworkID: net.ID,
-			},
-		},
-	}, nil, containerName)
-	if err != nil {
-		panic(err)
-	}
+func (p *Platform) Cleanup() {
+	dockerClient, err := docker.GetInstance()
+	Expect(err).NotTo(HaveOccurred())
```
adecaro, line 205: can we take the chance and add an error message to the new Expect introduced?
In this case, something like
Expect(err).NotTo(HaveOccurred(), "failed to get docker client")

integration/nwo/common/docker/utils.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+	return nil
+}
+
+func (d *Docker) CreateNetwork(networkID string) error {
+	_, err := d.Client.CreateNetwork(
+		docker.CreateNetworkOptions{
+			Name:   networkID,
+			Driver: "bridge",
```
adecaro, line 65: do we always use bridge?

adecaro (changes requested):

integration/nwo/common/docker/utils.go
mbrandenburger: good catch

integration/nwo/common/docker/utils.go
mbrandenburger: added a few comments where I find it useful. Note that we actually do not have new functions here. I was just moving code around and reuse existing things.

integration/nwo/orion/platform.go
mbrandenburger: I added more context to the actual error.

integration/nwo/common/docker/utils.go
adecaro: but did you fix it?

integration/nwo/common/docker/utils.go
adecaro: yeah, but still 😅

PR #309, opened 2022-07-07
Rework github actions

mbrandenburger:
- Merge integration tests workflows in a single flow
- The new test workflow first runs our `checks` target, if successful, the integration tests and unit tests are started in parallel using the github actions matrix features. Running the checks first gives as a fast fail without wasting github action resources on long unit/integration tests.
- The job matrix features helps to configure a job with different configurations. That is, we don't need to create a workflow file per integration test we would like to run in parallel.
- Docker images are only pulled for integration tests


mbrandenburger:
The unit tests are currently supposed to fail.
- PR #308 fixes the docker issue with the fsccli unit test
- The failure of the race issue needs to be tackled.

mbrandenburger:
Removed `unit-test-race` for now. I will create a separate PR for that.

PR #310, opened 2022-07-08
view-sdk: enhancements 

adecaro:
This PR introduces utility functions needed to address https://github.com/hyperledger-labs/fabric-token-sdk/issues/283 

Enhancements:
- add ability to retrive responder from the registry
- additional logging

Bug fix:
- fabric-sdk: update orderers list when apply config blocks after a restart

### DIFF
--- a/platform/view/core/manager/manager.go
+++ b/platform/view/core/manager/manager.go
@@ -101,6 +101,35 @@ func (cm *manager) RegisterResponderWithIdentity(responder view.View, id view.Id
 	return nil
 }
 
+func (cm *manager) GetResponder(initiatedBy interface{}) (view.View, error) {
+	var initiatedByID string
+	switch t := initiatedBy.(type) {
+	case view.View:
+		initiatedByID = cm.GetIdentifier(t)
+	case string:
+		initiatedByID = t
+	default:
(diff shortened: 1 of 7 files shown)


platform/view/core/manager/manager.go
```diff
@@ -101,6 +101,35 @@ func (cm *manager) RegisterResponderWithIdentity(responder view.View, id view.Id
...
+	}
+
+	entries, ok := cm.views[responderID]
+	if !ok {
+		return nil, errors.Errorf("responder not found for [%s], initiator [%s]", responderID, initiatedByID)
+	}
+	if len(entries) == 0 {
+		return nil, errors.Errorf("responder not found for [%s], initiator [%s]", responderID, initiatedByID)
+	}
+	return entries[0].View, nil
```
mbrandenburger, line 130: from this code it's not really clear why we just use the first element in `entries`. I would add a short comment here.

mbrandenburger (approved):

platform/view/core/manager/manager.go
adecaro: addressed. Thanks Marcus

PR #318, opened 2022-07-13
Add NWO support for existing fabric bins

mbrandenburger:


adecaro (approved):
LGTM

I guess, we should not have problems for the samples and the token-sdk, right?

PR #319, opened 2022-07-14
Generate connection.yaml

mbrandenburger:
In order to use FSC NWO to setup and run a stand-alone Fabric
network, applications based on the Fabric Client SDK require a
connection file to connect to the network. This commit introduces
the generation of such connection profiles.

Currently, it enabled with FPC only but can later extended.


adecaro (approved):
LGTM. Maybe in a future PR we can extend this to fabric

PR #320, opened 2022-07-21
Use time.NewTimer instead of time.After

yacovm:
Using time.After enqueues an entry in Go's runtime scheduler and doesn't garbage collect it even if the function the timer was fired from terminates.
As a result, there can be a temporary high memory pressure and it degrades performance.

Replaced all places where time.After was used except in places that are used as samples or tests.

### DIFF
--- a/platform/fabric/core/generic/committer/committer.go
+++ b/platform/fabric/core/generic/committer/committer.go
@@ -306,13 +306,17 @@ func (c *committer) listenTo(txid string, timeout time.Duration) error {
 		iterations = 1
 	}
 	for i := 0; i < iterations; i++ {
+		timeout := time.NewTimer(c.pollingTimeout)
+
 		select {
 		case event := <-ch:
 			if logger.IsEnabledFor(zapcore.DebugLevel) {
 				logger.Debugf("Got an answer to finality of [%s]: [%s]", txid, event.Err)
 			}
+			timeout.Stop()
 			return event.Err
-		case <-time.After(c.pollingTimeout):
+		case <-timeout.C:
+			timeout.Stop()
 			if logger.IsEnabledFor(zapcore.DebugLevel) {
 				logger.Debugf("Got a timeout for finality of [%s], check the status", txid)
 			}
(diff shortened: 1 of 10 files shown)


mbrandenburger:
Thanks @yacovm! I've added one more replacement of `timeAfter`

platform/fabric/core/generic/committer/committer.go
```diff
@@ -306,13 +306,17 @@ func (c *committer) listenTo(txid string, timeout time.Duration) error {
 		iterations = 1
 	}
 	for i := 0; i < iterations; i++ {
+		timeout := time.NewTimer(c.pollingTimeout)
```
adecaro, line 309: wasn't this the case where .After is okay?

platform/fabric/services/state/recipients.go
```diff
@@ -99,13 +99,16 @@ func (f RequestRecipientIdentityView) Call(context view.Context) (interface{}, e
 		return nil, err
 	}
 
+	timeout := time.NewTimer(time.Minute)
+	defer timeout.Stop()
```
adecaro, line 103: sometimes defer is used, other times no. Why this difference?

platform/orion/core/generic/committer/committer.go
```diff
@@ -312,13 +312,17 @@ func (c *committer) listenToFinality(txID string, timeout time.Duration) error {
 		iterations = 1
 	}
 	for i := 0; i < iterations; i++ {
+		timeout := time.NewTimer(c.pollingTimeout)
```
adecaro, line 315: wasn't this the case where After is okay?

adecaro (changes requested):

platform/fabric/services/state/recipients.go
yacovm: you cannot use defer in a loop

platform/orion/core/generic/committer/committer.go
yacovm: It's true that reading from the `ch` channel terminates the function, however I thought that this function can perhaps be called many times in a short period of time so it's better to stop the timer and not let them wait for being terminated with the timeout.

platform/fabric/core/generic/committer/committer.go
yacovm: same explanation as the other place

adecaro (approved):

PR #330, opened 2022-07-28
KVS extensions

adecaro:
- Return the key when iterating 
- Add support function for splitting keys

### DIFF
--- a/platform/view/services/kvs/kvs_test.go
+++ b/platform/view/services/kvs/kvs_test.go
@@ -106,7 +106,7 @@ func testRound(t *testing.T, cfg *fakeProv) {
 
 	for ctr := 0; it.HasNext(); ctr++ {
 		val = &stuff{}
-		err = it.Next(val)
+		_, err = it.Next(val)
 		assert.NoError(t, err)
 		if ctr == 0 {
 			assert.Equal(t, &stuff{"santa", 1}, val)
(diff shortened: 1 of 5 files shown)


platform/view/services/kvs/kvs_test.go
```diff
@@ -125,7 +125,7 @@ func testRound(t *testing.T, cfg *fakeProv) {
 
 	for ctr := 0; it.HasNext(); ctr++ {
 		val = &stuff{}
-		err = it.Next(val)
+		_, err = it.Next(val)
```
mbrandenburger, line 128: can you please add a test case to check that we get the expected key as well?!

mbrandenburger (changes requested):

platform/view/services/kvs/kvs_test.go
adecaro: done

platform/view/services/kvs/kvs_test.go
```diff
@@ -125,9 +127,10 @@ func testRound(t *testing.T, cfg *fakeProv) {
...
 	for ctr := 0; it.HasNext(); ctr++ {
 		val = &stuff{}
-		_, err = it.Next(val)
+		key, err := it.Next(val)
 		assert.NoError(t, err)
 		if ctr == 0 {
+			assert.Equal(t, k1, key)
 			assert.Equal(t, &stuff{"santa", 1}, val)
 		} else {
 			assert.Fail(t, "expected 2 entries in the range, found more")
```
mbrandenburger, line 136: do you want to take the chance to fix the error message here? It should be "expected 1 entry in the range, ..."

mbrandenburger (approved):
Thanks! LGTM

Issue #50, opened 2021-06-26
staticcheck: reduce warnings


adecaro:

mbrandenburger:
#305 Replaced golint with staticchecks.

mbrandenburger:
An assessment of the current code is needed and staticchecks issues should be resolved.

adecaro:
Shall address also this?

`go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.`

alexandrosfilios:
All PR's included
- [ ] ST1005
- [ ] ST1019
- [ ] S1023
- [ ] SA1019
- [ ] SA1019
- [ ] ST1001
- [ ] S1000
- [ ] S1034
- [ ] S1039
- [ ] S1011
- [ ] SA4023
- [ ] S021, S1038, S1025, SA4009, S1005, SA9003
- [ ] SA4004
- [ ] SA9005
- [ ] SA4006

PR #332, opened 2022-07-29
Reduce staticcheck warnings (ST1001)

alexandrosfilios:
* ST1001: should not use dot imports (excluded assertion-related packages)

### DIFF
--- a/staticcheck.conf
+++ b/staticcheck.conf
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0
+
+# onsi/gomega is used only for testing purposes (in test files and in the integration-test logic)
+dot_import_whitelist=["github.com/onsi/gomega"]
--- a/integration/nwo/fabric/network/deploy.go
+++ b/integration/nwo/fabric/network/deploy.go
@@ -22,7 +22,7 @@ import (
 	"github.com/onsi/gomega/gexec"
 
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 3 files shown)


staticcheck.conf
```diff
@@ -0,0 +1 @@
+dot_import_whitelist=["github.com/onsi/gomega", "github.com/onsi/gomega/gstruct"]
```
alexandrosfilios, line 1: I saw that other config files didn't have any header (license), but I can add it if it should be so.

Also I didn't add any comment for the exclusion, but if you prefer to have it, I can comment too

mbrandenburger: Let's add at least.
```
# SPDX-License-Identifier: Apache-2.0
```

mbrandenburger, line 1: Please add an empty line at the end of the file

mbrandenburger: I agree with adding `github.com/onsi/gomega` but do we really need to ignore `gstruct`?

It seems `gstruct` is imported only twice, therefore I think we should change these in a proper way.

mbrandenburger: > Also I didn't add any comment for the exclusion, but if you prefer to have it, I can comment too 

Adding short comment that gomega is heavily used in our tests would not harm but I think it's also not really needed

mbrandenburger (changes requested):

mbrandenburger (approved):
LGTM

PR #345, opened 2022-08-02
Reduce staticcheck warnings (S1021, S1038, S1025, SA4009, S1005, SA9003)

alexandrosfilios:
* S1021: merge var assignment and declaration
* S1038, S1025: remove redundant sprintf's
* SA4009: remove overwritten assignment
* S1005: removed unnecessary assignment to blank identifier
* SA9003: removed empty branch

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -78,8 +78,7 @@ func New(startPort int, path string, topologies ...api.Topology) (*Infrastructur
 
 	buildServer := common.NewBuildServer()
 	buildServer.Serve()
-	var builder api.Builder
-	builder = buildServer.Client()
+	builder := buildServer.Client()
 
 	n := &Infrastructure{
 		TestDir:      testDir,
--- a/integration/nwo/fabric/fpc/fpc.go
+++ b/integration/nwo/fabric/fpc/fpc.go
@@ -118,7 +118,7 @@ func (n *Extension) PostRun(load bool) {
 	}
 
 	if load {
-		// run docker containers
+		panic("Load is not supported. Restart from scratch.")
 	}
 
 	// deploy and run containers
--- a/integration/nwo/weaver/util.go
+++ b/integration/nwo/weaver/util.go
@@ -180,8 +180,11 @@ func gzipDir(dir, dest string) error {
 	defer t.Flush()
 
 	err = filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
-		var f *os.File
-		f, err = os.Open(path)
+		if err != nil {
+			return fmt.Errorf("failed accessing %s: %v", path, err)
+		}
+
+		f, err := os.Open(path)
 		if err != nil {
 			return fmt.Errorf("failed opening %s: %v", path, err)
 		}
--- a/integration/nwo/cmd/main.go
+++ b/integration/nwo/cmd/main.go
@@ -70,10 +70,8 @@ func VersionCmd(pName string) *cobra.Command {
 			}
 			// Parsing of the command line is done so silence cmd usage
 			cmd.SilenceUsage = true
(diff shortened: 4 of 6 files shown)


integration/integration.go
```diff
@@ -78,8 +78,7 @@ func New(startPort int, path string, topologies ...api.Topology) (*Infrastructur
 
 	buildServer := common.NewBuildServer()
 	buildServer.Serve()
-	var builder api.Builder
-	builder = buildServer.Client()
+	var builder = buildServer.Client()
```
mbrandenburger, line 81:
```suggestion
	builder := buildServer.Client()
```

integration/nwo/fabric/fpc/fpc.go
```diff
@@ -117,9 +117,9 @@ func (n *Extension) PostRun(load bool) {
 		return
 	}
 
-	if load {
-		// run docker containers
-	}
+	//if load {
```
mbrandenburger, line 120: we can remove these lines completely, the docker containers are spawned in line 130 `n.deployChaincode(chaincode)`

integration/nwo/weaver/util.go
```diff
@@ -179,7 +179,7 @@ func gzipDir(dir, dest string) error {
 	defer t.Close()
 	defer t.Flush()
 
-	err = filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
+	err = filepath.Walk(dir, func(path string, info fs.FileInfo, _ error) error {
 		var f *os.File
```
mbrandenburger, line 183: we should actually check if the err exists. See `Walk` docu (https://pkg.go.dev/path/filepath#Walk)

what about the following:
```go
err = filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
    if err != nil {
        return fmt.Errorf("failed accessing path=%s: %v", path, err)
    }
    // if no error, we can open the file
    var f *os.File
    f, err = os.Open(path)
    ...
```

mbrandenburger (changes requested):

integration/nwo/fabric/fpc/fpc.go
alexandrosfilios: Sure, then the `load` variable is not used at all in this method. (I can't remove it from the signature though, because it's part of the interface it implements.)

integration/nwo/weaver/util.go
alexandrosfilios: I applied the change. If a sub-directory is not accessible for some reason (permission denied or corrupt data maybe), this would have been ignored before (and we would move on to the next sub-dirs), but not anymore. I guess this is never the case (and it's not a test case either), but just to be sure, is it OK to keep this change?

mbrandenburger: > If a sub-directory is not accessible for some reason (permission denied or corrupt data maybe), this would have been ignored before (and we would move on to the next sub-dirs)

We still move on the next sub-dir in case of an error. Now we are just doing right away instead of trying to open the file even in the presence of an error.

integration/nwo/fabric/fpc/fpc.go
```diff
@@ -112,15 +112,11 @@ func (n *Extension) GenerateArtifacts() {
 	}
 }
 
-func (n *Extension) PostRun(load bool) {
+func (n *Extension) PostRun(bool) {
```
mbrandenburger, line 115: does static check complain if we keep the variable but not using it? Ideally, we can keep the method header unchanged

alexandrosfilios: It doesn't complain indeed. I removed the variable, because it's marked as unused in the IDE and the other implementations of the `PostRun` method don't define the variable. Let me revert it back

mbrandenburger (approved):

PR #347, opened 2022-08-02
Reduce staticcheck warnings (SA9005)

alexandrosfilios:
* SA9005: missing marshaling mechanism

### DIFF
--- a/platform/orion/core/generic/vault/rwset.go
+++ b/platform/orion/core/generic/vault/rwset.go
@@ -419,3 +419,7 @@ func (r *readSet) clear(ns string) {
 	}{}
 	r.orderedReads[ns] = []string{}
 }
+
+func (rws *readWriteSet) String() string {
+	return "{}"
+}
--- a/platform/orion/core/generic/vault/rwset.go
+++ b/platform/orion/core/generic/vault/rwset.go
@@ -419,7 +419,3 @@ func (r *readSet) clear(ns string) {
 	}{}
 	r.orderedReads[ns] = []string{}
 }
-
-func (rws *readWriteSet) String() string {
-	return "{}"
-}
--- a/platform/orion/core/generic/vault/vault.go
+++ b/platform/orion/core/generic/vault/vault.go
@@ -8,7 +8,6 @@ package vault
 
 import (
 	"bytes"
-	"encoding/json"
 	"sync"
 
 	"github.com/pkg/errors"
@@ -146,8 +145,7 @@ func (db *Vault) CommitTX(txid string, block uint64, indexInBloc int) error {
 	db.storeLock.Lock()
 	defer db.storeLock.Unlock()
 
-	m, _ := json.Marshal(i.rws)
-	logger.Debugf("committing \n[%s]\n", string(m))
+	logger.Debugf("committing \n[%s]\n", i.rws.String())
 
 	err = db.store.BeginUpdate()
 	if err != nil {
--- a/platform/orion/core/generic/vault/vault.go
+++ b/platform/orion/core/generic/vault/vault.go
@@ -145,8 +145,6 @@ func (db *Vault) CommitTX(txid string, block uint64, indexInBloc int) error {
 	db.storeLock.Lock()
 	defer db.storeLock.Unlock()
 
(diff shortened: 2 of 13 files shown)


integration/fabric/atsa/chaincode/chaincode/asset_transfer.go
```diff
@@ -437,24 +436,24 @@ func transferAssetState(ctx contractapi.TransactionContextInterface, asset *Asse
...
-
-	assetReceipt := receipt{
-		price:     price,
-		timestamp: timestamp,
-	}
-	receipt, err := json.Marshal(assetReceipt)
-	if err != nil {
-		return fmt.Errorf("failed to marshal receipt: %v", err)
-	}
+	//txTimestamp, err := ctx.GetStub().GetTxTimestamp()
```
alexandrosfilios, line 439: Removing this logic doesn't seem to break any tests and from a static code analysis, it seems redundant. Can I remove it?

platform/orion/core/generic/vault/vault.go
```diff
@@ -146,8 +145,7 @@ func (db *Vault) CommitTX(txid string, block uint64, indexInBloc int) error {
 	db.storeLock.Lock()
 	defer db.storeLock.Unlock()
 
-	m, _ := json.Marshal(i.rws)
-	logger.Debugf("committing \n[%s]\n", string(m))
+	logger.Debugf("committing \n[%s]\n", i.rws.String())
```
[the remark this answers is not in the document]
mbrandenburger: since the implementation of the db driver will also use json Marshal on the rws later, this here was just for debugging purpose. 
If logging this rws as json is desired this should happen only in a `if logger.Level() == DEBUG` case

integration/fabric/atsa/chaincode/chaincode/asset_transfer.go
mbrandenburger: Are you sure that static check can catch this one? Where should be other other impl of this redundant code?

mbrandenburger (changes requested):

platform/orion/core/generic/vault/rwset.go
```diff
@@ -419,3 +419,7 @@ func (r *readSet) clear(ns string) {
 	}{}
 	r.orderedReads[ns] = []string{}
 }
+
+func (rws *readWriteSet) String() string {
+	return "{}"
+}
```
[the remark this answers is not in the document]
alexandrosfilios: No indeed neither this nor the commented code is for merging. I wanted to ask for your opinion on these. This is the equivalent behavior to the current code (like the commented code).

integration/fabric/atsa/chaincode/chaincode/asset_transfer.go
alexandrosfilios: this code is only written for the test, so if the integration test doesn't fail there I don't think there should be any other reason for it to be there. This code is not repeated somewhere else in the codebase. I didn't find a reason for it to be there from its first commit and I didn't find any file where `receipt` implemented `Marshaler` (or `Unmarshaler`)

integration/fabric/atsa/chaincode/chaincode/asset_transfer.go
```diff
@@ -437,24 +436,24 @@ func transferAssetState(ctx contractapi.TransactionContextInterface, asset *Asse
...
+	//if err = :txTimestamp.CheckValid(); err != nil {
+	//	return err
+	//}
+	//timestamp := txTimestamp.AsTime()
+	//
+	//assetReceipt := receipt{
+	//	price:     price,
+	//	timestamp: timestamp,
+	//}
+	receipt := []byte{}
```
alexandrosfilios, line 453: Check under fabric-samples if it is used.

mbrandenburger: https://github.com/hyperledger/fabric-samples/blob/main/asset-transfer-basic/chaincode-go/chaincode/smartcontract.go

mbrandenburger (approved):

PR #351, opened 2022-08-02
Reduce staticcheck warnings (SA4023)

alexandrosfilios:
* SA4023: replaced unnecessary null checks

### DIFF
--- a/platform/fabric/services/chaincode/endorse.go
+++ b/platform/fabric/services/chaincode/endorse.go
@@ -66,16 +66,14 @@ func (i *endorseChaincodeView) Endorse(context view.Context) (*fabric.Envelope, 
 	}
 
 	var chaincode Chaincode
-	chaincode = &stdChaincode{ch: channel.Chaincode(i.ChaincodeName)}
-	if chaincode == nil {
-		return nil, errors.Errorf("fabric chaincode [%s:%s:%s] not found", i.Network, i.Channel, i.ChaincodeName)
-	}
-	if chaincode.IsPrivate() {
-		logger.Debugf("chaincode [%s:%s:%s] is a FPC", i.Network, i.Channel, i.ChaincodeName)
+	stdChannelChaincode := channel.Chaincode(i.ChaincodeName)
+	if stdChannelChaincode.IsPrivate() {
 		// This is a Fabric Private Chaincode, use the corresponding service
 		fpcChannel := fpc.GetChannel(context, i.Network, i.Channel)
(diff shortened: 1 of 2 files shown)


platform/fabric/services/chaincode/endorse.go
```diff
@@ -65,11 +65,7 @@ func (i *endorseChaincodeView) Endorse(context view.Context) (*fabric.Envelope,
 		i.SignerIdentity = fNetwork.IdentityProvider().DefaultIdentity()
 	}
 
-	var chaincode Chaincode
-	chaincode = &stdChaincode{ch: channel.Chaincode(i.ChaincodeName)}
-	if chaincode == nil {
-		return nil, errors.Errorf("fabric chaincode [%s:%s:%s] not found", i.Network, i.Channel, i.ChaincodeName)
-	}
+	var chaincode Chaincode = &stdChaincode{ch: channel.Chaincode(i.ChaincodeName)}
```
mbrandenburger, line 68: what about the following?

```go
var chaincode Chaincode
if channel.Chaincode(i.ChaincodeName).IsPrivate() {
    // This is a Fabric Private Chaincode, use the corresponding service
    fpcChannel := fpc.GetChannel(context, i.Network, i.Channel)
    chaincode = &fpcChaincode{fpcChannel.Chaincode(i.ChaincodeName)}
    logger.Debugf("chaincode [%s:%s:%s] is a FPC", i.Network, i.Channel, i.ChaincodeName)
    
} else {
    chaincode = &stdChaincode{ch: channel.Chaincode(i.ChaincodeName)}
    logger.Debugf("chaincode [%s:%s:%s] is a standard chaincode", i.Network, i.Channel, i.ChaincodeName)
}
```

similar construct below

mbrandenburger (changes requested):

platform/fabric/services/chaincode/endorse.go
alexandrosfilios: Sure, I wanted to make the same change, but I didn't know if I'm missing something conceptual there. Should I then remove the method `Chaincode.IsPrivate`, as well? It won't be used anymore after this change. (Beware the `Chaincode` struct returned by `channel.Chaincode` does not implement the `Chaincode` interface that is implemented by `fpcChaincode` and `stdChaincode`.)

mbrandenburger: `Chaincode.IsPrivate` seems to be a method of a local interface, therefore I believe we can remove it. If static check does not complain about that I would keep it for now since this PR actually deals with unnecessary null checks

mbrandenburger (approved):

Issue #316, opened 2022-07-12
The fsc node approver and iou process memory leak, the memory grows to 16G in about 10 minutes under 50-60TPS and exits abnormally


yaofeiliang:

adecaro:
@yaofeiliang , this is very interesting. Can you give us more information? Do you have profile data that you can share for us to analyze? Do you know where the memory is leaked?

mbrandenburger:
Hi @yaofeiliang 

can you please provide additional information which let me reproduce this issue. I was wondering how you invoke the 50k transactions? How did you measure the memory consumption?

Could you also try to enable monitoring in the IOU topology and share the Grafana logs.
```go
	// Monitoring
	monitoringTopology := monitoring.NewTopology()
	//monitoringTopology.EnableHyperledgerExplorer()
	monitoringTopology.EnablePrometheusGrafana()

	return []api.Topology{
		fabricTopology,
		fscTopology,
		monitoringTopology,
	}
```

Once the network is up and running using `./iou network start --path ./testdata`, you can log into Grafana via your browser on `localhost:3000` as `admin` with the default password `admin`.

The Fabric Backend Dashboard will show you the memory consumption of the FSC nodes and Fabric peers.

To run some workload over time I was testing with the following script but I did not observe any abnormal memory consumption.

```bash
#!/usr/bin/env bash

outFile="out.txt"

for i in {1..1000}; do
  echo "run: $i"
  ./iou view -c ./testdata/fsc/nodes/borrower/client-config.yaml -f create -i "{\"Amount\":10}" > $outFile
  id=$(head -n 1 $outFile | tr -d '"'s)
  ./iou view -c ./testdata/fsc/nodes/borrower/client-config.yaml -f query -i "{\"LinearID\":\"$id\"}"
  ./iou view -c ./testdata/fsc/nodes/lender/client-config.yaml -f query -i "{\"LinearID\":\"$id\"}"
  ./iou view -c ./testdata/fsc/nodes/borrower/client-config.yaml -f update -i "{\"LinearID\":\"$id\",\"Amount\":8}"
done
```

yaofeiliang:
@adecaro @mbrandenburger 
![6711657789789_ pic](https://user-images.githubusercontent.com/31167188/178947186-7e0a5303-d2cf-4c86-ae79-d79ee2a110c5.jpg)
[...]

PR #355, opened 2022-08-04
FSC memory enhancements

### LINKED ISSUE #339: Enhance KVS cache
Currently KVS impl uses a simple map for caching. (https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/view/services/kvs/kvs.go#L60)

While for testing that would be OK - for long living nodes this is not ideal.

The KVS cache should be revisited and replaced with a more robust caching solution. Potential candidates: 
- Ristretto as this is already used within badger
- Secondcache (https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/view/services/cache/secondcache/second_chance.go)

Moreover, the cache size (num items or size) and TTL should be configurable.

mbrandenburger:
This PR addresses the issue of increasing memory consumption as reported in #316.

- Improve integration test runner
The current implementation of runners use goexec to run processes (i.e., orderers, peers, fsc nodes) during our integration tests. goexec uses unbounded buffers to capture the process outputs. While this works nicely for short-living integration tests, other uses of our integration test suite, such as the IOU sample, may ran for long time and will suffer from high memory consumption over the time. Replaces `goexec.Start` with native `cmd.Run` and avoids the use of unbounded buffers. In order to detect when a processes has reached the ready state, we will use a closeable buffer to collect process logs and close it if not needed anymore.

- Replace KVS cache and thereby resolves #339.
The current implementation of our KVS uses an unbounded cache. This cache is replaced with our secondchance cache.

- Vault txidstore is loaded now from config; if not available default cache size is used. default cache size has been reduced to a sane value


mbrandenburger:
TODOs
- [ ] vault cache size via config
- [ ] kvs cache size via config

mbrandenburger:
@yaofeiliang Can you please check with this PR if the situation gets better with your setup.
The IOU process has a very small memory footprint now. The FSC node is also better. But my feeling is there is still room for improvements :D

platform/fabric/core/generic/channel.go
```diff
@@ -172,6 +177,27 @@ func newChannel(network *network, name string, quiet bool) (*channel, error) {
...
 }
 
+var defaultCacheSize = 100
+
+// cacheSizeFromConfig returns the KVS cache size from current configuration.
+// Returns defaultCacheSize, if no configuration found.
+// Returns an error and defaultCacheSize, if the loaded value from configuration is invalid (must be >= 0).
+func cacheSizeFromConfig(sp view2.ServiceProvider) (int, error) {
+	configService := view2.GetConfigService(sp)
+	configKey := "fabric.vault.txidstore.cache.size"
```
alexandrosfilios, line 187: You could probably use this as a constant.

I know it's implemented like this also for the others, but maybe it would make sense for the config service to provide methods like
```
GetInt(key string) (int, boolean)
```

mbrandenburger: > You could probably use this as a constant.

Good catch!  added constants for the config keys used in the KVS.

Regarding your other proposal to change the config API, I think this is a great idea. We should revisit the configuration component and do a cleanup in the entire repo. @alexandrosfilios could you please create an issue for that?

For this PR I would like to keep the `GetInt` methods as is to follow the existing pattern. I've also seen that even that change required a few tweaks here and there to make other unit tests run again. I've added a few todos to address that.

integration/nwo/fabric/topology/core_template.go
```diff
@@ -310,6 +310,10 @@ fabric:
         type: file
         opts:
           path: {{ FSCNodeVaultPath }}
+      txidstore:
```
adecaro, line 313: do we need this also for orion?

platform/fabric/core/generic/channel.go
```diff
@@ -172,6 +177,27 @@ func newChannel(network *network, name string, quiet bool) (*channel, error) {
 	return c, nil
 }
 
+var defaultCacheSize = 100
```
adecaro, line 180: shall we have this variable included in the already existing var block at the top of the file?

platform/fabric/core/generic/channel.go
```diff
@@ -172,6 +177,27 @@ func newChannel(network *network, name string, quiet bool) (*channel, error) {
...
+
+// cacheSizeFromConfig returns the KVS cache size from current configuration.
+// Returns defaultCacheSize, if no configuration found.
+// Returns an error and defaultCacheSize, if the loaded value from configuration is invalid (must be >= 0).
+func cacheSizeFromConfig(sp view2.ServiceProvider) (int, error) {
+	configService := view2.GetConfigService(sp)
```
adecaro, line 186: this would not work because we might have multiple fabric configurations. Use channel's config field to get that

adecaro (changes requested):

integration/nwo/fabric/topology/core_template.go
mbrandenburger: Orion does not use a txidstore cache at all! So if we want it let's create an issue for that

platform/fabric/core/generic/channel.go
mbrandenburger: moved it to vault

platform/fabric/core/generic/channel.go
mbrandenburger: good catch! updated!

integration/nwo/fabric/topology/core_template.go
adecaro: I see, that's true. All right. Let's address it elsewhere.

adecaro (approved):
LGTM

PR #370, opened 2022-08-17
Add WithEndorsersByMSPIDs to query chaincode view

mbrandenburger:
This adds a missing option to select the endorsing peers via MSPIDs when
performing a query. We already have this option for invoke.


adecaro (commented):
LGTM

PR #371, opened 2022-08-18
Cleanup deps.go

alexandrosfilios:
Some of the blank dependencies in deps.go are not needed anymore.

### DIFF
--- a/deps.go
+++ b/deps.go
@@ -9,7 +9,6 @@ SPDX-License-Identifier: Apache-2.0
 package deps
 
 import (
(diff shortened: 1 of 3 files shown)


adecaro (approved):
LGTM.

Do we need such an update for the Token-SDK as well?

alexandrosfilios:
> LGTM.
> 
> Do we need such an update for the Token-SDK as well?

Yep there is one already in Draft, but this one needs to be merged first, because the changes here are used in token SDK.

Issue #359, opened 2022-08-05
Fix flaky tests


mbrandenburger:
Some tests seem to be flaky, such as the FPC integration test.

- Short-term solution: 
Upgrade to ginkgo v2 as proposed in #331 and add `ginkgo --flake-attempts=N` to the corresponding make target of our flaky candidate.

- Long-term:
Investigate what is wrong and provide a proper fix.

mbrandenburger:
PR 375 disables [integration/nwo/fsc/tracing/aggregator_test.go](https://github.com/hyperledger-labs/fabric-smart-client/pull/375/files#diff-2d88ca8cb6f12ee4cd7e1b104b4634228b7bf0cf0a3056fac1f19e15a7e11248) temporary

adecaro:
@mbrandenburger can we close this?

PR #375, opened 2022-08-25
fabric-sdk: do not add resolvers for Fabric peers

adecaro:
This PR updates NWO to remove from the resolvers references to Fabric peers.
The FPC integration was relaying in these resolvers. To fix this, the PR updates the FPC integration to leverage Fabric Discovery.

In addition, a flaky unit-test has been disabled.

### DIFF
--- a/integration/nwo/fsc/tracing/aggregator_test.go
+++ b/integration/nwo/fsc/tracing/aggregator_test.go
@@ -85,6 +85,7 @@ func measure() {
 }
 
 func TestNewAggregatorFakeInput(t *testing.T) {
+	t.Skip()
 	var l Listener
 	var err error
 	go func() {
--- a/integration/fabric/fpc/echo/topology.go
+++ b/integration/fabric/fpc/echo/topology.go
@@ -28,6 +28,7 @@ func Topology() []api.Topology {
 
 	// Create an empty FSC topology
 	fscTopology := fsc.NewTopology()
+	fscTopology.SetLogging("debug", "")
 
 	// Alice
 	alice := fscTopology.AddNodeByName("alice")
--- a/integration/nwo/fabric/network/resolver.go
+++ b/integration/nwo/fabric/network/resolver.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package network
 
 import (
-	"fmt"
 	"path/filepath"
 
 	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api"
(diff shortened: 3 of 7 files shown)


integration/nwo/fsc/tracing/aggregator_test.go
```diff
@@ -85,6 +85,7 @@ func measure() {
 }
 
 func TestNewAggregatorFakeInput(t *testing.T) {
+	t.Skip()
```
mbrandenburger, line 88: 🙅

mbrandenburger (commented):
DO we need another cleanup in the fabric config?

A quick search for resolvers show a few places. Here a few pointers I question if we still need them. 
- https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/config/endpoint.go
- https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/config/config.go#L185

adecaro:
I didn't remove all resolvers. They are still needed to have some of the integration tests working.

Let's take the `IOU` test. Alice asks the Approver to approve. The Approver signs using a Fabric identity. Alice must check that this signature is valid, therefore Alice needs to know which key the approver will use. This key must be in Alice's resolvers.

integration/nwo/fsc/tracing/aggregator_test.go
adecaro: I know, but it was failing continuously for no reason 😢

mbrandenburger:
> Let's take the `IOU` test. Alice asks the Approver to approve. The Approver signs using a Fabric identity. Alice must check that this signature is valid, therefore Alice needs to know which key the approver will use. This key must be in Alice's resolvers.

Ok - makes sense to me.
However, in that case it's wired to look for these identities in the `fabric.endpoint.resolvers` section. I would expect to have this in a `fabric.identity.resolvers` section.

adecaro:
I understand your point but it is out of scope for this PR to change that configuration path.
I would prefer to dedicate a PR to the configuration restructuring accompanied by proper documentation.
Wdyt?

mbrandenburger:
BTW Can't we use the org ca cert to check these identities?

adecaro:
what do you mean?

mbrandenburger:
> what do you mean?

If Alice forms a transaction and requires Bob signature, in particular, any signature from Bob's org, for the endorsement. So instead of looking up the identity of Bob, could Alice simply check the Bob's cert (attached with the endorsement) belongs to Bob org? Anyway, this seems orthogonal to this PR.

adecaro:
do you need any additional change then? :)

adecaro:
@alexandrosfilios , please, have a look ASAP. Thanks :)

mbrandenburger (approved):
LGTM! Please angelo, rebase and squash the fixups. We would like to keep the commit for disabling the flaky test separately. :)

integration/nwo/fsc/tracing/aggregator_test.go
mbrandenburger: Captured this temporary disablement in https://github.com/hyperledger-labs/fabric-smart-client/issues/359

PR #369, opened 2022-08-12
fabric-sdk: support for PKCS11 #281

adecaro:
- removed bccsp section that was never used
- the default msp must be included directly under msps
- possibiilty to pass options
- PKCS11 BCCSP support and configuration
- Option to add to an FSC node a Fabric identity whose secret key is stored in the HSM
- Extend NWO to generate the HSM identities:
  - Add an HSM flag to `PeerIdentity`
  - `integration/nwo/fabric/network/network_support.go Line 678`: Parse HSM identities from Opts and add `PeerIdentity`
  - Extend `cryptogen` to use HSM
- Extend the `IOU` integration test to use this identity. The burrower can use this identity for the ownership of the IOU state and the approver to approve operations.

PR #383, opened 2022-08-30
hsm support for mac

sapthasurendran:
Signed-off-by: sapthasurendran <saptha.surendran@ibm.com>
Updated softhsm setup script to add support for mac

### DIFF
--- a/ci/scripts/install_softhsm.sh
+++ b/ci/scripts/install_softhsm.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+#Install softhsm
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    brew install softhsm
+else
+    sudo apt-get install -y softhsm2
+fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -98,6 +98,10 @@ jobs:
         working-directory: ${{ env.FSC_PATH }}
         run: make docker-images
 
+      - name: Set up softhsm
+        working-directory: ${{ env.FSC_PATH }}
+        run: make install-softhsm
+
       - name: Run ${{ matrix.tests }}
         working-directory: ${{ env.FSC_PATH }}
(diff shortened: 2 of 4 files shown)


mbrandenburger:
Can you please rebase this PR as https://github.com/hyperledger-labs/fabric-smart-client/pull/369 is merged.

ci/scripts/install_softhsm.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+#Install softhsm
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    brew install softhsm
```
davidkel, line 9: I'm pretty sure brew installs softhsm 2.6. This version of softhsm hasn't played well with fabric in the past and could also have similar issues with the FSC implementation of HSM support as it's likely to be using similar code. Unfortunately I can't remember what the problem was with softhsm 2.6, but it's explicitly documented in the latest fabric docs not to use 2.6.

mbrandenburger: Interesting,

I was looking for this statement in the docs but cannot find it. https://hyperledger-fabric.readthedocs.io/en/latest/hsm.html

davidkel: @mbrandenburger You will find it here https://hyperledger-fabric.readthedocs.io/en/latest/dev-setup/devenv.html
I'm trying to remember what the conflict was but has been so long since I worked on fabric HSM I can't remember. It's also possible now that newer versions of the Go C Runtime addressed the issue that occurred with Softhsm 2.6
@sapthasurendran This could still be a problem with fabric though even if the Go client application doesn't exhibit a problem, so might be good to test the newer versions of fabric with Softhsm 2.6 again to see if it is resolved

sapthasurendran: @mbrandenburger The integration test as well as fabric-samples seems to work fine with soft-hsm 2.6 version. I think this PR is good to merge. 
@davidkel I will  test Fabric with softhsm 2.6v on a linux vm as you suggested as a side project .

mbrandenburger (approved):
LGTM

PR #394, opened 2022-09-27
minor enhancements

adecaro:
- NWO, fabric CheckTopologyFSCNodes fails if an FSC node doesn't have an organization
- NWO: skip orion is no role is set
- String function added to Envelope (fabric and orion)
- Improved logs and errors


mbrandenburger (approved):
LGTM

PR #399, opened 2022-09-28
Minor updates to IOU sample README and code for simplicity and clarity

bestbeforetoday:


adecaro:
@bestbeforetoday , thanks for submitting this PR. Would it be possible to link directly the file containing the code rather than copying it?

bestbeforetoday:
> @bestbeforetoday , thanks for submitting this PR. Would it be possible to link directly the file containing the code rather than copying it?

We could use a GitHub permalink to link to a block of code at a specific commit. I think this will mean the code is not visible in the README.md itself, although I can check if GitHub does anything clever to inline links. I'm not sure if the document will be so clear if people need to click off to separate pages to see some of the code snippets.

mbrandenburger:
I believe "includes" are not supported in markdown. Could we generate the markdown to ensure the readme is consistent with the sample code? Maybe a "simple" shell script could do the job.

bestbeforetoday:
I think additional enhancements to dynamically or programmatically generate the README content would be better in a separate PR

mbrandenburger (approved):
Thanks! LGTM!

PR #400, opened 2022-09-28
Upgrade github.com/libp2p/go-libp2p version

bestbeforetoday:
Latest versions allow samples to be run with Go 1.19.


adecaro:
@bestbeforetoday , my only worry here is that we don't know if the rest of our stacks will be able to switch to go1.19 yet :(
Indeed, I have asked Yacov to revert the upgrade to go1.18 in the meantime that we understand if we can move the rest to it.

bestbeforetoday:
That's fine. I noticed Yacov's revert PR; just didn't see any accompanying rationale. I wasn't trying to move anything to Go1.19. The intent is just to allow it to run on Go1.19 without breaking compatibility with earlier Go version, since I happened to be using Go1.19 when running the samples, and they cannot run at all with Go1.19 as-is.

mbrandenburger (approved):
LGTM. Tested locally with go 1.18.

mbrandenburger:
@adecaro As we do not revert back to oooold go 1.17, this PR seems to be nice as it works nicely with go 1.18 and newer

adecaro:
I'm wondering if this new version can cause issues with load balancers and so on 😅

bestbeforetoday:
This has been sat here a long time and now has merge conflicts, so I'll close it.

PR #415, opened 2022-10-11
restructure fsc and fabric driver core to be more intuitive

davidkel:
1. create a grpc section in core and move appropriate keys to be children
2. sort out tls for fabric-driver section
3. make web client configuration keys match intent

### DIFF
--- a/node/node/start.go
+++ b/node/node/start.go
@@ -149,12 +149,13 @@ func serve() error {
 		callback(err)
 		return err
 	}
+
 	callback(nil)
 
 	cs := view.GetConfigService(node.Registry())
 	logger.Infof("Started peer with ID=[%s], address=[%s]",
 		cs.GetString("fsc.id"),
-		cs.GetString("fsc.address"),
+		cs.GetString("fsc.grpc.address"),
 	)
 	return <-serve
 }
--- a/platform/fabric/core/generic/config/config.go
+++ b/platform/fabric/core/generic/config/config.go
@@ -95,23 +95,11 @@ func (c *Config) TLSClientCertFile() string {
 }
 
 func (c *Config) KeepAliveClientInterval() time.Duration {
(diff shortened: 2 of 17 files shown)


node/node/start.go
```diff
@@ -149,12 +149,14 @@ func serve() error {
 		callback(err)
 		return err
 	}
+
+	// TODO: This code is not called
```
adecaro, line 153: why? node.Start is non-blocking, no?

platform/fabric/core/generic/config/config.go
```diff
@@ -95,13 +95,14 @@ func (c *Config) TLSClientCertFile() string {
...
-	return c.configService.GetDuration("fabric." + c.prefix + "tls.keepalive.client.interval")
+	return c.configService.GetDuration("fabric." + c.prefix + "keepalive.interval")
 }
 
 func (c *Config) KeepAliveClientTimeout() time.Duration {
-	return c.configService.GetDuration("fabric." + c.prefix + "tls.keepalive.client.timeout")
+	return c.configService.GetDuration("fabric." + c.prefix + "keepalive.timeout")
 }
 
+/*
```
adecaro, line 105: let's remove this

adecaro (changes requested):
what about some documentation describing the configuration?

node/node/start.go
davidkel: When I ran this code it never gets passed the Start so I never see the log message so as far as I could tell it was blocking

platform/fabric/core/generic/config/config.go
davidkel: oops thought I had done that, will remove

adecaro (approved):

mbrandenburger (approved):

PR #417, opened 2022-10-14
Fix npe if no core file found

### LINKED ISSUE #406: Missing core.yaml shouldn't cause a segmentation fault
A missing `core.yaml` is a problem - and should stop the code from starting. But it shouldn't cause a segmentation fault. 
```
   Received unexpected error Config File "core" Not Found in "[/home/matthew/_workspace/src/github.com/gunners-coin/bin]"
```
A valid error case but shouldn't then cause
```
 [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12d0193]
```

davidkel:
closes #406

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -107,7 +107,9 @@ func (n *node) Start() (err error) {
 
 func (n *node) Stop() {
 	n.running = false
-	n.cancel()
+	if n.cancel != nil {
+		n.cancel()
+	}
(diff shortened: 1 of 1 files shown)


adecaro (approved):
LGTM, let's wait the CI and I'll merge

PR #420, opened 2022-10-18
sample config with doc

davidkel:
This is a sample core.yaml file for fsc and fabric-driver, for discussion to add and improve the associated documentation

### DIFF
--- a/exampleconfigs/README.md
+++ b/exampleconfigs/README.md
@@ -1,0 +1,4 @@
+# Core.yaml examples
+
+This directory contains descriptive core.yaml files
+- core-fabric-example.yaml is a core.yaml that describes the fsc and fabric-driver configuration properties. 
--- a/exampleconfigs/core-fabric-example.yaml
+++ b/exampleconfigs/core-fabric-example.yaml
@@ -1,0 +1,343 @@
+---
+# ------------------- Logging section ---------------------------
+logging:
+ # format is same as fabric [<logger>[,<logger>...]=]<level>[:[<logger>[,<logger>...]=]<level>...]
+ # available loggers: TBD
+ spec: debug
+
+# ------------------- FSC Node Configuration -------------------------
+fsc:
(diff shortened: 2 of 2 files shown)


exampleconfigs/core-fabric-example.yaml
```diff
@@ -0,0 +1,328 @@
...
+    # and that entry must have an address with an entry P2P.
+    bootstrapNode: approver
+
+  # ------------------- KVS Configuration -------------------------
+  # Internal key/value store used by the node to store information
+  # such as bindings (eg resolvers)
+  kvs:
+    cache:
+      # Sets the maximum number of cached items
+      size:
```
mbrandenburger, line 87: empty size, will it be set to 0 in that case?

mbrandenburger (commented):
Great step forward! The example config has a few references to the `approver` identity. If one does not carefully read, it might be understood as a default value or a FSC specific role.
Should we replace them with some other generic identities? E.g., `someOtherId`? Alternative we should also briefly explain the setting where we have a node that uses this config and another one having `fsc.id: approver`. WDYT?

exampleconfigs/core-fabric-example.yaml
davidkel: That's a good point, if any of the entries doesn't have a value, what is it's default. I will look at adding those

exampleconfigs/README.md
```diff
@@ -0,0 +1,4 @@
+# Core.yaml examples
```
adecaro, line 1: shall we link this document from the documentation we have under `./doc`?

exampleconfigs/core-fabric-example.yaml
```diff
@@ -0,0 +1,343 @@
+---
```
adecaro, line 1: we need to add a section also for Orion

davidkel:
Closing as #427 supercedes this

PR #421, opened 2022-10-18
enhancement pack

adecaro:
- NWO fabric and orion: ability to remove the vault of FSC nodes
- Introduce the `PostStart` interface. PostStart allows a platform to start services after all the other platforms have been started. This feature has been introduce to allow platforms to add hookups in the fabric and orion commit pipelines before the pipelines start. Fabric and Orion commit pipelines are now activated in post-start.
- fabric chaincode: when querying a chiancode, discovery performs now a `peer query` and not an `endorser query`.
- Fabric and Orion Committer: if the status of a transaction is requested, if this transaction does not exist in the vault but its envelope exists, then the envelope is loaded into the vaul.

The changes to the commit pipelines and `PostStart` allow a node to rebuild its vault if the KVS has not been compromised.

Used by: https://github.com/hyperledger-labs/fabric-token-sdk/pull/406

### DIFF
--- a/platform/orion/core/generic/delivery/delivery.go
+++ b/platform/orion/core/generic/delivery/delivery.go
@@ -52,7 +52,6 @@ type DeliverStream interface {
 }
 
 type delivery struct {
-	ctx                 context.Context
 	sp                  view2.ServiceProvider
 	network             Network
 	waitForEventTimeout time.Duration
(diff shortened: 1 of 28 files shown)


platform/orion/core/generic/delivery/delivery.go
```diff
@@ -52,51 +52,54 @@ type DeliverStream interface {
...
 	var err error
 	for {
 		select {
-		case <-d.ctx.Done():
+		case <-d.stop:
+			// Time to cancel
+			return errors.New("stop")
+		case <-ctx.Done():
 			// Time to cancel
 			return errors.New("context done")
```
mbrandenburger, line 104: why would you return an error when you send a "graceful" stop?

mbrandenburger: In line `155` there the we also stop without an error.

mbrandenburger (approved):

platform/orion/core/generic/delivery/delivery.go
adecaro: addressed

PR #423, opened 2022-10-19
NWO to only use prebuilt fabric binaries

davidkel:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 # pinned versions
-FABRIC_VERSION=2.2
+FABRIC_VERSION ?= 2.2.8
+FABRIC_TWO_DIGIT_VERSION = $(shell echo $(FABRIC_VERSION) | cut -d '.' -f 1,2)
 ORION_VERSION=v0.2.5
+
+# need to install fabric binaries outside of fsc tree for now (due to chaincode packaging issues)
+FABRIC_BINARY_BASE=$(PWD)/../fabric
+FAB_BINS ?= $(FABRIC_BINARY_BASE)/bin
 
 # integration test options
 GINKGO_TEST_OPTS ?=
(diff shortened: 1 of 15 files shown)


integration/nwo/fabric/network/network.go
```diff
@@ -181,16 +181,23 @@ func (n *Network) GenerateArtifacts() {
...
 }
 
-func (n *Network) Members() []grouper.Member {
+func (n *Network) Members() ([]grouper.Member, error) {
 	members := grouper.Members{}
 
-	if r := n.OrdererGroupRunner(); r != nil {
-		members = append(members, grouper.Member{Name: n.Prefix + ".orderers", Runner: r})
+	runner, err := n.OrdererGroupRunner()
+	if err != nil {
```
adecaro, line 188: If we use `Expect(err).NotTo(HaveOccurred(), "failed to get orderer group members`)`, we could avoid to return an error here. We do this everywhere in NWO.
What do you think? 
This would avoid also the changes to the Platform interface.

samples/fabric/iou/README.md
```diff
@@ -457,12 +457,19 @@ func Topology() []api.Topology {
 
 ### Boostrap the networks
```
adecaro, line 458: shall we add this section also to the other samples?

adecaro:
@davidkel, this looks good to me. Just a few comments to better understand certain aspects.

README.md
```diff
@@ -39,6 +39,23 @@ git clone https://github.com/hyperledger-labs/fabric-smart-client.git $FSC_PATH
 
 You are ready to run the [`samples`](./samples/README.md) in `$FSC_PATH`.
 
+### Further information
```
adecaro, line 42: Shall we add a reference to the Fabric version we test against?

integration/nwo/fabric/network/network.go
davidkel: I can do that, I wasn't sure whether it was better to surface the error back up to the main point of NWO to panic it there or not. I guess this call will actually do a panic as well. I could just do that at the point where it looks for the command and the command was not found. What do you think ?

samples/fabric/iou/README.md
davidkel: Yep definitely, I missed the weaver sample and will look to putting something in the main sample readme as well

README.md
davidkel: We could, the version we test against is defined in the Makefile so we could either be explicit or reference the makefile. What do you think ?

integration/nwo/fabric/network/network.go
adecaro: yep, that works as well 👍

samples/fabric/iou/README.md
adecaro: 👍

README.md
adecaro: let's reference the Makefile, so that people know how to change that if they want :)

Makefile
```diff
@@ -71,65 +80,67 @@ orion-server-images:
...
-integration-tests-atsafsc:
-	cd ./integration/fabric/atsa/fsc; ginkgo $(GINKGO_TEST_OPTS) .
+integration-tests-atsafsc: download-fabric
+	cd ./integration/fabric/atsa/fsc; export FAB_BINS=$(FAB_BINS); ginkgo $(GINKGO_TEST_OPTS) .
 
 .PHONY: integration-tests-twonets
-integration-tests-twonets:
-	cd ./integration/fabric/twonets; ginkgo $(GINKGO_TEST_OPTS) .
+integration-tests-twonets: download-fabric
+	cd ./integration/fabric/twonets; export FAB_BINS=$(FAB_BINS); ginkgo $(GINKGO_TEST_OPTS) .
```
mbrandenburger, line 123: If a developer runs the test locally, `FAB_BINS` is not required to be set manually, instead it triggers the `download-fabric` target and exports `FAB_BINS`. I am wondering what happens if the developer sets`FAB_BINS` and then runs `make integration-tests-X`. The `download-fabric` target is called and maybe something is downloaded (if not already exists) but then the `FAB_BINS` set by the developer is overwritten within the integration test target. Is it working like this or do I miss something?

mbrandenburger: I would remove the `download-fabric` target as dep and just call it in our CI setup phase - similar what we do with our go tools.

Makefile
```diff
@@ -1,7 +1,12 @@
 # pinned versions
-FABRIC_VERSION=2.2
+FABRIC_VERSION ?= 2.2.8
+FABRIC_TWO_DIGIT_VERSION = $(shell echo $(FABRIC_VERSION) | cut -d '.' -f 1,2)
 ORION_VERSION=v0.2.5
 
+# need to install fabric binaries outside of fsc tree for now (due to chaincode packaging issues)
+FABRIC_BINARY_BASE=$(PWD)/../fabric
+FAB_BINS=$(FABRIC_BINARY_BASE)/bin
```
mbrandenburger, line 8: If I want to use my custom bins within this makefile, should we be able to override like this ... `FAB_BINS ?= $(FABRIC_BINARY_BASE)/bin` ??

mbrandenburger (commented):
This is a nice PR

Makefile
davidkel: The only limitation here is using your own builds of fabric, otherwise if you want to try different versions of fabric then all you do is 
```
FABRIC_VERSION=2.4.6 make integration-tests
```
Without it being there, contributors to FSC have to perform the download-fabric themselves first (and manually do it if they want to change versions). I guess it depends on the usability aspects here of users here. However I think it's only for real experts that they would provide their own builds of fabric and as such I would think it would be trivial for them to run the tests and we could allow someone to override FAB_BINS as per your next suggestion

Happy to change it if you still prefer that

Makefile
davidkel: Yep, that's a great idea, thanks

Makefile
mbrandenburger: Well, let's assume I have already somewhere my fab bins and I just want to use them. Apparently, there are users running on a platform that is not compatible with whatever you can download :)

If I now use `FAB_BINS` to override it, running a test target will work fine. However, the `download-fabric` target is still triggered and downloads the bins to the folder as specified in `FABRIC_BINARY_BASE`. Even if I don't need them, the download is triggered. I think that is an annoying behaviour.

I understand that we are adding another manual step to prep the system before running the tests. However, we already do that for the go-tools and docker images anyway. To make life easy, these targets could be bundled in a `setup-dev-env` (or call it similarly) target. 

Letting the user manually trigger the `download-fabric` target has also security implication. We don't want to force users to download "arbitrary" binaries and just execute them. By doing this step manually by default, users need to decide if they are fine with just downloading them or taking over the responsibility to get their own bins from somewhere. For CI this argument does not hold, but I kind of appreciate it if our build system does not download executables without actively asking me if this is what I want :)

davidkel:
Closing PR due to messed up git repo, will resubmit

PR #439, opened 2022-11-28
key validation removal

NorhanKhaled:
Currently, FSC Vault is more restrictive when parsing RW sets imposing rules that Fabric does not actually impose.
This results in panic when transactions with a RW sets containing keys that do not pass the FSC check are committed to Fabric.

This PR fixes this issue by removing the extra checks.

Signed-off-by: Norhan Ali Araba <Norhan.Ali.Araba-CICNetherlands@ibm.com>


adecaro (approved):
LGTM

Issue #419, opened 2022-10-17
Web client needs to handle error responses


davidkel:
In platform/view/services/client/web/client.go

```
	resp, err := c.c.Do(req)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to process http request to [%s], input length [%d]", url, len(in))
	}
	buff, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, errors.Wrapf(err, "failed to read response from http request to [%s], input length [%d]", url, len(in))
	}

	response := &protos2.CommandResponse_CallViewResponse{}
	err = json.Unmarshal(buff, response)
```

always assumes that the response will be a callviewresponse but this may not be the case. For example if you try to send http to an https server or if both are http and you try to call a view the response isn't of the format and you get unhelpful messages about the response

Issue #408, opened 2022-10-06
Web Server should be able to be configured to not accept client requests to call views


davidkel:
The Web Server has multiple purposes
1. Accept client requests to call views
2. respond to /version request
3. respond to /healthz request
4. respond to /metrics request (prometheus)

Currently the web server can either be enabled with or without tls but if tls is enabled it will only work with mutual tls. Without TLS is really there for test environments. mutual tls is being mandated as an authorisation mechanism (rather than using basic auth for example, in a similar manner to the way that osnadmin works on fabric). 

However this makes K8s checks for /healthz difficult if not impossible to manage for example. 

Admins should have the option to not use mutualTLS but if it's disabled then maybe we should disable the ability to call views but still allow access to the other routes for node management.

yacovm:
> mutual tls is being mandated as an authorisation mechanism (rather than using basic auth for example, in a similar manner to the way that osnadmin works on fabric).

> However this makes K8s checks for /healthz difficult if not impossible to manage for example.

No, it does not. You can make mutual TLS optional and then enforce checks for all endpoints except /healthz.

davidkel:
@yacovm In the webserver I can't find anything detailing how to make mutualtls optional, all I can find is to turn tls off altogether which I see disables the ability to call views. Can you provide more information on how I turn off mutual tls but keep tls itself ? Looking at the code it's not obvious (unless just not specifying something in core.yaml) actually controls that. In which case that would be useful to know and I can include it as part of the information I am putting together about the contents of core.yaml

PR #442, opened 2022-12-06
bug fix and enhancements

adecaro:
This PR brings the following:
- fix a bug that caused a panic when the web server was disabled
- ability to disable the grpc service
- remove the admin section from configuration and added a description of the client section
- Web client needs to handle error responses #419
- Web Server should be able to be configured to not accept client requests to call views #408


platform/view/core/id/provider.go
```diff
@@ -59,9 +59,9 @@ func (p *provider) Load() error {
 		return errors.WithMessagef(err, "failed loading default identity")
 	}
 
-	if err := p.loadAdminIdentities(); err != nil {
-		return errors.WithMessagef(err, "failed loading admin identities")
-	}
+	//if err := p.loadAdminIdentities(); err != nil {
```
HagarMeir, line 62: why comment out and not erase?

HagarMeir (approved):

PR #475, opened 2023-05-04
introduce query policy

adecaro:
This PR allows the developer to choose a policy to decide how many responses are needed to accept the result of a query. The PR also allows the developer to make use of an external context to track the progress.

### DIFF
--- a/platform/fabric/services/chaincode/query.go
+++ b/platform/fabric/services/chaincode/query.go
@@ -10,12 +10,26 @@ import (
 	"time"
 
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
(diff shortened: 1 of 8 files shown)


platform/fabric/services/chaincode/query.go
```diff
@@ -10,13 +10,27 @@ import (
...
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 	"github.com/pkg/errors"
 )
 
+// QueryPolicy defines the policy to use to decide if a query is successful
+type QueryPolicy int
+
+const (
+	// QueryAll requires an answer from all selected peers
+	QueryAll QueryPolicy = iota
```
alexandrosfilios, line 24: Maybe we can use the constant defined above

adecaro:
I close this. If needed we will address this topic later.

PR #484, opened 2023-07-26
bidirectional channel between the invoker of a view and the view itself

adecaro:
This PR introduces the support for a bidirectional channel between the invoker of a view and the view itself.
This PR supports this channel in two ways:
- Using GRPC stream, and
- Using WebSockets

The `pingpong` integration test has been extended to test this functionality. 
This PR is in support of https://github.com/hyperledger-labs/fabric-token-sdk/pull/478


integration/nwo/client/web.go
```diff
@@ -21,17 +20,14 @@ func NewWebClientConfigFromFSC(confDir string) (*web.Config, error) {
...
+	config.Host = configProvider.GetString("fsc.web.address")
 	if configProvider.GetBool("fsc.web.tls.enabled") {
-		config.URL = fmt.Sprintf("https://%s", configProvider.GetString("fsc.web.address"))
-		tlsRootCertFile := configProvider.GetString("fsc.web.tls.serverRootCert.file")
-		if len(tlsRootCertFile) == 0 {
-			return nil, errors.New("web configuration must have serverRootCert with file key defined")
+		config.TLSCertPath = configProvider.GetPath("fsc.web.tls.cert.file")
+		config.TLSKeyPath = configProvider.GetPath("fsc.web.tls.key.file")
+		if len(config.TLSCertPath) == 0 {
+			return nil, errors.New("web configuration must have sc.web.tls.cert.file with file key defined")
```
mbrandenburger, line 28: I suggest to define the keys here as `const strings`

integration/nwo/common/context/context.go
```diff
@@ -39,7 +39,8 @@ type Context struct {
 	hostByOrdererID    map[string]string
 	extensionsByPeerID map[string]api.Extensions
 
-	ViewClients             map[string]api.ViewClient
+	ViewClients             map[string]api.GRPCClient
+	WebClients              map[string]api.WebClient
```
mbrandenburger, line 43: mh ... I think these terms are not really good here. 

What is a ViewClient? Something that can trigger view actions from remote, right? We have two implementations for that a) GRPC and b) web. What if we add another impl in the future?

Is there a need that both impls are used at the same time?  

We are giving up a bit of abstracting here. For instance, `ViewIdentities` may also have different implementations ... we could have idemix, or whatever, right?

mbrandenburger: So could we also try to hide the impl of the actual view client here and let the system figure out via configuration, if we want to start the FSC node with GRPC or web support, or both?

platform/view/core/manager/context.go
```diff
@@ -269,6 +269,10 @@ func (ctx *ctx) ResetSessions() error {
 	return nil
 }
 
+func (ctx *ctx) PutService(service interface{}) error {
+	return ctx.localSP.RegisterService(service)
+}
+
```
mbrandenburger, line 275: While I strongly believe that there should exist a `PutX` if we already have a `GetX` function, I believe call it `RegisterService` makes more sense to me :D

platform/view/services/client/web/wsclient.go
```diff
@@ -0,0 +1,58 @@
...
+	return &WSStream{conn: ws}, nil
+}
+
+func (c *WSStream) Send(v interface{}) error {
+	return c.conn.WriteJSON(v)
+}
+
+func (c *WSStream) Recv(v interface{}) error {
+	return c.conn.ReadJSON(v)
+}
```
mbrandenburger, line 42: What do we send and receive here? the same serialzed proto messages as we use for GRPC? What happens if we send them as JSON? would it be more efficient to send as bytes directly?!

platform/view/services/server/view/protos/finality.proto
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 syntax = "proto3";
 
-option go_package = "protos";
+option go_package = "./protos";
```
mbrandenburger, line 9: this looks wrong to me :/

platform/view/services/server/view/protos/service.proto
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Apache-2.0
 
 syntax = "proto3";
 
-option go_package = "protos";
+option go_package = "./protos";
```
mbrandenburger, line 9: same comment :D

platform/view/services/server/view/view.go
```diff
@@ -98,3 +149,34 @@ func (s *viewHandler) runView(view view.View, context *view.Context) {
...
+	return c.SendProtoMsg(s)
+}
+
+func (c *Stream) Recv(m interface{}) error {
+	s := &protos2.CallViewResponse{}
+	if err := c.RecvProtoMsg(s); err != nil {
+		return err
+	}
+	return json.Unmarshal(s.Result, m)
+}
```
mbrandenburger, line 174: ohh my ... what are we doing here? Is this a wild mix of json + GRPC ....

platform/view/services/view/stream.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"reflect"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view"
+)
+
+type Stream interface {
+	Recv(m interface{}) error
+	Send(m interface{}) error
+}
```
mbrandenburger, line 18: I suggest that we use bytes already here and let the caller perform the serialization.

PR #495, opened 2023-09-14
support for multiple fabric transaction format

adecaro:
This PR brings in the following:
- Support for multiple Fabric transaction formats.
- Removal of `fmt.Errorf`


alexandrosfilios (approved):
LGTM

PR #498, opened 2023-10-06
Identity Provider fetches identity by Alias too

adecaro:


platform/view/core/endpoint/endpoint.go
```diff
@@ -198,6 +198,15 @@ func (r *service) GetIdentity(endpoint string, pkid []byte) (view.Identity, erro
 				break
 			}
 		}
+		if !found {
```
alexandrosfilios, line 201: you can do
```
found = found || slices.Contains(resolver.Aliases, endpoint)
```

platform/view/core/endpoint/endpoint.go
```diff
@@ -232,6 +241,14 @@ func (r *service) AddResolver(name string, domain string, addresses map[string]s
 			r.resolversMutex.RUnlock()
 			return resolver.Id, r.Bind(resolver.Id, id)
 		}
+		for _, alias := range resolver.Aliases {
+			for _, newAlias := range aliases {
```
alexandrosfilios, line 245: also here you could use
```
for _, alias := range resolver.Aliases {
 if slices.Contains(aliases, alias) {
  r.resolvers...
 }
}
```

alexandrosfilios (approved):

platform/view/core/endpoint/endpoint.go
adecaro: beautiful 💌

PR #501, opened 2023-10-09
load params from configuration

adecaro:
This PR has the scope to remove the magic constants and load them from the configuration


alexandrosfilios (approved):
LGTM

PR #511, opened 2023-11-28
Upgrade FPC version

mbrandenburger:
- Upgrade FPC dependency
- Remove unsupported idemix tests from fpc echo test


adecaro (approved):
LGTM

PR #516, opened 2023-12-13
cleanup dependencies

adecaro:
- remove dependency to libp2p from packages that should not have it
- fabric dependencies from the view platform

### DIFF
--- a/platform/view/services/comm/pki.go
+++ b/platform/view/services/comm/pki.go
@@ -1,0 +1,64 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package comm
+
+import (
+	"crypto/ecdsa"
+	"crypto/x509"
+	"encoding/pem"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/hash"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type PKISupport struct{}
+
+func (p PKISupport) ExtractPublicKey(id view.Identity) any {
+	certRaw, _ := pem.Decode(id)
+	switch {
+	case certRaw != nil:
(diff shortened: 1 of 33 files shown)


platform/view/services/comm/pki.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil
+	}
+}
+
+func (p PKISupport) PublicKeyID(key any) []byte {
+	switch d := key.(type) {
+	case *ecdsa.PublicKey:
+		raw, err := x509.MarshalPKIXPublicKey(d)
+		if err != nil {
+			return nil
```
alexandrosfilios, line 49: Should we log the error or is it expected that it happens?

platform/view/services/comm/pki.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type PKISupport struct{}
+
+func (p PKISupport) ExtractPublicKey(id view.Identity) any {
+	certRaw, _ := pem.Decode(id)
+	switch {
```
alexandrosfilios, line 24: Maybe replacing the switch statement with a nil check at the beginning could be more readable:
```
certRaw, _ := pem.Decode(id)
if certRaw == nil {
 return nil
}
cert, err := x509.ParseCertificate(certRaw.Bytes)
...
```

alexandrosfilios (approved):

platform/view/services/hash/hashable.go
```diff
@@ -13,6 +13,21 @@ import (
 
 type Hashable []byte
 
+func (id Hashable) Raw() []byte {
```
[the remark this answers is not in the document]
adecaro: 👍

platform/view/services/comm/pki.go
adecaro: good point 👍

platform/view/services/comm/pki.go
adecaro: 👍

PR #523, opened 2024-02-05
remove support for git download

adecaro:
https://github.com/hyperledger-labs/fabric-smart-client/security/dependabot/89

### DIFF
--- a/integration/nwo/common/builder.go
+++ b/integration/nwo/common/builder.go
@@ -12,18 +12,13 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/flogging"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


integration/nwo/common/builder.go
```diff
@@ -120,58 +115,13 @@ func (a *artifact) build(args ...string) error {
...
-			logger.Infof("building [%s,%s] with race detection ", input, args)
-			return gexec.Build(input, args...)
-		}
-
-		logger.Infof("building [%s,%s] ", input, args)
+	if a.raceDetectorEnabled && !strings.HasPrefix(input, "github.com/hyperledger/fabric/") {
+		logger.Infof("building [%s,%s] with race detection ", input, args)
 		return gexec.Build(input, args...)
 	}
+
```
alexandrosfilios, line 122: Maybe a check could be added here on what you expect input to be (e.g. not start with git@).

alexandrosfilios (approved):

integration/nwo/common/builder.go
adecaro: no one ever used `git@`, the PR removes completely the support for that.

PR #524, opened 2024-02-07
add sql as possible driver for kvs and vault

arner:
This PR allows the user to select a golang/sql compatible driver for the vault and key/valuestore as an alternative to Badger. 

Reasons to do that may include:

- Using a managed database for high availability, failover and backups with very little overhead
- Having a stateless instance of the Fabric Smart Client with no filesystem writes and an external database
- The ability to inspect stored data using standard tooling
- Compliance to organization policies

The driver has been tested with the following sql drivers:

- SQLite: (pure go): https://modernc.org/sqlite
- Postgres (pure Go): https://github.com/lib/pq 

Note that for SQLite it's recommended to add a pragma to prevent BUSY_TIMEOUT errors, e.g. `file:/var/fsc/data/fsc.sqlite?_pragma=busy_timeout(5000)`.


platform/view/services/db/driver/sql/init.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func getTableName(prefix, name string) (table string, valid bool) {
+	table = fmt.Sprintf("%s_%s", prefix, name)
+	r := regexp.MustCompile("^[a-zA-Z_]+$")
+	return table, r.MatchString(name)
+}
+
+func init() {
+	db.Register("sql", &Driver{})
```
mbrandenburger, line 136: I am wondering if this should happen as part of the `sdk.go` as it is responsible to orchestrate all the drivers needed. Having the `init` function without the use of build flags comes with the risk that we lose track on where we load certain code.

platform/view/services/db/driver/sql/sql_test.go
```diff
@@ -0,0 +1,146 @@
...
+	p := NewUnversioned(db, key)
+	if err := p.CreateSchema(); err != nil {
+		t.Fatal(err)
+	}
+	return p
+}
+
+// https://testcontainers.com/guides/getting-started-with-testcontainers-for-go/
+// Note: Before running tests: docker pull postgres:16.0-alpine
+// Test may time out if image is not present on machine.
```
mbrandenburger, line 119: testcontainers does not download the image if not locally available?
if not, we should add it to https://github.com/hyperledger-labs/fabric-smart-client/blob/main/Makefile#L51

mbrandenburger: Since we are using docker at other places, would we use the already existing docker api instead of introducing another dependency here?

platform/view/services/db/driver/sql/base.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (db *base) Close() error {
+	logger.Info("closing database")
+
+	// TODO: what to do with db.txn if it's not nil?
+
+	err := db.db.Close()
+	if err != nil {
+		return errors.Wrap(err, "could not close DB")
```
mbrandenburger, line 33: `github.com/pkg/errors` has been archived and we should transition to move away from it. 
Simple wrapping can be done with the `fmt` package. For example: `fmt.Errorf("decompress %v: %w", name, err)`

platform/view/services/db/driver/sql/base.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if err != nil {
+		return false, errors.Wrap(err, "cannot check if key exists")
+	}
+	return true, nil
+}
+
+func (db *base) DeleteState(ns, key string) error {
+	if db.txn == nil {
+		panic("programming error, writing without ongoing update")
```
mbrandenburger, line 111: is there a better way to deal with this error other than `panic`? Does a user of FSC may trigger is panic or can it only happen if we mess up something inside the FSC code?

platform/view/services/db/driver/sql/base.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if ns == "" || key == "" {
+		return errors.New("ns or key is empty")
```
mbrandenburger, line 114: Same comment here. I suggest to use https://pkg.go.dev/fmt#Errorf

platform/view/services/db/driver/sql/sql_test.go
```diff
@@ -0,0 +1,146 @@
...
+func startPostgresContainer(t *testing.T) (func(), string) {
+	// if os.Getenv("TESTCONTAINERS") != "true" {
+	// 	t.Skip("set environment variable TESTCONTAINERS to true to include postgres test")
+	// }
+	if testing.Short() {
+		t.Skip("skipping postgres test in short mode")
+	}
```
mbrandenburger, line 126: I guess it would also be useful to allow to use an existing postgres instance running on the developer machine. For instance, a URL/credentials to an existing database could be passed via env vars and uses instead of starting a fresh test container.

platform/view/services/db/driver/sql/init.go
arner: This is how it's done everywhere in FSC and token sdk... @adecaro might have an opinion?

platform/view/services/db/driver/sql/sql_test.go
arner: Changed code to use docker api and added image to makefile

platform/view/services/db/driver/sql/base.go
arner: I'm not sure, copied this from the badger db. @adecaro?

platform/view/services/db/driver/sql/sql_test.go
arner: I'm not sure about that - the code creates tables and data and doesn't clean it up, so using an ephemeral container is nice. The way the testing code is set up, it would be easy to set up the feature of using your own db if someone needs it (for instance by providing a postgres connection string)

mbrandenburger:
@adecaro @alexandrosfilios Can you guys please have a look at this one. Thanks

platform/view/services/db/driver/sql/init.go
adecaro: I think we will discuss this in the near future given that we are discussing the introduction of dependency injection frameworks. 
Right now, we use the same approach that is used by the `sql` package in the Go sdk.

platform/view/services/db/driver/sql/base.go
adecaro: That function, and others, should only be called if a transaction is currently active.
Now, we can decide to limit further the use of panics to really only situations that cannot be recovered. 
Are we in such a case? Not sure, honestly.
In general, also a user can trigger a panic. Yes.

adecaro:
It could be interesting to discuss here the following topic: Both the KVS and the Vault are key-value based stores.
Aren't there key-value stores that met all the requirements that @arner required?

Makefile
```diff
@@ -49,7 +49,7 @@ unit-tests-race:
 	cd integration/nwo/; export FAB_BINS=$(FAB_BINS); go test -cover ./...
 
 .PHONY: docker-images
-docker-images: fabric-docker-images weaver-docker-images fpc-docker-images orion-server-images monitoring-docker-images
+docker-images: fabric-docker-images weaver-docker-images fpc-docker-images orion-server-images monitoring-docker-images testing-docker-images
```
alexandrosfilios, line 52: I guess the name of the command is temporary or is it for tests?

platform/view/services/db/driver/sql/init.go
```diff
@@ -0,0 +1,137 @@
...
+	db.SetMaxOpenConns(maxOpenConns)
+	if err = db.Ping(); err != nil {
+		return nil, err
+	}
+	logger.Infof("connected to [%s,%d]", driverName, maxOpenConns)
+
+	return db, nil
+}
+
+func (d *Driver) getOps(config driver.Config) (Opts, error) {
```
alexandrosfilios, line 107: It doesn't need to be a method of the driver

platform/view/services/db/driver/sql/unversioned.go
```diff
@@ -0,0 +1,146 @@
...
+
+	val = append([]byte(nil), val...)
+
+	// Portable upsert
+	exists, err := db.exists(db.txn, ns, key)
+	if err != nil {
+		return err
+	}
+	if exists {
+		query := fmt.Sprintf("UPDATE %s SET val = $1 WHERE ns = $2 AND pkey = $3", db.table)
```
alexandrosfilios, line 47: I know this is Postgres specific, but this implementation is also related only to Postgres, so you could also use the Postgres version of UPSERT (INSERT INTO ... ON CONFLICT ...) to avoid the existence check.

platform/view/services/db/driver/sql/unversioned.go
```diff
@@ -0,0 +1,146 @@
...
+		args = []interface{}{ns, startKey, endKey}
+	} else if startKey != "" {
+		where = "AND pkey >= $2"
+		args = []interface{}{ns, startKey}
+	} else if endKey != "" {
+		where = "AND pkey < $2"
+		args = []interface{}{ns, endKey}
+	}
+
+	query := fmt.Sprintf("SELECT pkey, val FROM %s WHERE ns = $1 ", db.table) + where + " ORDER BY pkey;"
```
alexandrosfilios, line 101: You could continue with Sprintf instead of concatenating

platform/view/services/db/driver/sql/unversioned.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return val, nil
+}
+
+func (db *Unversioned) GetStateRangeScanIterator(ns string, startKey string, endKey string) (driver.ResultsIterator, error) {
+	where := ""
+	args := []interface{}{ns}
+
+	// To match badger behavior, we don't include the endKey
+	if startKey != "" && endKey != "" {
```
alexandrosfilios, line 90: You could extract this string creation into a util function in case you need it again.

platform/view/services/db/driver/sql/versioned.go
```diff
@@ -0,0 +1,288 @@
...
+	err = decoder.Decode(&m)
+	return
+}
+
+func (db *Persistence) GetStateRangeScanIterator(ns string, startKey string, endKey string) (driver.VersionedResultsIterator, error) {
+	where := ""
+	args := []interface{}{ns}
+
+	// To match badger behavior, we don't include the endKey
+	if startKey != "" && endKey != "" {
```
alexandrosfilios, line 186: We could reuse the logic we had in another method before.

arner:
> It could be interesting to discuss here the following topic: Both the KVS and the Vault are key-value based stores. Aren't there key-value stores that met all the requirements that @arner specified?

Ok I created this PR to just make it easy for a Token SDK user to connect to the same database for FSC. So forgive me if I take it too far now, but... Now that I can see the data in a viewer I noticed that the data stored in VersionedPersistence and SimpleTxIDStore is (or can be) quite structured:

**Write (from a read/write set)**
- namespace
- block
- indexInBlock
- value
- metadata (key/val)
- version

**Transaction**
- id (auto incrementing counter)
- txid
- status (Valid, Invalid, Busy, Unknown, HasDependencies)

Depending on how they're used they could be stored in a single table or linked via a txid. Transient maps and envelopes are stored in the 'Unversioned' KVS but are closer to this and can be linked 1:1 to a Transaction. Besides transientMaps and envelopes the KVS also stores signers and identities, which are also be more relational by nature than I initially thought. It seems we have something like this:

**Endpoint**
- id
- ephemeral (bytes, public key)
- public key (bytes)
- addresses (see below)

**Endpoint address**
- endpoint id
- portname (Listen, View, P2P)
- address

**Signer**
- id (e.g. "bob" if it's an owner)
- ephemeral key
(or something like that)

**Endpoint binding**
- signer
- endpoint

This does not mean an RDBMS is necessarily a better fit and I don't know if it's worth the effort, but you could argue that normalizing the data model (and optimizing the code for that...) could improve simplicity and clarity a lot.

Makefile
arner: It's for tests 😅. Do you have an idea for a better name?

platform/view/services/db/driver/sql/versioned.go
```diff
@@ -0,0 +1,288 @@
...
+	if tx == nil {
+		panic("programming error, writing without ongoing update")
+	}
+	logger.Debugf("set state [%s,%s]", ns, key)
+
+	val = append([]byte(nil), val...)
+
+	// Note: INSERT ON CONFLICT works for postgres and sqlite, but there is no single-shot upsert in the sql standard.
+	// Here we sacrifice a bit of performance to remain compatible with other databases.
+	exists, err := db.exists(tx, ns, key)
```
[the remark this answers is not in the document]
arner: while I like the idea, I propose to have it as a future improvement to keep the initial PR simple, what do you think?

Makefile
alexandrosfilios: Oh ok I thought Postgres was going to be an option for production, my bad.

platform/view/services/db/dbtest/bench.go
```diff
@@ -0,0 +1,84 @@
...
+		k = fmt.Sprintf("key_%d", i)
+
+		err = db.BeginUpdate()
+		_ = err
+		err = db.SetState(namespace, k, payload, 0, 0)
+		_ = err
+		err = db.Commit()
+	}
+	b.StopTimer()
+	returnErr = err
```
alexandrosfilios, line 81: Do we use returnErr and returnVal somewhere else? Because here we could also do this with the local var

PR #526, opened 2024-02-15
Reuse code

alexandrosfilios:


adecaro:
what is the goal of this PR?

alexandrosfilios:
> what is the goal of this PR?

To move the dependencies of some services into their constructor in order to remove some usages of the serviceProvider. Note that in this project we don't use any dependency injection. I just put the services in the constructor instead of passing the whole SP.

PR #535, opened 2024-03-14
fabric: commit unknown fix

adecaro:


platform/fabric/core/generic/committer.go
```diff
@@ -125,29 +127,36 @@ func (c *Channel) commitUnknown(txID string, block uint64, indexInBlock int, env
...
-		if err := c.FetchAndStoreEnvelope(txID); err != nil {
+		envelopeRaw, err = c.FetchEnvelope(txID)
+		if err != nil {
 			return errors.WithMessagef(err, "failed getting rwset for tx [%s]", txID)
 		}
 	}
 
 	// shall we commit this unknown envelope
-	if ok, err := c.filterUnknownEnvelope(txID); err != nil || !ok {
+	if ok, err := c.filterUnknownEnvelope(txID, envelopeRaw); err != nil || !ok {
```
alexandrosfilios, line 147: It doesn't make a big difference, but this method actually needs a `common.Envelope`, so we do a redundant unmarshaling. Then `StoreEnvelope` needs the byte array, so I would calculate both variables `envelope` and `envelopeRaw` in the if statements above. Then I'd pass `envelope` in filterUnknownEnvelope` and `envelopeRaw` in `StoreEnvelope`.

platform/fabric/core/generic/rwset.go
```diff
@@ -70,3 +70,25 @@ func (c *RWSetLoader) GetRWSetFromETx(txID string) (driver.RWSet, driver.Process
...
+		return nil, nil, errors.Wrapf(err, "failed unpacking envelope [%s]", txID)
+	}
+	logger.Debugf("retrieve rws [%s,%s]", c.Channel, txID)
+
+	rws, err := c.Vault.InspectRWSet(upe.Results)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return rws, upe, nil
```
alexandrosfilios, line 93: If you don't need `upe`, you can remove it from the outputs for now.

alexandrosfilios (approved):
LGTM modulo comments

adecaro:
@alexandrosfilios , if you are okay, I would address your comments in another PR because I want to change the way we handle the unknown transaction :)

alexandrosfilios:
> @alexandrosfilios , if you are okay, I would address your comments in another PR because I want to change the way we handle the unknown transaction :)

Sure totally fine! :)

PR #539, opened 2024-03-27
enhanced commit pipelines (fabric and orion)

adecaro:
This PR introduces the following to the commit pipelines of fabric and orion:
- introduce ErrDiscardTX to explicitly signal that a transaction needs to be discarded even though it is valid for the backend.
- introduce validation message to be stored in the vault.


platform/fabric/core/generic/committer/committer.go
```diff
@@ -195,12 +199,15 @@ func (c *Committer) IsFinal(ctx context.Context, txID string) error {
 					if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 199: the cases for Busy and HasDependencies implement the same logic. We could group them. Also we could remove one nesting by doing
```
for _, id := range deps {
...
}
if len(deps) > 0 {
return nil
}
```
But this is more of personal taste

platform/fabric/core/generic/committer/committer.go
```diff
@@ -195,12 +199,15 @@ func (c *Committer) IsFinal(ctx context.Context, txID string) error {
...
-					err := c.Finality.IsFinal(txID, c.Network.PickPeer(driver.PeerForFinality).Address)
+					peer := c.Network.PickPeer(driver.PeerForFinality).Address
+					err := c.Finality.IsFinal(txID, peer)
 					if err == nil {
+						logger.Debugf("Tx [%s] is final, remote check on [%s]", txID, peer)
 						return nil
 					}
 
-					if vd, _, err2 := committer.Status(txID); err2 == nil && vd == driver.Unknown {
+					if vd, _, _, err2 := committer.Status(txID); err2 == nil && vd == driver.Unknown {
```
alexandrosfilios, line 209: We can rename `err2` to `err`

platform/fabric/core/generic/vault/txidstore/cache.go
```diff
@@ -10,15 +10,20 @@ import (
 	fdriver "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
 )
 
+type Entry struct {
+	ValidationCode    fdriver.ValidationCode
+	ValidationMessage string
+}
+
 type cache interface {
```
alexandrosfilios, line 18: We could use generics to avoid casting

platform/fabric/core/generic/rwset.go
```diff
@@ -56,6 +60,10 @@ func (c *RWSetLoader) GetRWSetFromEvn(txID string) (driver.RWSet, driver.Process
 }
 
 func (c *RWSetLoader) GetRWSetFromETx(txID string) (driver.RWSet, driver.ProcessTransaction, error) {
+	if !c.TransactionService.Exists(txID) {
```
alexandrosfilios, line 63: I would skip calling `Exists` because in the background it calls load, so we end up calling load twice.

alexandrosfilios (approved):
LGTM

PR #543, opened 2024-04-10
fabric refactor

adecaro:
This PR does the following: It refactors the fabric platform without changing its functionalities. The services are better separated and then composed in the `Network` and `Channel` interface. The committer service has been unified and cleaned up.


platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	peerMapping := map[driver.PeerFunctionType][]*grpc.ConnectionConfig{}
+	for _, v := range peers {
+		v.TLSEnabled = tlsEnabled
+		if v.TLSDisabled {
+			v.TLSEnabled = false
+		}
+		usage := strings.ToLower(v.Usage)
+		switch {
```
alexandrosfilios, line 120: pretty minor but we could have a static map:
```
var funcTypeMap = map[string][PeerFunctionType]{
 "": driver.PeerForAnything,
 "delivery": driver.PeerForDelivery,
 ...
}
```
Then the loop would be:
```
for _, v := range peers {
 v.TLSEnabled = !v.TLSDisabled && tlsEnabled
 if funcType, ok := funcTypeMap[strings.ToLower(v.Usage)]; ok {
  peerMapping[funcType] = append(peerMapping[funcType], v)
 } else {
  logger.Warn(...)
 }
}
```

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+		if err := channel.Verify(); err != nil {
+			return err
+		}
+		channelIDs = append(channelIDs, channel.Name)
+		channelConfigs = append(channelConfigs, channel)
+	}
+	s.channels = channels
+	s.channelIDs = channelIDs
+	s.channelConfigs = channelConfigs
+	for _, channel := range channels {
```
alexandrosfilios, line 154: We could do this in the previous loop as well.

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+
+func (s *Service) VaultTXStoreCacheSize() int {
+	defaultCacheSize := 100
+	v := s.configService.GetString("fabric." + s.prefix + "vault.txidstore.cache.size")
+	cacheSize, err := strconv.Atoi(v)
+	if err != nil {
+		return defaultCacheSize
+	}
+
+	if cacheSize < 0 {
```
alexandrosfilios, line 230: We could unify the two if statements (but it's just a matter of taste)

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+func (s *Service) VaultPersistenceType() string {
+	return s.configService.GetString("fabric." + s.prefix + "vault.persistence.type")
+}
+
+func (s *Service) VaultPersistencePrefix() string {
+	return VaultPersistenceOptsKey
+}
+
+func (s *Service) VaultTXStoreCacheSize() int {
+	defaultCacheSize := 100
```
alexandrosfilios, line 223: This could also be a constant

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+	if cacheSize < 0 {
+		return defaultCacheSize
+	}
+
+	return cacheSize
+}
+
+// DefaultMSP returns the default MSP
+func (s *Service) DefaultMSP() string {
+	return s.configService.GetString("fabric." + s.prefix + "defaultMSP")
```
alexandrosfilios, line 239: This could be
```
return s.GetString("defaultMSP")
```

Similar for the methods above

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+		return err
+	}
+	for _, channel := range channels {
+		if err := channel.Verify(); err != nil {
+			return err
+		}
+		channelIDs = append(channelIDs, channel.Name)
+		channelConfigs = append(channelConfigs, channel)
+	}
+	s.channels = channels
```
alexandrosfilios, line 151: Given how these fields are used I would propose to:
* convert channelConfigs and channels into maps and expose the methods `Channel(name string)` and `ChannelConfig(name string)`.
* remove channelIDs, because it's duplicated information. When you need them you can iterate through the map keys.

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	return v
+}
+
+func (s *Service) BroadcastRetryInterval() time.Duration {
+	return s.configService.GetDuration("fabric." + s.prefix + "ordering.retryInterval")
+}
+
+func (s *Service) OrdererConnectionPoolSize() int {
+	k := "fabric." + s.prefix + "ordering.connectionPoolSize"
```
alexandrosfilios, line 324: We could have a method getIntOrDefault, since we do this many times

platform/fabric/core/generic/config/service.go
```diff
@@ -0,0 +1,360 @@
...
+	if s.configService.IsSet(k) {
+		return s.configService.GetInt(k)
+	}
+	return DefaultOrderingConnectionPoolSize
+}
+
+func (s *Service) SetConfigOrderers(orderers []*grpc.ConnectionConfig) error {
+	s.orderers = append(s.orderers[:len(s.configuredOrderers)], orderers...)
```
alexandrosfilios, line 332: We don't seem to need `s.configuredOrderers`, but only its length.

platform/fabric/core/generic/delivery.go
```diff
@@ -8,69 +8,145 @@ package generic
...
-			if err != nil {
-				logger.Errorf("[%s] unmarshal payload failed: %s", c.ChannelName, err)
-				return false, err
-			}
-			channelHeader, err := protoutil.UnmarshalChannelHeader(payload.Header.ChannelHeader)
-			if err != nil {
-				logger.Errorf("[%s] unmarshal Channel header failed: %s", c.ChannelName, err)
-				return false, err
-			}
+				env, err := protoutil.UnmarshalEnvelope(tx)
```
alexandrosfilios, line 109: The same logic is also under fabric/core/generic/committer/committer.go. We could extract and re-use.

platform/fabric/core/generic/ledger.go
```diff
@@ -8,46 +8,68 @@ package generic
...
+	return transaction.NewProcessedTransaction(pt)
 }
 
-// NewQueryExecutor gives handle to a query executor.
-// A client can obtain more than one 'QueryExecutor's for parallel execution.
-// Any synchronization should be performed at the implementation level if required
-func (c *Channel) NewQueryExecutor() (driver.QueryExecutor, error) {
-	return c.Vault.NewQueryExecutor()
+func (c *Ledger) GetBlockNumberByTxID(txID string) (uint64, error) {
+	res, err := c.ChaincodeManager.Chaincode("qscc").NewInvocation(GetBlockByTxID, c.ChannelName, txID).WithSignerIdentity(
```
alexandrosfilios, line 53: We could use a constant for "qscc" to explain also what it does.

platform/fabric/core/generic/ledger.go
```diff
@@ -8,46 +8,68 @@ package generic
...
+	return &Ledger{ChannelName: channelName, ChaincodeManager: chaincodeManager, LocalMembership: localMembership, ConfigService: configService}
 }
 
-// GetEphemeralRWSet returns an ephemeral RWSet for this ledger whose content is unmarshalled
-// from the passed bytes.
-// If namespaces is not empty, the returned RWSet will be filtered by the passed namespaces
-func (c *Channel) GetEphemeralRWSet(rwset []byte, namespaces ...string) (driver.RWSet, error) {
-	return c.Vault.InspectRWSet(rwset, namespaces...)
+func (c *Ledger) GetTransactionByID(txID string) (driver.ProcessedTransaction, error) {
+	raw, err := c.ChaincodeManager.Chaincode("qscc").NewInvocation(GetTransactionByID, c.ChannelName, txID).WithSignerIdentity(
```
alexandrosfilios, line 35: We could re-use some code to avoid duplication:
```
func (c *Ledger) queryChaincode(function string, param any, result proto.Message) error {
	raw, err := c.ChaincodeManager.Chaincode("qscc").
		NewInvocation(function, c.ChannelName, param).
		WithSignerIdentity(c.LocalMembership.DefaultIdentity()).
		WithEndorsersByConnConfig(c.ConfigService.PickPeer(driver.PeerForQuery)).
		Query()
	if err != nil {
		return err
	}
	
	return proto.Unmarshal(raw, result)
}

func (c *Ledger) GetTransactionByID(txID string) (driver.ProcessedTransaction, error) {
	pt := &peer.ProcessedTransaction{}
	if err := c.queryChaincode(GetTransactionByID, txID, pt); err != nil {
		return nil, err
	}
	return transaction.NewProcessedTransaction(pt)
}

func (c *Ledger) GetBlockNumberByTxID(txID string) (uint64, error) {
	block := &common.Block{}
	if err := c.queryChaincode(GetBlockByTxID, txID, block); err != nil {
		return 0, err
	}
	return block.Header.Number, nil
}

// GetBlockByNumber fetches a block by number
func (c *Ledger) GetBlockByNumber(number uint64) (driver.Block, error) {
	block := &common.Block{}
	if err := c.queryChaincode(GetBlockByNumber, number, block); err != nil {
		return nil, err
	}
	return &Block{Block: block}, nil
}
```

platform/fabric/core/generic/membership/membership.go
```diff
@@ -0,0 +1,90 @@
...
+	ChannelResources channelconfig.Resources
+}
+
+func NewService() *Service {
+	return &Service{}
+}
+
+// Resources returns the active Channel configuration bundle.
+func (c *Service) Resources() channelconfig.Resources {
+	c.ResourcesLock.RLock()
```
alexandrosfilios, line 35: I'm not sure we need to lock here

platform/fabric/core/generic/network.go
```diff
@@ -86,60 +73,15 @@ func (f *Network) Name() string {
...
 	}
 
-	chanQuiet := false
-	for _, chanDef := range f.channelConfigs {
-		if chanDef.Name == name {
-			chanQuiet = chanDef.Quiet
-			break
-		}
-	}
+	chanQuiet := f.ConfigService().IsChannelQuite(name)
```
alexandrosfilios, line 84: `IsChannelQuiet` there is a typo in the method name

platform/fabric/core/generic/ordering/ordering.go
```diff
@@ -62,25 +38,29 @@ type TransactionWithEnvelope interface {
...
+		GetEndorserTransactionService: getEndorserTransactionService,
+		SigService:                    sigService,
+		Metrics:                       metrics,
+		Broadcasters:                  map[string]BroadcastFnc{},
+		BroadcastMutex:                sync.RWMutex{},
+		Broadcaster:                   nil,
 	}
-	s.Broadcasters["BFT"] = NewBFTBroadcaster(network, poolSize, metrics).Broadcast
-	cft := NewCFTBroadcaster(network, poolSize, metrics)
+	s.Broadcasters["BFT"] = NewBFTBroadcaster(configService, metrics).Broadcast
```
alexandrosfilios, line 62: We could use an enum for "BFT", "etcdraft", "solo"

platform/fabric/core/generic/peer.go
```diff
@@ -0,0 +1,143 @@
...
+
+type connCreator struct {
+	ConfigService driver.ConfigService
+	Singer        driver.Signer
+}
+
+func (c *connCreator) NewPeerClientForAddress(cc grpc.ConnectionConfig) (peer2.Client, error) {
+	logger.Debugf("Creating new peer client for address [%s]", cc.Address)
+	var certs [][]byte
+	if cc.TLSEnabled {
```
alexandrosfilios, line 49: We can re-use the code from platform/view/services/grpc

platform/fabric/core/generic/peer.go
```diff
@@ -0,0 +1,143 @@
...
+	override := c.ConfigService.TLSServerHostOverride()
+	clientConfig := &grpc.ClientConfig{}
+	clientConfig.Timeout = c.ConfigService.ClientConnTimeout()
+	if clientConfig.Timeout == time.Duration(0) {
+		clientConfig.Timeout = grpc.DefaultConnectionTimeout
+	}
+
+	secOpts := grpc.SecureOptions{
+		UseTLS:            UseTLS,
+		RequireClientCert: c.ConfigService.TLSClientAuthRequired(),
```
alexandrosfilios, line 94: We could make this
```
RequireClientCert: !UseTLS && c.ConfigService.TLSClientAuthRequired()
```

platform/fabric/core/generic/peer.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		clientConfig.SecOpts.ServerRootCAs = tlsRootCerts
+	}
+
+	clientConfig.KaOpts = grpc.KeepaliveOptions{
+		ClientInterval: c.ConfigService.KeepAliveClientInterval(),
+		ClientTimeout:  c.ConfigService.KeepAliveClientTimeout(),
+	}
+
+	return clientConfig, override, nil
```
alexandrosfilios, line 126: We don't need to return override here. We could get it directly in the caller method, since we don't do any calculations with it.

platform/fabric/core/generic/vault.go
```diff
@@ -7,30 +7,59 @@ SPDX-License-Identifier: Apache-2.0
...
+}
+
+func NewVaultService(vault *vault.Vault) *VaultService {
+	return &VaultService{Vault: vault}
+}
+
+// NewRWSet returns a RWSet for this ledger.
+// A client may obtain more than one such simulator; they are made unique
+// by way of the supplied txid
+func (c *VaultService) NewRWSet(txid string) (driver.RWSet, error) {
```
alexandrosfilios, line 30: Most probably we don't need to define this and the next method.

alexandrosfilios (approved):
LGTM

PR #546, opened 2024-04-11
transaction filter

adecaro:
This PR does the following: It allows both orion and fabric commit pipeline to process transactions that are unknown but are accepted by the custom filters.

### DIFF
--- a/integration/fabric/iou/views/utils.go
+++ b/integration/fabric/iou/views/utils.go
@@ -9,6 +9,7 @@ package views
 import (
 	"sync"
 
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/core"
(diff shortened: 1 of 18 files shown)


platform/orion/core/generic/vault/vault.go
```diff
@@ -241,6 +241,24 @@ func (db *Vault) SetBusy(txid string) error {
 	return nil
 }
 
+func (db *Vault) SetStatus(txID string, code odriver.ValidationCode) error {
```
alexandrosfilios, line 244: Once we merge the race-conditions PR, we need to check that the error is not caused by a unique-constraint violation.

alexanderfilios (approved):
LGTM

PR #547, opened 2024-04-12
common platform

adecaro:
This PR does the following: It introduces a `common` platform that contains shared code between the fabric and the orion platform. This code can be of help for other platforms in the future.
In addition, this PR removes `TxStatusChangeListener` and replace it `FinalityListener`.


platform/fabric/core/generic/committer/em.go
```diff
@@ -0,0 +1,168 @@
...
+	c.EventQueue <- event
+}
+
+func (c *EventManager) Dispatch(event TxEvent) {
+	defer func() {
+		if r := recover(); r != nil {
+			logger.Errorf("caught panic while running dispatching event [%v]: [%s][%s]", event, r, debug.Stack())
+		}
+	}()
+	l := c.cloneListeners(event.TxID)
```
alexandrosfilios, line 91: `l` can be nil

platform/fabric/core/generic/committer/em.go
```diff
@@ -0,0 +1,168 @@
...
+	for _, listener := range l {
+		if err := listener.OnStatus(event.TxID, int(event.ValidationCode), event.ValidationMessage); err != nil {
+			logger.Errorf("failed on status call [%v]: [%s][%s]", event, err, debug.Stack())
+		}
+		c.DeleteListener(event.TxID, listener)
```
alexandrosfilios, line 96: You could delete all listeners the moment you clone them, right? We don't have any revert mechanism and we keep the write lock for O(n^2) time like this while we can do it in O(1).

platform/fabric/core/generic/committer/em.go
```diff
@@ -0,0 +1,168 @@
...
+	for {
+		select {
+		case <-context.Done():
+			return
+		case <-ticker.C:
+			txIDs := c.txIDs()
+			for _, txID := range txIDs {
+				// check txID status, if it is valid or invalid, post an event
+				logger.Debugf("check tx [%s]'s status", txID)
+				vc, message, err := c.Vault.Status(txID)
```
alexandrosfilios, line 128: We could have a `Status` method that accepts multiple txIDs to optimise the query

platform/fabric/core/generic/committer/em.go
```diff
@@ -0,0 +1,168 @@
...
+	} else {
+		var ok bool
+		ls, ok = c.txIDListeners[txID]
+		if !ok {
+			return
+		}
+	}
+	for i, l := range ls {
+		if l == ch {
+			ls = append(ls[:i], ls[i+1:]...)
```
alexandrosfilios, line 74: We need to delete the key if len(ls) == 0. Otherwise it stays in the txIDs() result.

alexanderfilios (approved):
LGTM

PR #549, opened 2024-04-13
add fallback option to use OrdererAddresses

arner:
If the Orderer MSP does not have Endpoints in its channel config, this PR adds a fallback check. Another valid configuration for orderer endpoints is OrdererAddresses in the ChannelConfig, see https://hyperledger-fabric.readthedocs.io/en/latest/config_update.html.

**Orderer addresses.** A list of addresses where clients may invoke the orderer Broadcast and Deliver functions. The peer randomly chooses among these addresses and fails over between them for retrieving blocks.

The OrdererAddresses do not have TLSRootCerts. The code only uses the OrdererAddresses key if there is one OrdererMSP in the channel, so that it can use the root certs from there.

### DIFF
--- a/platform/fabric/core/generic/committer/committer.go
+++ b/platform/fabric/core/generic/committer/committer.go
@@ -790,19 +790,33 @@ func (c *Service) applyBundle(bundle *channelconfig.Bundle) error {
 
 	// update the list of orderers
 	ordererConfig, exists := c.MembershipService.ChannelResources.OrdererConfig()
-	if exists {
-		logger.Debugf("[Channel: %s] Orderer config has changed, updating the list of orderers", c.ChannelConfig.ID())
+	if !exists {
+		logger.Infof("no orderer configuration found in Channel config")
+		return nil
+	}
+	logger.Debugf("[Channel: %s] Orderer config has changed, updating the list of orderers", c.ChannelConfig.ID())
 
-		var newOrderers []*grpc.ConnectionConfig
-		orgs := ordererConfig.Organizations()
-		for _, org := range orgs {
-			msp := org.MSP()
-			var tlsRootCerts [][]byte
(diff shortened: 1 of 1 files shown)


adecaro:
Hi @arner , thanks for the PR. In which condition, no Endpoints are defined? Maybe I'm missing something, are these the endpoints of the orderers? If yes, how can the list be empty?

arner:
This happened to me when installing open source Fabric with the ansible scripts (https://github.com/hyperledger-labs/fabric-ansible-collection) which use the Fabric Console and Fabric Operator, all Hyperledger Labs projects. The following field is present in the test network k8s of Fabric Samples but not in the network I deployed with ansible:

```
channel_group > groups > Orderer > groups > OrdererOrg > values

                                                    "Endpoints": {
                                                        "mod_policy": "Admins",
                                                        "value": {
                                                            "addresses": [
                                                                "org0-orderer1.test-network.svc.cluster.local:6050",
                                                                "org0-orderer2.test-network.svc.cluster.local:6050",
                                                                "org0-orderer3.test-network.svc.cluster.local:6050"
                                                            ]
                                                        },
                                                        "version": "0"
                                                    },
```

And the following is present in the ansible deployed network but not in the testnetwork:
```
                                    "OrdererAddresses": {
                                        "mod_policy": "/Channel/Orderer/Admins",
                                        "value": {
                                            "addresses": [
                                                "fabric-orderingservicenode1-orderer.mynetwork.example.com:443"
                                            ]
                                        },
                                        "version": "0"
                                    }
```

Fabric documentation is not clear about whether either of them are mandatory or expected, but at least it describes the OrdererAddresses field. Between all the different components and different Fabric versions, having this fallback seemed like a clean solution to me.

adecaro:
Thanks for the clarification, @arner 👍

adecaro (approved):
LGTM

PR #555, opened 2024-04-22
Allow to start a toplogy without FSC nodes

mbrandenburger:
This PR allows to run topologies even without FSC nodes, which is very helpful when developing nwo platforms and extensions.

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -186,6 +186,7 @@ func (i *Infrastructure) Start() {
 }
 
 func (i *Infrastructure) Stop() {
+	logger.Infof("stopping ...")
 	if i.NWO == nil {
 		panic("call generate or load first")
 	}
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

integration/integration.go
```diff
@@ -293,24 +293,28 @@ func (i *Infrastructure) initNWO() {
...
 		default:
+			logger.Infof("Register %s", label)
 			factory, ok := i.PlatformFactories[label]
 			Expect(ok).To(BeTrue(), "expected to find platform [%s]", label)
 			platforms = append(platforms, factory.New(i.Ctx, topology, i.BuildServer.Client()))
 		}
 	}
-	// Add FSC platform
-	fcsPlatform := i.PlatformFactories["fsc"].New(i.Ctx, fscTopology, i.BuildServer.Client()).(*fsc.Platform)
-	platforms = append(platforms, fcsPlatform)
```
alexandrosfilios, line 306: I think the point here was to have fscPlatform as the last one maybe

integration/integration.go
```diff
@@ -293,24 +293,28 @@ func (i *Infrastructure) initNWO() {
 		switch label {
 		case "fsc":
 			// treat fsc as special
-			fscTopology = topology
-			continue
+			factory, ok := i.PlatformFactories["fsc"]
+			Expect(ok).To(BeTrue(), "expected to find platform [%s]", label)
```
alexandrosfilios, line 297: do we need this assertion? I think it should work without FSC factories (though I don't remember whether we have any such tests)

integration/integration.go
mbrandenburger: ah good point

integration/integration.go
mbrandenburger: true - a few lines before we register fsc factory in case it does not exist

PR #559, opened 2024-04-23
Upgrade to go 1.21

mbrandenburger:


adecaro (commented):
shall we update also the go.mod under tools?

Issue #568, opened 2024-05-08
FSC panics if web is not enabled


arner:
Because the handler is nil.

```
6:40:34.601 [fsc] Start INFO Starting sdks...
2024/05/08 16:43:05 Start triggered panic: runtime error: invalid memory address or nil pointer dereference
goroutine 1 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.22.1/libexec/src/runtime/debug/stack.go:24 +0x64
github.com/hyperledger-labs/fabric-smart-client/pkg/node.(*node).Start.func1()
	/Users/arne/go/pkg/mod/github.com/hyperledger-labs/fabric-smart-client@v0.3.1-0.20240508105356-0c52e84ffe0b/pkg/node/node.go:106 +0x4c
panic({0x102cceec0?, 0x103fad660?})
	/opt/homebrew/Cellar/go/1.22.1/libexec/src/runtime/panic.go:770 +0xf0
github.com/hyperledger-labs/fabric-smart-client/platform/view/services/server/web.(*HttpHandler).RegisterURI(0x0, {0x10264d0dc, 0xd}, {0x10260fe25, 0x3}, {0x103011550, 0x14000d84000})
	/Users/arne/go/pkg/mod/github.com/hyperledger-labs/fabric-smart-client@v0.3.1-0.20240508105356-0c52e84ffe0b/platform/view/services/server/web/handler.go:75 +0xd8
github.com/hyperledger-labs/fabric-smart-client/platform/view/services/server/web.(*Dispatcher).WireViewCaller(0x14000d84000, {0x103007e40, 0x140009ea030})
	/Users/arne/go/pkg/mod/github.com/hyperledger-labs/fabric-smart-client@v0.3.1-0.20240508105356-0c52e84ffe0b/platform/view/services/server/web/dispatcher.go:47 +0x90
github.com/hyperledger-labs/fabric-smart-client/platform/view/services/server/web.InstallViewHandler({0x103021c18, 0x140009689f0}, 0x140009ea000, 0x0)
	/Users/arne/go/pkg/mod/github.com/hyperledger-labs/fabric-smart-client@v0.3.1-0.20240508105356-0c52e84ffe0b/platform/view/services/server/web/view.go:57 +0x264
github.com/hyperledger-labs/fabric-smart-client/platform/view/sdk.(*SDK).installViewHandler(0x14000c36000)
[...]

PR #569, opened 2024-05-08
fix bug web service

adecaro:
This PR does the following: It fixes the bug reported in #568 .
The integration tests by default have the web server disabled. Only the `pingpong` test uses it.
The web server can be enabled by setting `WebEnabled` to `true` on the fsc topology.


alexandrosfilios (approved):
LGTM

PR #571, opened 2024-05-17
Replace deprecated logging param in nwo

mbrandenburger:
Currently, NWO uses the deprecated --logging-level param when invoking the peer cli. This commit replaces this param with FABRIC_LOGGING_SPEC env variable.

### DIFF
--- a/integration/nwo/fsc/fsc.go
+++ b/integration/nwo/fsc/fsc.go
@@ -550,7 +550,7 @@ func (p *Platform) fscNodeCommand(node *node2.Replica, command common.Command, t
 		cmd.Args = append(cmd.Args, "--keyfile", keyfilePath)
 	}
 
-	cmd.Args = append(cmd.Args, "--logging-level", p.Topology.Logging.Spec)
+	cmd.Env = append(cmd.Env, fmt.Sprintf("FSCNODE_LOGGING_SPEC=%s", p.Topology.Logging.Spec))
 
 	return cmd
 }
--- a/cmd/fsccli/main.go
+++ b/cmd/fsccli/main.go
@@ -33,14 +33,6 @@ func main() {
 	replacer := strings.NewReplacer(".", "_")
 	viper.SetEnvKeyReplacer(replacer)
 
-	// Define command-line flags that are valid for all peer commands and
-	// subcommands.
-	mainFlags := mainCmd.PersistentFlags()
-
-	mainFlags.String("logging-level", "", "Legacy logging level flag")
-	viper.BindPFlag("logging_level", mainFlags.Lookup("logging-level"))
-	mainFlags.MarkHidden("logging-level")
-
 	mainCmd.AddCommand(artifactgen.NewCmd())
 	mainCmd.AddCommand(cryptogen.NewCmd())
 	mainCmd.AddCommand(view.NewCmd())
--- a/integration/nwo/cmd/main.go
+++ b/integration/nwo/cmd/main.go
@@ -44,14 +44,6 @@ func (m *Main) Execute() {
 	replacer := strings.NewReplacer(".", "_")
 	viper.SetEnvKeyReplacer(replacer)
 
-	// Define command-line flags that are valid for all peer commands and
-	// subcommands.
-	mainFlags := m.mainCmd.PersistentFlags()
-
-	mainFlags.String("logging-level", "", "Legacy logging level flag")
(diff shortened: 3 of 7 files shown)


integration/nwo/fsc/fsc.go
```diff
@@ -550,7 +550,7 @@ func (p *Platform) fscNodeCommand(node *node2.Replica, command common.Command, t
 		cmd.Args = append(cmd.Args, "--keyfile", keyfilePath)
 	}
 
-	cmd.Args = append(cmd.Args, "--logging-level", p.Topology.Logging.Spec)
+	cmd.Env = append(cmd.Env, fmt.Sprintf("FABRIC_LOGGING_SPEC=%s", p.Topology.Logging.Spec))
```
adecaro, line 553: I think we can remove this completely. This code is related to the execution of the FSC node.

adecaro (changes requested):
LGTM

integration/nwo/fsc/fsc.go
mbrandenburger: don't we also use `flogging` within the FSC nodes?

mbrandenburger: If we remove this, how do we pass the FSC node logging level when running it via nwo?

adecaro: yes, we do use the FSC's flogging. But the effect of setting FABRIC_LOGGING_SPEC is to have viper overridding the content of the key `fabric.logging.spec`. The FSC nodes uses the key `fsc.logging.spec` instead. Indeed, if you look at the code, a few lines before we set FSC_LOGGING_SPEC

adecaro: actually, we have a bug there. The code at line 535 has
```go
	cmd.Env = append(cmd.Env, "FSCNODE_LOGGING_SPEC="+p.Topology.Logging.Spec)
``` 
it should be
```go
	cmd.Env = append(cmd.Env, "FSC_LOGGING_SPEC="+p.Topology.Logging.Spec)
```

adecaro: my explanation is not completely correct.
This is the code that initializes the FSC logger:

```go
func (p *provider) load() error {
	p.v = viper.New()
	err := p.initViper(p.v, CmdRoot)
	if err != nil {
		return err
	}

	err = p.v.ReadInConfig() // Find and read the config file
	if err != nil {          // Handle errors reading the config file
		// The version of Viper we use claims the config type isn't supported when in fact the file hasn't been found
		// Display a more helpful message to avoid confusing the user.
		if strings.Contains(fmt.Sprint(err), "Unsupported Config Type") {
			return errors.Errorf("Could not find config file. "+
				"Please make sure that FSCNODE_CFG_PATH is set to a path "+
				"which contains %s.yaml", CmdRoot)
		} else {
			return errors.WithMessagef(err, "error when reading %s config file", CmdRoot)
		}
	}

	// read in the legacy logging level settings and, if set,
	// notify users of the FSCNODE_LOGGING_SPEC env variable
	var loggingLevel string
	if p.v.GetString("logging_level") != "" {
		loggingLevel = p.v.GetString("logging_level")
	} else {
		loggingLevel = p.v.GetString("logging.level")
	}
	if loggingLevel != "" {
		logger.Warning("CORE_LOGGING_LEVEL is no longer supported, please use the FSCNODE_LOGGING_SPEC environment variable")
	}

	loggingSpec := os.Getenv("FSCNODE_LOGGING_SPEC")
	loggingFormat := os.Getenv("FSCNODE_LOGGING_FORMAT")

	if len(loggingSpec) == 0 {
		loggingSpec = p.v.GetString("logging.spec")
	}
	if len(loggingFormat) == 0 {
		loggingFormat = p.v.GetString("logging.format")
	}

	flogging.Init(flogging.Config{
		Format:  loggingFormat,
		Writer:  logOutput,
		LogSpec: loggingSpec,
	})

	return nil
}
```

adecaro: maybe, you can take the chance and cleanup this code. What do you think?

mbrandenburger: ok, I've removed the legacy logging level flag and pass now `FSCNODE_LOGGING_SPEC` via nwo.

adecaro (approved):
LGTM

PR #574, opened 2024-05-23
Cache fix

alexandrosfilios:
This PR addresses two issues detected by the integration tests of Token SDK:
* When two nodes try to set the state at the same time, one gets a duplicate error and doesn't update the vault. Then we don't fail, because we know that another process has updated it, but the txidstore cache remains with the same previous value. Now we invalidate the cache and fetch the new value.
* When multiple replicas are trying to commit the same TX and rewrite the same keys, we get a deadlock error from the DB. Now we retry every time we get a deadlock error until the operation is successful. This means that either the other writes have failed so we can write the new values, or the other writes were successful so we update the state (using the same values).


adecaro (approved):
LGTM

PR #589, opened 2024-06-13
SchemaManager handles options for Verification and KeyImport

ale-linux:
Use the `SerializedIdemixIdentity` proto definition from the idemix repo and not from fabric, whence idemix has been removed.


adecaro:
The code has been ported to the token-sdk

PR #597, opened 2024-06-17
Performance

alexandrosfilios:


platform/view/core/manager/manager.go
```diff
@@ -183,25 +183,22 @@ func (cm *manager) Initiate(id string) (interface{}, error) {
...
-	// Create the context
-	cm.contextsSync.Lock()
-	ctx := cm.ctx
-	cm.contextsSync.Unlock()
-	if ctx == nil {
-		ctx = context.Background()
-	}
-	viewContext, err := NewContextForInitiator("", ctx, cm.sp, cm.commLayer, cm.endpointService, id, view, cm.viewTracer)
+func (cm *manager) InitiateViewWithIdentity(view view.View, id view.Identity, ctx context.Context) (interface{}, error) {
+	newCtx, span := cm.viewTracer.Start(ctx, "initiate_view_with_identity")
```
adecaro, line 194: could we start this trace by adding to the name also the ID of the viewContext?

adecaro (approved):
LGTM. I had just a comment.

platform/view/core/manager/manager.go
alexandrosfilios: We can add it as an attribute. Let's take this up in a separate PR

PR #604, opened 2024-06-21
Support for traces in sessions

alexandrosfilios:


platform/view/view/context.go
```diff
@@ -72,6 +76,12 @@ type MutableContext interface {
 
 // Context gives a view information about the environment in which it is in execution
 type Context interface {
+	// StartSpanFrom creates a new child span from the passed context
+	StartSpanFrom(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span)
```
adecaro, line 80: can we move these two functions to a different interface that is imported by Context?

platform/view/view/session.go
```diff
@@ -23,6 +24,7 @@ type Message struct {
 	FromPKID     []byte // PK identifier of the caller
 	Status       int32  // Message Status (OK, ERROR)
 	Payload      []byte // Payload
+	Ctx          context.Context
```
adecaro, line 27: is this guaranteed to be serializable?

platform/view/view/session.go
```diff
@@ -53,9 +55,15 @@ type Session interface {
 	// Send sends the payload to the endpoint
 	Send(payload []byte) error
 
+	// SendWithContext sends the payload to the endpoint with the passed context
```
adecaro, line 58: does this mean that the context is part of the message or the context is used to control the send operation itself?

platform/view/view/session.go
alexandrosfilios: it isn't, but it doesn't have to be

platform/view/view/session.go
alexandrosfilios: no it's part of the message. it carries some information. in this case the span context

adecaro: I would clarify in the comment then

alexandrosfilios: No but the context doesn't have to do only with the span. The span uses the context but the context has more functions in general, e.g. canceling the context, or adding baggage values

adecaro (approved):
LGTM

PR #605, opened 2024-06-27
allow a second registration of the same fabric driver

arner:
In our project, we have  integration tests that create two FSC nodes in a single process. It's a nice and quick way to start them. Since the new dig dependency injection, our test panics because the Fabric SDK is registered twice. This PR fixes that by just warning instead of panicking. But is this the way to go?

(Also, I'd be interested to know whether the driver registration + dig might be doing the same thing in two different ways? Do we need both?)


arner:
@adecaro, @alexandrosfilios, what do you think of this?

adecaro:
@arner , I'll close this given the PR that should fix the same issue. Feel free to reopen, if needed.

PR #611, opened 2024-07-02
improve logs

arner:


platform/fabric/sdk/dig/sdk.go
```diff
@@ -190,17 +190,17 @@ func registerRWSetLoaderHandlerProviders(in struct {
 	for _, network := range in.CoreConfig.Names() {
 		fsn, err := in.FSNProvider.FabricNetworkService(network)
 		if err != nil {
-			return fmt.Errorf("could not find network service for %s", network)
+			return fmt.Errorf("could not find network service for %s: %w", network, err)
```
adecaro, line 193: Please, Arne, given that you are changing this, could replace `fmt` with the package `github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors`. Thanks.

platform/view/services/db/driver/sql/sqlite/persistence.go
```diff
@@ -25,7 +25,7 @@ const sqlitePragmas = `
 
 const driverName = "sqlite"
 
-var logger = flogging.MustGetLogger("postgres-db")
+var logger = flogging.MustGetLogger("sqlite-db")
```
adecaro, line 28: Please, can you add the prefix `view-sdk.db.<the rest>`, so we can catch it more easily when setting the logs level.

adecaro (approved):
LGTM, please, just address the comments left.

PR #618, opened 2024-07-06
Fix data race errors 

saoussen:


platform/common/core/generic/vault/interceptor.go
```diff
@@ -345,7 +348,7 @@ func (i *Interceptor[V]) Equals(other interface{}, nss ...string) error {
 
 func (i *Interceptor[V]) Done() {
 	i.Logger.Debugf("Done with [%s], closed [%v]", i.TxID, i.Closed)
-	if !i.Closed {
+	if i.IsClosed() {
```
arner, line 351:
```suggestion
	if !i.IsClosed() {
```

arner (changes requested):

platform/view/services/comm/p2p.go
```diff
@@ -247,13 +247,19 @@ func (s *streamHandler) send(msg proto.Message) error {
 	return s.writer.WriteMsg(msg)
 }
 
+func (s *streamHandler) isStopping() bool {
+	s.node.streamsMutex.Lock()
```
adecaro, line 251: can we use a read lock here, please?

platform/common/core/generic/vault/interceptor.go
```diff
@@ -365,6 +368,8 @@ func (i *Interceptor[V]) Reopen(qe VersionedQueryExecutor) error {
 }
 
 func (i *Interceptor[V]) IsClosed() bool {
+	i.Mutex.Lock()
```
adecaro, line 371: please, let's use a read lock here. We need also a lock when we set Closed to false.

adecaro (changes requested):

platform/common/core/generic/vault/interceptor.go
```diff
@@ -344,27 +347,32 @@ func (i *Interceptor[V]) Equals(other interface{}, nss ...string) error {
...
 }
 
 func (i *Interceptor[V]) Reopen(qe VersionedQueryExecutor) error {
-	i.Logger.Debugf("Reopen with [%s], closed [%v]", i.TxID, i.Closed)
-	if !i.Closed {
+	i.Logger.Debugf("Reopen with [%s], closed [%v]", i.TxID, i.IsClosed())
+	if !i.IsClosed() {
 		return errors.Errorf("already open")
 	}
 	i.QE = qe
```
adecaro, line 366: should we protect also QE? We might have race conditions on that as well. It might happen that Done is called on QE while a GetState is happening at the same time, no?

adecaro:
Hi @saoussen , please, can you rebase this PR. The communication layer has changed a lot. I just want to make sure your changes are compatible with what exists on main. Thanks :)

arner:
@adecaro I made the requested changes

adecaro (approved):
LGTM

PR #619, opened 2024-07-08
Remove race conditions and simplify endpoint service constructor

arner:
This PR removes a few race conditions and simplifies the constructor for the endpoint service.

### DIFF
--- a/platform/view/core/endpoint/endpoint.go
+++ b/platform/view/core/endpoint/endpoint.go
@@ -32,6 +32,7 @@ type Resolver struct {
 	Addresses      map[driver.PortName]string
 	Aliases        []string
 	PKI            []byte
+	PKILock        sync.RWMutex
 	Id             []byte
 	IdentityGetter func() (view.Identity, []byte, error)
 }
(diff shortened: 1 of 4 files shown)


platform/view/core/endpoint/endpoint.go
```diff
@@ -15,7 +15,6 @@ import (
 
 	"golang.org/x/exp/slices"
```
adecaro, line 17: please, remove the blank line

platform/view/core/endpoint/endpoint.go
```diff
@@ -241,20 +237,20 @@ func (r *Service) SetPublicKeyIDSynthesizer(publicKeyIDSynthesizer driver.Public
 }
 
 func (r *Service) pkiResolve(resolver *Resolver) []byte {
-	r.pkiExtractorsLock.RLock()
+	resolver.PKILock.Lock()
```
adecaro, line 240: can we have a read lock here and if len(resolver.PKI) == 0 then we get a write lock?

adecaro (changes requested):

platform/view/core/endpoint/endpoint.go
```diff
@@ -241,20 +236,22 @@ func (r *Service) SetPublicKeyIDSynthesizer(publicKeyIDSynthesizer driver.Public
 }
 
 func (r *Service) pkiResolve(resolver *Resolver) []byte {
-	r.pkiExtractorsLock.RLock()
+	resolver.PKILock.RLock()
 	if len(resolver.PKI) != 0 {
-		r.pkiExtractorsLock.RUnlock()
```
adecaro, line 246: missing read unlock

platform/view/core/endpoint/endpoint.go
```diff
@@ -241,20 +236,22 @@ func (r *Service) SetPublicKeyIDSynthesizer(publicKeyIDSynthesizer driver.Public
...
 		return resolver.PKI
 	}
-	r.pkiExtractorsLock.RUnlock()
+	resolver.PKILock.RUnlock()
 
+	resolver.PKILock.Lock()
+	defer resolver.PKILock.Unlock()
 	resolver.PKI = r.ExtractPKI(resolver.Id)
```
adecaro, line 247: check again if resolver.PKI is different from nil. Another thread might have already set it.

adecaro (changes requested):

platform/view/services/comm/host/rest/websocket/stream_test.go
```diff
@@ -104,19 +105,24 @@ func TestReader(t *testing.T) {
 	for _, message := range input {
 		assert.NoError(t, conn.ReadValue(message))
 	}
+	wg := sync.WaitGroup{}
```
adecaro, line 108: why the changes in this file? It seems unrelated to the rest.

adecaro (approved):
LGTM modulo a clarification.

platform/view/services/comm/host/rest/websocket/stream_test.go
arner: This also caused the tests to fail when run with `-race`.

PR #620, opened 2024-07-09
Added traces

alexandrosfilios:


platform/view/sdk/dig/sdk.go
```diff
@@ -121,7 +121,7 @@ func (p *SDK) Install() error {
 		p.C.Provide(operations.NewOperationSystem),
 		p.C.Provide(view3.NewResponseMarshaler, dig.As(new(view3.Marshaller))),
 		p.C.Provide(func(o *operations.Options, l operations.OperationsLogger) metrics2.Provider {
-			return operations.NewMetricsProvider(o.Metrics, l, false)
+			return operations.NewMetricsProvider(o.Metrics, l, true)
```
adecaro, line 124: why skipping the error?

platform/view/services/tracing/span.go
```diff
@@ -102,8 +102,8 @@ func newSpan(backingSpan trace.Span, labelNames []LabelName, operations metrics.
 }
 
 func defaultNow(t time.Time) time.Time {
-	if t.IsZero() {
-		return time.Now()
-	}
-	return t
+	//if t.IsZero() {
```
adecaro, line 105: forget something here?

adecaro (approved):
LGTM. Please, check the comments though.

platform/view/services/tracing/span.go
alexandrosfilios: It is never the case so far, but yes I should fix this

platform/view/sdk/dig/sdk.go
alexandrosfilios: Because it happens to register the same metric multiple times. It should just ignore it and skip the double registration.

PR #621, opened 2024-07-09
hide pkcs11 behind a build tag so that fsc is pure go by default

arner:
The `github.com/hyperledger/fabric/bccsp/pkcs11` dependency is the only reason we need CGO_ENABLED=1 when building an application that imports FSC. It makes things like cross-architecture builds unnecessarily difficult.

This PR makes it so that if you want to use pkcs11, you have to build the binary with `go build -tags pkcs11`. It includes a default implementation that panics if the user configures HSM in a binary that's not built with `pkcs11` as a build tag.

### DIFF
--- a/docs/core-fabric.md
+++ b/docs/core-fabric.md
@@ -245,7 +245,8 @@ fabric:
               Hash: SHA2
               Security: 256
             # Definition of PKCS11 configuration parameters when using a Hardware HSM
-            # Only needs to be defined if the BCCSP Default is set to PKCS11
+            # Only needs to be defined if the BCCSP Default is set to PKCS11.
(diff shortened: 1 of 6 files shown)


adecaro (commented):
So, can we test this PR against the token-sdk? The token-sdk has a few integration tests where pkcs11 is used. I would like to make sure this works there too.

arner:
> So, can we test this PR against the token-sdk? The token-sdk has a few integration tests where pkcs11 is used. I would like to make sure this works there too.

This branch uses the current FSC commit and the code for FTS to support pkcs11 too: https://github.com/hyperledger-labs/fabric-token-sdk/actions/runs/10175721245?pr=721. As far as I can tell, the only failing tests are due to flakiness, but HSM works.

adecaro (approved):
LGTM

PR #638, opened 2024-07-24
Responder views inherit span of initiator. Span names are prefixed wi…

alexandrosfilios:
…th the node name. Added further spans/events.


adecaro (approved):
LGTM

PR #641, opened 2024-07-30
improvements

adecaro:
- Communication layer: quieter logs;
- DB: support for transactional unversioned persistance;
- DB: memory driver replaced by sqlite in memory;
- KVS: support for concurrent put and delete;
- Expanded errors package.


platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
```diff
@@ -186,15 +186,15 @@ func (c *multiplexedClientConn) readIncoming() {
...
+		logger.Debugf("client connection closed: %v", err)
 	}()
 	var mm MultiplexedMessage
 	for {
 		//c.writeMu.Lock()
 		err := c.ReadJSON(&mm)
 		//c.writeMu.Unlock()
 		if err != nil {
-			logger.Warnf("Client connection errored: %v", err)
+			logger.Debugf("Client connection errored: %v", err)
```
arner, line 197: for consistency, maybe all of them should start with lowercase? (also on line 207 and 209)

platform/view/services/db/driver/sql/driver.go
```diff
@@ -116,6 +127,6 @@ func getOps(config driver.Config) (common.Opts, error) {
 
 func getTableName(prefix, name string) (table string, valid bool) {
 	table = fmt.Sprintf("%s_%s", prefix, name)
-	r := regexp.MustCompile("^[a-zA-Z_]+$")
+	r := regexp.MustCompile("^[a-zA-Z_0-9]+$")
```
arner, line 130: Numbers are allowed in SQL, but not at the start of the table name (unless you quote it). We can allow numbers if we need them, but maybe the regex should block strings that start with a number then.

platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
adecaro: done

platform/view/services/db/driver/sql/driver.go
adecaro: I have introduced a new function that generates UUID without digits: GenerateUUIDOnlyLetters

platform/common/core/generic/vault/helpers.go
```diff
@@ -336,7 +336,7 @@ func TTestQueryExecutor(t *testing.T, ddb VersionedPersistence, vp artifactsProv
 	}
 	var expected = RemoveNils([]VersionedRead{
 		{Key: "k1", Raw: []byte("k1_value"), Block: 35, TxNum: 3},
-		{Key: "k5"},
+		//{Key: "k5"},
```
arner, line 339: should we remove this instead of commenting out?

platform/view/services/db/dbtest/helpers.go
```diff
@@ -595,6 +596,28 @@ func TTestMultiWritesAndRangeQueries(t *testing.T, db driver.TransactionalVersio
...
+	n := 20
+	wg.Add(n)
+	for i := 0; i < n; i++ {
+		go func(i int) {
+			write(
+				t,
+				db,
+				ns,
+				fmt.Sprintf("TTestMultiWrites_key_%d", i),
+				[]byte(fmt.Sprintf("TTestMultiWrites_vallue_%d", i)),
```
arner, line 611: value*

platform/view/services/kvs/kvs.go
```diff
@@ -123,33 +123,36 @@ func (o *KVS) Exists(id string) bool {
...
+		if err1 := tx.Discard(); err1 != nil {
 			logger.Debugf("got error %v; discarding caused %v", err, err1)
 		}
 
 		if !errors.HasCause(err, driver.UniqueKeyViolation) {
 			return errors.Wrapf(err, "failed to commit value for id [%s]", id)
 		}
 	} else {
-		if err := o.store.Commit(); err != nil {
+		if err := tx.Commit(); err != nil {
```
arner, line 146: We're only doing a single SetState in this transaction. Why do we need a TransactionalUnversionedPersistence? I noticed we do it in Delete as well, and also in NewSimpleTXIDStore and in the vault's SetStatus. Wouldn't it be better to just fire the SetState directly in those cases?

platform/view/services/kvs/kvs.go
```diff
@@ -195,29 +198,27 @@ func (o *KVS) Delete(id string) error {
...
-
-	if err := o.store.BeginUpdate(); err != nil {
+	tx, err := o.store.NewWriteTransaction()
+	if err != nil {
 		return errors.Wrapf(err, "begin update for id [%s] failed", id)
 	}
-
-	if err := o.store.DeleteState(o.namespace, id); err != nil {
-		if err1 := o.store.Discard(); err1 != nil {
+	if err := tx.DeleteState(o.namespace, id); err != nil {
```
arner, line 205: same as above

platform/view/services/kvs/kvs.go
adecaro: it is to allow parallel puts

platform/common/core/generic/vault/helpers.go
adecaro: will do

platform/view/services/db/dbtest/helpers.go
adecaro: will do

platform/view/services/kvs/kvs.go
arner: As discussed offline, I don't think it's necessary for the sql drivers, but we'll handle it in a new issue / PR.

arner (approved):
LGTM, with the sidenote that we'll open a new issue for single-shot updates instead of opening transactions for single put and deletes.

pkg/utils/uuid.go
```diff
@@ -35,6 +36,26 @@ func GenerateUUID() string {
...
+	uuidStr := strings.ReplaceAll(uuidWithDigits, "-", "")
+	replaceDigits := func(r rune) rune {
+		if r >= '0' && r <= '9' {
+			// Generate a random letter between A-Z or a-z
+			var b [1]byte
+			_, err := rand.Read(b[:])
+			if err != nil {
+				panic(err)
+			}
+			return rune('a' + (b[0] % 26))
```
alexandrosfilios, line 52: We want A-Z, too, so maybe we need to do %52?

platform/view/services/db/driver/memory/driver.go
```diff
@@ -7,27 +7,51 @@ SPDX-License-Identifier: Apache-2.0
...
+	if err != nil {
+		return nil, err
+	}
+	return &unversioned.Transactional{TransactionalVersioned: backend}, nil
 }
 
-func (v *Driver) NewVersioned(string, driver.Config) (driver.VersionedPersistence, error) {
-	return NewVersionedPersistence(), nil
+func (d *Driver) NewTransactionalVersioned(dataSourceName string, config driver.Config) (driver.TransactionalVersionedPersistence, error) {
+	return sql.NewPersistence(strings.ReplaceAll(utils.GenerateUUIDOnlyLetters(), "-", "_"), opts, sql.VersionedConstructors)
```
alexandrosfilios, line 52: This method is not supposed to return any hyphens i think

platform/view/services/db/driver/badger/badger.go
```diff
@@ -297,10 +303,34 @@ func (w *WriteTransaction) SetState(namespace driver2.Namespace, key string, val
...
+	err := w.txn.Delete([]byte(dbKey))
+	if err != nil {
+		return errors.Wrapf(err, "could not delete value for key %s", dbKey)
+	}
+
+	return nil
+}
+
 func (w *WriteTransaction) Commit() error {
+	//if err := w.retryRunner.RunWithErrors(func() (bool, error) {
```
alexandrosfilios, line 318: We can remove the comment?

platform/view/services/db/driver/badger/badger.go
```diff
@@ -297,10 +303,34 @@ func (w *WriteTransaction) SetState(namespace driver2.Namespace, key string, val
...
+	//	}
+	//	return true, err
+	//}); err != nil {
+	//	return errors.Wrap(err, "failed to commit")
+	//}
+
 	err := w.txn.Commit()
-	if err != nil {
-		return errors.Wrap(err, "could not commit transaction")
+	if err == nil {
```
alexandrosfilios, line 332: You mean `!=` ?

platform/view/services/db/driver/badger/badger.go
adecaro: yes

platform/view/services/db/driver/badger/badger.go
adecaro: yes

pkg/utils/retry.go
```diff
@@ -39,7 +43,22 @@ func NewRetryRunner(maxTimes int, delay time.Duration, expBackoff bool) *retryRu
...
+		expBackoff:    expBackoff,
+		maxTimes:      maxTimes,
+		probabilistic: true,
+		interval:      interval,
+		logger:        flogging.MustGetLogger("retry-runner"),
+	}
+}
+
 func (f *retryRunner) nextDelay() time.Duration {
+	if f.probabilistic && f.delay == 0 {
```
alexandrosfilios, line 59: We could have this in the constructor?

pkg/utils/retry.go
```diff
@@ -39,7 +43,22 @@ func NewRetryRunner(maxTimes int, delay time.Duration, expBackoff bool) *retryRu
 	}
 }
 
+// NewProbabilisticRetryRunner returns a new runner that sets delay to time.Duration(rand.Int63n(f.interval)+1) * time.Millisecond
+func NewProbabilisticRetryRunner(maxTimes int, interval int64, expBackoff bool) *retryRunner {
```
alexandrosfilios, line 47: We could use `time.Duration` as `interval`?

platform/view/services/kvs/kvs.go
```diff
@@ -128,43 +129,31 @@ func (o *KVS) Put(id string, state interface{}) error {
...
+		logger.Debugf("store [%d] bytes into key [%s:%s]", len(raw), o.namespace, id)
+		if err := tx.SetState(o.namespace, id, raw); err != nil {
 			if err1 := tx.Discard(); err1 != nil {
 				logger.Debugf("got error %v; discarding caused %v", err, err1)
 			}
-			return errors.Wrapf(err, "committing value for id [%s] failed", id)
+
+			if !errors.HasCause(err, driver.UniqueKeyViolation) {
+				return false, errors.Wrapf(err, "failed to commit value for id [%s]", id)
+			}
```
alexandrosfilios, line 145: We could do `return true, nil` here and avoid the else nesting.

platform/common/core/generic/vault/db/helpers.go
```diff
@@ -33,7 +33,7 @@ func OpenMemory() (*db.UnversionedPersistence, error) {
 }
 
 func openMemory[V any](open opener[V]) (V, error) {
-	return open(&mem.Driver{}, "", nil)
+	return open(&mem.Driver{}, "memory", nil)
```
alexandrosfilios, line 36: Can you use `mem.MemoryPersistence` just to keep track of where we use this constant.

platform/view/services/db/driver/unversioned/unversioned_test.go
```diff
@@ -150,9 +150,9 @@ func TestSimpleReadWriteBadger(t *testing.T) {
 func TestSimpleReadWriteMemory(t *testing.T) {
 	c := &mocks.Config{}
 	c.UnmarshalKeyReturns(nil)
-	db, err := db.Open(&mem.Driver{}, "", c)
-	defer db.Close()
+	db, err := db.Open(&mem.Driver{}, "memory", c)
```
alexandrosfilios, line 153: `mem.MemoryPersistence`

alexandrosfilios (approved):
LGTM modulo comments

platform/common/core/generic/vault/db/helpers.go
adecaro: yes

platform/view/services/db/driver/unversioned/unversioned_test.go
adecaro: yes

platform/view/services/kvs/kvs.go
adecaro: okay

pkg/utils/retry.go
adecaro: interval is used to compute delay as described in the comments. So, it is a different beast, I think. No?

pkg/utils/retry.go
adecaro: I wanted to avoid to put this in the contractor because if there is a success at the first trial, then we don't need to sample randomness that might be expensive.

pkg/utils/retry.go
alexandrosfilios: No instead of passing the milliseconds as parameter (lets say `5`), you can pass `5 * time.Millisecond` that is more readable

pkg/utils/retry.go
alexandrosfilios: This check is performed every time we call `nextDelay`, so several times per `Run`. But the constructor is created once per service and is then reuse for multiple invocations of `Run`

adecaro: but this check is cheaper than sampling randomness, no?

pkg/utils/retry.go
adecaro: yeah, but in which range do I sample?

pkg/utils/retry.go
adecaro: I changed strategy, now it should be better :)

PR #644, opened 2024-08-06
add functions to initiate a view context from an existing go context

arner:
for people who don't InitiateView directly but create the view.Context first, this PR creates the opportunity to pass a go context. This is useful for tracing a request through the process.

### DIFF
--- a/platform/view/core/manager/manager.go
+++ b/platform/view/core/manager/manager.go
@@ -242,38 +242,18 @@ func (cm *manager) InitiateViewWithIdentity(view view.View, id view.Identity, c 
 }
 
 func (cm *manager) InitiateContext(view view.View) (view.Context, error) {
-	cm.contextsSync.Lock()
-	ctx := cm.ctx
-	cm.contextsSync.Unlock()
-
-	return cm.InitiateContextFrom(ctx, view, cm.me(), "")
+	return cm.InitiateContextFrom(cm.ctx, view, cm.me(), "")
 }
 
 func (cm *manager) InitiateContextWithIdentity(view view.View, id view.Identity) (view.Context, error) {
-	cm.contextsSync.Lock()
-	ctx := cm.ctx
-	cm.contextsSync.Unlock()
-
-	return cm.InitiateContextFrom(ctx, view, id, "")
+	return cm.InitiateContextFrom(cm.ctx, view, id, "")
 }
 
 func (cm *manager) InitiateContextWithIdentityAndID(view view.View, id view.Identity, contextID string) (view.Context, error) {
-	cm.contextsSync.Lock()
-	ctx := cm.ctx
-	cm.contextsSync.Unlock()
-
-	return cm.InitiateContextFrom(ctx, view, id, contextID)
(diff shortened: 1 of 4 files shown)


platform/view/core/manager/manager.go
```diff
@@ -242,18 +242,35 @@ func (cm *manager) InitiateViewWithIdentity(view view.View, id view.Identity, c
...
 	cm.contextsSync.Lock()
 	ctx := cm.ctx
 	cm.contextsSync.Unlock()
+
+	return cm.InitiateContextFrom(ctx, view, id, contextID)
+}
+
+func (cm *manager) InitiateContextFrom(ctx context.Context, view view.View, id view.Identity, contextID string) (view.Context, error) {
+	if ctx == nil {
+		cm.contextsSync.Lock()
```
adecaro, line 270: I think we don't need the lock here to access cm.ctx.
The field is set when Start is called and Start is called only once at very beginning before any view can be initiated. Wdyt?

arner: updated the PR

platform/view/core/manager/manager.go
```diff
@@ -242,21 +242,18 @@ func (cm *manager) InitiateViewWithIdentity(view view.View, id view.Identity, c
 }
 
 func (cm *manager) InitiateContext(view view.View) (view.Context, error) {
-	return cm.InitiateContextWithIdentity(view, cm.me())
+	return cm.InitiateContextFrom(cm.ctx, view, cm.me(), "")
```
alexandrosfilios, line 245: We don't need to modify these two methods, right? InitiateContext and InitiateContextWithIdentity

platform/view/core/manager/manager.go
```diff
@@ -242,21 +242,18 @@ func (cm *manager) InitiateViewWithIdentity(view view.View, id view.Identity, c
...
 }
 
 func (cm *manager) InitiateContextWithIdentity(view view.View, id view.Identity) (view.Context, error) {
-	return cm.InitiateContextWithIdentityAndID(view, id, "")
+	return cm.InitiateContextFrom(cm.ctx, view, id, "")
 }
 
 func (cm *manager) InitiateContextWithIdentityAndID(view view.View, id view.Identity, contextID string) (view.Context, error) {
-	// Create the context
```
alexandrosfilios, line 253: I am iffy about this. Let's first make sure that it works on the CI and on the experiment before merging.

platform/view/core/manager/manager.go
arner: I wanted to keep their signatures as they were to not change the API. That's why I pass the default values to the new InitiateContextFrom function (which has an additional parameter for the context).

platform/view/core/manager/manager.go
arner: For what it's worth, I ran this in a test application and it works. CI also passes. Do you want to test it anywhere else?

arner:
test with token sdk succeeds: https://github.com/hyperledger-labs/fabric-token-sdk/pull/762

alexandrosfilios (approved):
LGTM

PR #647, opened 2024-08-13
Lock func

alexandrosfilios:
When CREATE OR REPLACE FUNCTION is called from two different threads (replicas), we get the following error

```
pq: tuple concurrently updated
```

Now we use a lock to coordinate the creation of the function and the trigger.


adecaro (approved):
LGTM, but please add a description to the PR

PR #659, opened 2024-09-02
Batcher runner implementation

alexandrosfilios:


adecaro:
@alexandrosfilios , shall we drop this?

PR #663, opened 2024-09-03
override configuration keys with environment variables

arner:
This PR enables the user to override any configuration key with an environment variable. Technically we were supposed to already have this feature (via viper's AutomaticEnv), but it was flaky and unpredictable:

- Didn't work structs that we got via UnmarshalKey 
- Removed higher level sibling keys when you overrode a low level value!

With this change you can for instance share a 'default' core.yaml, but override node specific data like the database connectionString or the node id. To prevent confusion the environment variables for logging and the vault config have been removed so that they can only be accessed via the same path as other configurations.

The commit also upgrades viper to the latest stable version. That means that if we merge this and upgrade Token SDK, we have to make sure it deals with lowercase map keys properly (like we already do with for instance the P2P field in FSC).


adecaro (approved):
LGTM, Arne could you document the variables that can be set from the environment?

platform/view/core/config/provider.go
```diff
@@ -116,25 +116,88 @@ func (p *provider) load() error {
...
+// that for UnmarshalKey values (see https://github.com/spf13/viper/pull/1699).
+// Example: CORE_LOGGING_FORMAT sets logging.format.
+func (p *provider) substituteEnv() {
+	for _, e := range os.Environ() {
+		if !strings.HasPrefix(e, strings.ToUpper(CmdRoot)+"_") {
+			continue
+		}
+
+		env := strings.Split(e, "=")
+		val := env[1]
```
alexandrosfilios, line 140: we don't need to set `val := env[1]`. We can do
```
if len(val) == 0 {continue}
val := strings.Join(env[1:], "=")
```

platform/view/core/config/provider.go
```diff
@@ -116,25 +116,88 @@ func (p *provider) load() error {
...
+		}
+		if len(val) == 0 {
+			continue
+		}
+		noprefix := strings.TrimLeft(env[0], strings.ToUpper(CmdRoot)+"_")
+		key := strings.ToLower(strings.ReplaceAll(noprefix, "_", "."))
+
+		// nested key
+		keys := strings.Split(key, ".")
+		if len(keys) == 1 {
```
alexandrosfilios, line 152: I think this condition is redundant. line 158 will yield `parent = ""`

platform/view/core/config/provider.go
```diff
@@ -116,25 +116,88 @@ func (p *provider) load() error {
...
+		}
+
+		k := p.v.GetStringMap(key)
+		if len(k) > 0 {
+			fmt.Println("-- skipping " + env[0] + ": cannot override maps")
+			continue
+		}
+
+		root := p.v.GetStringMap(keys[0])
+		setDeepValue(root, keys, val)
```
alexandrosfilios, line 172: Maybe it makes more sense to pass `keys[1:]`, because `keys[0]` is not in the map. Just for consistency in your method inputs.

platform/view/core/config/provider.go
```diff
@@ -116,25 +116,88 @@ func (p *provider) load() error {
...
+// Function to set the value at the deepest level
+func setDeepValue(m map[string]any, keys []string, value any) error {
+	// key = root but we don't have the map by reference
+	if len(keys) < 2 {
+		return errors.New("can't set root key")
+	}
+
+	current := m
+	// traverse to the last map
+	for i := 1; i < len(keys)-1; i++ {
```
alexandrosfilios, line 187: It's a matter of taste, but we could also do it with recursion
```
if len(keys) <= 1 {
 m[keys[0]] = value
 return nil
}
return setDeepValue(m[keys[0]].(map[string]any), keys[1:], value)
```

platform/view/core/config/provider.go
arner: agree that it would be more logical in some sense, but somehow it wasn't easy to change so I left it.

alexandrosfilios (approved):
LGTM

PR #667, opened 2024-09-06
CI Fixes

alexandrosfilios:


platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
```diff
@@ -34,8 +34,11 @@ const (
 
 type SubConnId = string
 
+var subConnId uint64
```
adecaro, line 37: why not using directly atomic.UInt64?

adecaro (approved):
LGTM, modulo the comment

PR #671, opened 2024-09-17
Added SetStates, DeleteStates, SetStateMetadatas and used typed param…

alexandrosfilios:
…eters for vault

### DIFF
--- a/platform/common/driver/vault.go
+++ b/platform/common/driver/vault.go
@@ -12,9 +12,16 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 )
 
+type (
+	PKey     = string
+	MKey     = string
+	RawValue = []byte
+	Metadata = map[MKey][]byte
+)
+
 type VersionedRead struct {
-	Key   string
-	Raw   []byte
+	Key   PKey
+	Raw   RawValue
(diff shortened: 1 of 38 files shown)


adecaro (approved):
LGTM

platform/common/driver/vault.go
```diff
@@ -12,19 +12,26 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 )
 
+type (
+	PKey     = string
```
mbrandenburger, line 16: Why is it called `PKey`? What is the `P` standing for?

alexandrosfilios: I guess property, but I chose this name because it was the most common name for this type of variable. for the metadata keys i used `MKey`

mbrandenburger: I see. I find it a bit misleading when looking at the code what uses now `driver.PKey` as it is not really clear what this type is meant for. Is it something vault specific or is is a Public Key? :)

alexandrosfilios: if rwset is a map, pkey is the key and rawvalue is the value. we could also name it `RWSKey` and `RWSValue`. I would do it in another separate pr to be honest once we agree on something

PR #672, opened 2024-09-18
Fix assertion error message

mbrandenburger:
This PR fixes the error message returned by some assertions in the IOU sample


adecaro (approved):
LGTM

PR #674, opened 2024-09-18
Fix error shadowing in fabric state service

mbrandenburger:
The error handling of rwSet.GetReadAt overrides the original error that results in returning a wrapped nil error.


adecaro (approved):
LGTM

PR #680, opened 2024-09-20
Optimize txidstore updates by batching them

alexandrosfilios:


adecaro (approved):
LGTM. I would suggest to remove the protos and the unnecessary unamrshalling.

PR #681, opened 2024-09-20
remove Track and IsTxFinal

adecaro:
This PR does remove the following view commands:
- `Track`. It was not implemented and never requested.
- `IsTxFinal`. Each application should provide this functionality only if needed.


integration/fabric/iou/views/finality.go
```diff
@@ -0,0 +1,41 @@
+/*
```
alexandrosfilios, line 1: Maybe we could reuse the view that we defined before?

alexandrosfilios (approved):
LGTM modulo comment

integration/fabric/iou/views/finality.go
adecaro: done

alexandrosfilios (approved):
LGTM

PR #683, opened 2024-09-23
Re-use TxIDs in listener manager instead of maintaining a different s…

alexandrosfilios:
…lice


arner (approved):
Tested with Token SDK and on my own sample application, LGTM!

PR #703, opened 2024-12-04
Created ScanBlock method to allow for custom and more efficient uses …

alexandrosfilios:
…of the delivery service


adecaro (approved):
LGTM.
The other optmization that I see here is that while processing the current block, the service fetches already the next one, if available.

PR #707, opened 2024-12-09
shared sig service

adecaro:


platform/common/services/sig/service.go
```diff
@@ -60,36 +61,46 @@ func (o *Service) RegisterSigner(identity view.Identity, signer driver.Signer, v
...
 
+	// check existence with a read lock
 	idHash := identity.UniqueID()
-	o.viewsSync.Lock()
+	o.mutex.RLock()
 	s, ok := o.signers[idHash]
-	o.viewsSync.Unlock()
+	o.mutex.RUnlock()
+	if ok {
+		logger.Infof("another signer bound to [%s]:[%s][%s] from [%s]", identity, GetIdentifier(s), GetIdentifier(signer), string(s.DebugStack))
```
arner, line 70: when does this happen? I saw a lot of those in the logs (used to be warn even) and always wondered if it was part of the normal flow.

adecaro: I did some exploration in past, I can check again

adecaro: we can define put it as debug though

platform/common/services/sig/service.go
```diff
@@ -102,21 +113,34 @@ func (o *Service) RegisterVerifier(identity view.Identity, verifier driver.Verif
...
 
+	// check existence with a read lock
 	idHash := identity.UniqueID()
-	o.viewsSync.Lock()
+	o.mutex.RLock()
 	v, ok := o.verifiers[idHash]
-	o.viewsSync.Unlock()
+	o.mutex.RUnlock()
+	if ok {
+		logger.Warnf("another verifier bound to [%s]:[%s][%s] from [%s]", idHash, GetIdentifier(v), GetIdentifier(verifier), string(v.DebugStack))
```
arner, line 122: Infof or Debugf? Or should this be a warning?

arner (approved):

platform/fabric/core/generic/driver/provider.go
```diff
@@ -31,86 +29,38 @@ type Provider struct {
 	metricsProvider    metrics.Provider
 	channelProvider    generic.ChannelProvider
 	sigService         *sig.Service
-	mspManagerProvider MSPManagerProvider
+	mspManagerProvider identity.MSPManagerProvider
 }
 
 func NewProvider(
```
alexandrosfilios, line 35: We could have another additional constructor that receives the sigService as a dependency (like we did before), so that other SDKs that want to instantiate a new provider can re-use this one by injecting their own sigservice.

alexandrosfilios (commented):
Comments

PR #708, opened 2024-12-11
enable race unit-tests

adecaro:
fix holder race condition

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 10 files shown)


.github/workflows/tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tests: [unit-tests]
+        tests: [unit-tests, unit-tests-race]
```
arner, line 36: do we need both? The logic is still tested if the `-race` flag is enabled so maybe just the second would be enough?

adecaro: good point, will remove

PR #714, opened 2024-12-13
pass the network driver to the constructor

adecaro:
This PR simplify the constructors used to create a new channel. It also replaces certain dependencies with interfaces and not structs.

This PR is tested against FTS in https://github.com/hyperledger-labs/fabric-token-sdk/pull/804

### DIFF
--- a/platform/fabric/core/generic/channelprovider.go
+++ b/platform/fabric/core/generic/channelprovider.go
@@ -28,41 +28,41 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-type VaultConstructor = func(configService driver.ConfigService, channel string, drivers []driver2.NamedDriver, metricsProvider metrics.Provider, tracerProvider trace.TracerProvider) (*vault.Vault, driver.TXIDStore, error)
+type VaultConstructor = func(
+	configService driver.ConfigService,
+	channel string,
+	drivers []driver2.NamedDriver,
+	metricsProvider metrics.Provider,
+	tracerProvider trace.TracerProvider,
+) (*vault.Vault, driver.TXIDStore, error)
 type LedgerConstructor func(
 	channelName string,
+	nw driver.FabricNetworkService,
 	chaincodeManager driver.ChaincodeManager,
-	localMembership driver.LocalMembership,
-	configService driver.ConfigService,
-	transactionManager driver.TransactionManager,
-) driver.Ledger
+) (driver.Ledger, error)
 type RWSetLoaderConstructor func(
-	network string,
 	channel string,
+	nw driver.FabricNetworkService,
 	envelopeService driver.EnvelopeService,
 	transactionService driver.EndorserTransactionService,
-	transactionManager driver.TransactionManager,
 	vault driver.RWSetInspector,
-) driver.RWSetLoader
+) (driver.RWSetLoader, error)
 type CommitterConstructor func(
-	configService driver.ConfigService,
+	nw driver.FabricNetworkService,
 	channelConfig driver.ChannelConfig,
 	vault driver.Vault,
 	envelopeService driver.EnvelopeService,
 	ledger driver.Ledger,
 	rwsetLoaderService driver.RWSetLoader,
-	processorManager driver.ProcessorManager,
 	eventsPublisher events.Publisher,
 	channelMembershipService *membership.Service,
-	orderingService committer.OrderingService,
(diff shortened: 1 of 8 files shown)


ci/fsc-fts-update.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
mbrandenburger, line 1: I believe this script belongs to the fts repo

platform/fabric/sdk/dig/generic/providers.go
```diff
@@ -27,6 +28,7 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/kvs"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics"
 	"github.com/hyperledger/fabric-protos-go/common"
+	"github.com/pkg/errors"
```
mbrandenburger, line 31:
```suggestion
	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
```

mbrandenburger (changes requested):
Overall looks good to me. Just two comments.

ci/fsc-fts-update.sh
adecaro: I really meant to have a script that updates FTS with the current FSC commit.

mbrandenburger: I understand. However, I think this script should move to the FTS repo as it creates a PR on the FTS repo.

Such a script could support multiple strategies:
- Get the commit hash from any local git clone / working branch.
- Get the lastest commit hash from main
- Provide the commit hash as a parameter to the script and update all go.mod files inside the project.

I believe the right way would be to use dependabot and schedule a daily check if the main branch on FSC has something new. As dependabot only "detects" versioned updated (not individual commits) it won't work like that. You can create a GH action that is fired once per day to check if there is a new commit on FSC main branch and then create a PR.

adecaro: okay, I'll remove the script.

mbrandenburger (approved):

platform/fabric/core/generic/channelprovider.go
```diff
@@ -176,36 +186,35 @@ func (p *provider) NewChannel(nw driver.FabricNetworkService, channelName string
...
-		nw.TransactionManager(),
 	)
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed creating ledger for channel [%s]", channelName)
+	}
 
-	committerService := p.newCommitter(
-		nw.ConfigService(),
+	committerService, err := p.newCommitter(
+		nw,
```
alexandrosfilios, line 199: Why do we pass the whole `nw`? We just want a few services, right?

platform/fabric/core/generic/ordering/ordering.go
```diff
@@ -133,6 +137,16 @@ func (o *Service) SetConsensusType(consensusType ConsensusType) error {
 	return nil
 }
 
+func (f *Service) SetConfigOrderers(o channelconfig.Orderer, orderers []*grpc.ConnectionConfig) error {
```
alexandrosfilios, line 140: We pass `o` that already contains `Consenters` and `orderers`. Do we need both? We could only pass ordererType.

platform/fabric/core/generic/channelprovider.go
```diff
@@ -28,41 +28,41 @@ import (
...
+type VaultConstructor = func(
+	configService driver.ConfigService,
+	channel string,
+	drivers []driver2.NamedDriver,
+	metricsProvider metrics.Provider,
+	tracerProvider trace.TracerProvider,
+) (*vault.Vault, driver.TXIDStore, error)
 type LedgerConstructor func(
 	channelName string,
+	nw driver.FabricNetworkService,
```
alexandrosfilios, line 40: I don't know if it's a good idea to replace the actual dependencies by a struct that contains these and many more. This creates a tighter coupling and looser cohesion.

alexandrosfilios (changes requested):

platform/fabric/core/generic/ordering/ordering.go
adecaro: good point, we can simplify this.

alexandrosfilios (commented):
LGTM

alexandrosfilios (approved):
LGTM

PR #715, opened 2024-12-16
when reading channel config, skip orderer endpoints which are empty

arner:
And set a default for the orderer retryInterval. Before, it defaulted to immediate.


adecaro (approved):
LGTM

PR #727, opened 2025-01-15
Decouple KVS implementation

alexandrosfilios:
Moved KVS functionality behind interfaces and grouped them based on domains for decoupling


adecaro (approved):
LGTM

PR #728, opened 2025-01-16
Performance enhancements

alexandrosfilios:
Performance enhancements: Flatten binding hierarchy. Now all bindings point to the long-term id


adecaro (approved):
LGTM

PR #734, opened 2025-01-20
Batch tx fetches from ledger

alexandrosfilios:
Reduced number of calls to delivery service when a listener is evicted


adecaro (approved):
LGTM

PR #740, opened 2025-01-28
fix delivery service

adecaro:
This PR does the following: It decouples the checking of the stop condition from the reading of the blocks from the delivery service that is a blocking operation. This avoids that the reading process gets stuck after the callback notifies to stop.


alexandrosfilios (approved):
LGTM

PR #743, opened 2025-01-30
return error properly

adecaro:
This PR does the following:
1. It reduces logging form Infof to Debugf;
2. It fixes a bug that was preventing an error returned by the callback function to be propagated as expected.


alexandrosfilios (approved):
LGTM

PR #750, opened 2025-02-04
Extracted querying by id to a different service

alexandrosfilios:


platform/fabric/core/generic/finality/listenermanager.go
```diff
@@ -43,6 +45,10 @@ type ListenerManager[T TxInfo] interface {
 	RemoveFinalityListener(txID string, e ListenerEntry[T]) error
 }
 
+type QueryByIDService[T TxInfo] interface {
+	QueryByID(ctx context.Context, lastBlock driver2.BlockNum, evicted map[driver2.TxID][]ListenerEntry[T]) (<-chan []T, error)
```
alexandrosfilios, line 49: it's better to call it starting block

platform/fabric/core/generic/delivery/service.go
```diff
@@ -206,9 +251,14 @@ func (p *processedTransaction) ValidationCode() int32 {
 }
 
 type fakeVault struct {
-	txID string
+	txID  string
```
alexandrosfilios, line 254: we can use types TxID and BlockNum. also for the method GetLastBlock

platform/fabric/core/generic/delivery/service.go
```diff
@@ -206,9 +251,14 @@ func (p *processedTransaction) ValidationCode() int32 {
...
+	block uint64
 }
 
 func (f *fakeVault) GetLast(context.Context) (*driver2.TxStatus, error) {
```
alexandrosfilios, line 258: We can call it `GetLastTxID() driver2.TxID`

PR #754, opened 2025-02-06
Add isolation option to the vault lock

alexandrosfilios:


platform/common/driver/vault.go
```diff
@@ -8,6 +8,7 @@ package driver
 
 import (
 	"context"
+	"database/sql"
```
adecaro, line 11: the vault should be sql independent

platform/common/driver/vault.go
```diff
@@ -112,16 +115,18 @@ type Writes map[Namespace]map[PKey]VaultValue
...
 
 type VaultStore interface {
 	// AcquireTxIDRLock acquires a read lock on a specific transaction.
 	// While holding this lock, other routines:
-	// - cannot update the transaction states or statuses of the locked transactions
-	// - can read the locked transactions
-	AcquireTxIDRLock(ctx context.Context, txID TxID) (VaultLock, error)
+	// - cannot update the transaction states of the locked transactions (based on the isolation level passed)
+	// - can read the locked states
+	AcquireTxIDRLock(ctx context.Context, txID TxID, isolationLevel sql.IsolationLevel) (VaultLock, error)
```
adecaro, line 129: It is not clear to me what's the purpose of `AcquireTxIDRLock`. Maybe renaming it to something like `NewTransaction`? How should the developer use this API?

platform/orion/vault.go
```diff
@@ -74,7 +74,7 @@ func (v *vault) NewRWSet(ctx context.Context, txid driver2.TxID) (*RWSet, error)
 }
 
 func (v *vault) GetRWSet(ctx context.Context, id driver2.TxID, results []byte) (*RWSet, error) {
```
adecaro, line 76: Rename also this function to `NewRWSetFromBytes`

platform/fabric/vault.go
```diff
@@ -146,7 +146,7 @@ func (c *Vault) NewRWSet(ctx context.Context, txID driver.TxID) (*RWSet, error)
 // A client may obtain more than one such simulator; they are made unique
 // by way of the supplied txid
 func (c *Vault) GetRWSet(ctx context.Context, txid driver.TxID, rwset []byte) (*RWSet, error) {
```
adecaro, line 148: Rename to `NewRWSetFromBytes`

adecaro:
@alexandrosfilios, can you add a performance test to measure how many transaction per second you can commit while producing RWSets?

alexandrosfilios:
> @alexandrosfilios, can you add a performance test to measure how many transaction per second you can commit while producing RWSets?

Yes writing perf tests as part of another pr

platform/common/driver/vault.go
```diff
@@ -112,24 +114,13 @@ type Writes map[Namespace]map[PKey]VaultValue
 
 // VaultLock represents a lock over a transaction or the whole vault
 type VaultLock interface {
```
adecaro, line 116: Is this interface still needed?

platform/common/core/generic/vault/interceptor.go
```diff
@@ -82,15 +82,6 @@ func NewInterceptor[V driver.ValidationCode](
 }
 
 func (i *Interceptor[V]) IsValid() error {
-	tx, err := i.vaultStore.GetTxStatus(context.Background(), i.txID)
```
adecaro, line 85: why is this check not needed anymore?

platform/view/services/db/driver/sql/sqlite/vault.go
```diff
@@ -51,18 +51,13 @@ func newTxCodePersistence(readDB *sql.DB, writeDB common.WriteDB, tables common.
 
 func (db *VaultPersistence) Store(ctx context.Context, txIDs []driver.TxID, writes driver.Writes, metaWrites driver.MetaWrites) error {
 	span := trace.SpanFromContext(ctx)
-	span.AddEvent("start_store")
-	defer span.AddEvent("end_store")
+	span.AddEvent("Start store")
```
adecaro, line 54: the rest of the code still uses the convention of separating words with `_` and keeping everything lowercase. Was this intentional?

adecaro (approved):
I left some questions to be checked just to make sure the semantic is still correct. I think an interface is not needed anymore.

platform/common/core/generic/vault/interceptor.go
alexandrosfilios: because the db makes the check

alexandrosfilios: And this method is only needed in tests

PR #759, opened 2025-02-17
fabric: scan block from

adecaro:
introduce scan block from to start the scanning of block from a given block num


alexandrosfilios (approved):
LGTM

Issue #767, opened 2025-03-12
nwo fsc topology: AddSDKWithBase


adecaro:
This Github Issue asks to add a new function `AddSDKWithBase` to compose two SDKs like this

```go
n.InstallSDK(business.NewFrom(base.NewSDK(n)))
```

PR #768, opened 2025-03-12
nwo fsc topology: AddSDKWithBase + *view.Registry

adecaro:
This PR fulfills the requests of #767 and register *view.Registry in DIG. Both features are used by the Token-SDK in this PR: https://github.com/hyperledger-labs/fabric-token-sdk/pull/949


integration/nwo/fsc/node/note_test.go
```diff
@@ -0,0 +1,36 @@
+/*
```
alexandrosfilios, line 1: node_test.go

integration/nwo/fsc/node/note_test.go
```diff
@@ -0,0 +1,36 @@
...
+package node
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type DummySDK struct {
```
alexandrosfilios, line 16: for documentation purposes, can you add another SDK here
```
type DummySDK2 struct {}
type DummySDK3 struct {}
```

and then test
```
node.AddSDKWithBase(&DummySDK{}, &DummySDK2{})
node.AddSDKWithBase(&DummySDK3{})
```
This way we know what to expect when we read it and we explain all possible ways to use it.

platform/view/sdk/dig/sdk.go
```diff
@@ -72,6 +72,7 @@ func NewSDKWithContainer(c *dig.Container, registry node.Registry) *SDK {
 		sdk.C.Provide(digutils.Identity[node.Registry](), dig.As(new(driver.ServiceProvider), new(node.Registry), new(view.ServiceProvider), new(finality.Registry))),
 		sdk.C.Provide(func() *view.ConfigService { return sdk.Config }),
 		sdk.C.Provide(digutils.Identity[*view.ConfigService](), dig.As(new(driver.ConfigService), new(id.ConfigProvider), new(endpoint.ConfigService))),
+		sdk.C.Provide(view.NewRegistry),
```
alexandrosfilios, line 75: In this case maybe it makes sense to remove the `node.Registry` from line 71 and only provide the `view.NewRegistry`?

alexandrosfilios (approved):
LGTM modulo comments

platform/view/sdk/dig/sdk.go
adecaro: agree, but let's reorganize the dependencies in a different PR because it might be disruptive

integration/nwo/fsc/node/note_test.go
adecaro: okay

integration/nwo/fsc/node/note_test.go
adecaro: Extended the test. For the second part, the code there doesn't use reflection. It is just combining strings.

PR #776, opened 2025-03-17
Split view factory from view manager

alexandrosfilios:
Created enhanced container to register factories


adecaro (approved):
LGTM

adecaro (changes requested):
I think this PR is too big

PR #781, opened 2025-03-19
Created enhanced container to register factories

alexandrosfilios:
Split view factory from view manager
Created enhanced container to register factories


adecaro (approved):
LGTM

PR #783, opened 2025-03-20
remove samples

adecaro:
These sample were not maintained and tested. Better to remove from here. The smart client is already exercised by the token-sdk sample [`here`](https://github.com/hyperledger/fabric-samples/blob/main/token-sdk/README.md)


alexandrosfilios (approved):
LGTM

PR #852, opened 2025-03-24
Removed link, Added repo name to postgres image and print URLs to Gra…

alexandrosfilios:
…fana and Prometheus UI


adecaro (approved):
LGTM

PR #863, opened 2025-03-25
Fix flaky test

mbrandenburger:
Before:
```bash
❯ time go test -count=100 -failfast -race ./platform/fabric
--- FAIL: TestEventListener (1.38s)
    events_test.go:86: expected channel to be closed
FAIL
FAIL    github.com/hyperledger-labs/fabric-smart-client/platform/fabric 27.400s
FAIL
go test -count=100 -failfast -race ./platform/fabric  2.08s user 1.79s system 13% cpu 29.501 total
```

Now:
```bash
❯ time go test -count=100 -failfast -race ./platform/fabric
ok      github.com/hyperledger-labs/fabric-smart-client/platform/fabric 10.572s
go test -count=100 -failfast -race ./platform/fabric  10.89s user 3.79s system 124% cpu 11.822 total
```

### DIFF
--- a/platform/fabric/events_test.go
+++ b/platform/fabric/events_test.go
@@ -8,7 +8,6 @@ package fabric
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"testing"
 	"time"
@@ -16,6 +15,13 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/committer"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/events"
 	"github.com/stretchr/testify/assert"
+)
+
+const (
+	publicationSnooze = time.Millisecond
+	timeout           = 75 * time.Millisecond
+	waitFor           = 4 * timeout
+	tick              = timeout / 10
 )
 
 type mockSubscriber struct {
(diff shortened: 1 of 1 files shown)


platform/fabric/events_test.go
```diff
@@ -18,74 +17,88 @@ import (
...
-		for i := 0; i < 1000; i++ {
-			subscriber.Publish("testChaincode", &committer.ChaincodeEvent{Payload: []byte(fmt.Sprintf("event %d", i))})
-			time.Sleep(1 * time.Millisecond)
+		msg := []byte("some msg")
+		for {
+			select {
+			case <-done:
+				return
+			default:
+				subscriber.Publish("testChaincode", &committer.ChaincodeEvent{Payload: msg})
```
ale-linux, line 45: Without a sleep or a gosched, this will consume 100% of the CPU (unless `Publish` has a lot of blocking calls). I'd add either one of those.

platform/fabric/events_test.go
```diff
@@ -18,74 +17,88 @@ import (
...
 
-	// Ensure the channel is closed
-	select {
-	case _, ok := <-eventChannel:
-		if ok {
-			t.Fatal("expected channel to be closed")
-		}
-	default:
-		t.Fatal("expected to read from closed channel")
+type mockSubscriber struct {
```
ale-linux, line 81: (super minor nit, more like a philosophical comment): it's always useful to minimise diffs if possible so that reviewers can see what really changed. The diff chunk below exists but nothing actually changed. It would have been more efficient to actually leave the code where it was.

ale-linux (commented):
LGTM - modulo a couple of nits

platform/fabric/events_test.go
mbrandenburger: good catch - I've added a sleep

platform/fabric/events_test.go
mbrandenburger: yeah - I agree

mbrandenburger:
@ale-linux thanks for the review. I've updated the PR accordingly.

ale-linux (approved):
Thanks @mbrandenburger !

PR #866, opened 2025-03-25
Bump viper dep

mbrandenburger:
@arner my expectations to remove 20 indirect deps with this PR .... not fulfilled 🤓


arner:
That's disappointing! The go.mod of viper itself went from 90 to 24 lines of code. Apparently we didn't use any of the transitive dependencies... v1.21 just got released, maybe good to bump it to the latest?

arner (approved):

PR #867, opened 2025-03-25
Remove go-cmp dep

mbrandenburger:
remove unused code in rwset impl


mbrandenburger:
not needed as #784 got merged!

PR #877, opened 2025-04-02
Pagination keyset2

HayimShaul:
added keyset pagination

### DIFF
--- a/platform/view/services/db/driver/sql/common/pagination.go
+++ b/platform/view/services/db/driver/sql/common/pagination.go
@@ -87,6 +87,67 @@ func (p *OffsetPagination) GoBack(numOfpages int) (driver.Pagination, error) {
 func (p *OffsetPagination) Prev() (driver.Pagination, error) { return p.GoBack(1) }
 func (p *OffsetPagination) Next() (driver.Pagination, error) { return p.GoForward(1) }
 
+type KeysetPagination struct {
+	offset      int
+	pageSize    int
+	pageInRange bool
+	// name of the field in the database that is a unique id of the records
+	idFieldName string
+	// the last id value read and the offset in which it was read
+	lastId     string // TODO: should this be int?
+	lastOffset int
+}
+
+func NewKeysetPagination(offset int, pageSize int, idFieldName string) (*KeysetPagination, error) {
+	if offset < 0 {
+		return nil, fmt.Errorf("offset shoud be grater than zero. Offset: %d", offset)
+	}
+	if pageSize < 0 {
+		return nil, fmt.Errorf("page size shoud be grater than zero. pageSize: %d", pageSize)
+	}
+	return &KeysetPagination{
+		offset:      offset,
+		pageSize:    pageSize,
+		pageInRange: true,
+		idFieldName: idFieldName,
+		lastId:      "",
+		lastOffset:  -1,
+	}, nil
+}
+
+func (p *KeysetPagination) GoToOffset(offset int) (driver.Pagination, error) {
+	if offset < 0 {
+		return NewEmptyPagination(), nil
+	}
+	return &KeysetPagination{
+		offset:      offset,
+		pageSize:    p.pageSize,
+		pageInRange: true,
+		idFieldName: p.idFieldName,
+		lastId:      p.lastId,
+		lastOffset:  p.lastOffset,
+	}, nil
+}
+
+func (p *KeysetPagination) GoToPage(pageNum int) (driver.Pagination, error) {
+	return p.GoToOffset(pageNum * p.pageSize)
+}
+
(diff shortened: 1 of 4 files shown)


platform/common/driver/vault.go
```diff
@@ -62,6 +64,84 @@ type TxStateIterator = collections.Iterator[*VaultRead]
...
+	table  string
+	fields []string
+	where  []string
+	limit  string
+	offset string
+	order  string
+	params []any
+}
+
+func (q *SqlQuery) SetTable(t string) {
```
alexandrosfilios, line 77: I would give them names like `From()` so that they look more like actual queries. The goal would be:

```
type FieldName struct {
 table Table
 name FieldName
}
type Table string
func (t Table) Join(o Table, ons ...Condition) Table
func (t Table) Field(f string) FieldName


myTable := Table("my_table")
otherTable := Table("other_table")
Select("id", "name")
 .From(myTable.Join(otherTable, ci.CmpFields(myTable.Field("id"), otherTable.Field("id"), "=")))
 .Where(...)
 .OrderBy()
 .Limit()
```

This may look complex but joins are not that frequent and we can have utility methods for these cases to make them shorter, such as
```
Select()
.From(myTable.JoinOn(otherTable, "id", "id"))
.Where()
...
```
Also, the way it is now, it is not very safe. We should leverage interfaces to follow a "poke-yoke" principle, i.e. not allow the user to create whatever query. an example:

```
type selectQuery interface {
 From(...) fromQuery
}

type fromQuery interface {
 whereQuery
 Where(...) whereQuery
}
type whereQuery interface {
 orderByQuery
 OrderBy(...) orderByQuery
}
type orderByQuery interface {
 Limit(...) limitQuery
}
type limitQuery interface {}
```

platform/view/services/db/driver/sql/common/pagination.go
```diff
@@ -85,25 +87,140 @@ func (p *OffsetPagination) GoBack(numOfpages int) (driver.Pagination, error) {
 func (p *OffsetPagination) Prev() (driver.Pagination, error) { return p.GoBack(1) }
 func (p *OffsetPagination) Next() (driver.Pagination, error) { return p.GoForward(1) }
 
+type KeysetPagination struct {
+	offset   int
+	pageSize int
+	// name of the field in the database that is a unique id of the records
+	sqlIdName string
```
alexandrosfilios, line 94: the type of this is `FieldName`. (only of this property. `idFieldName` is not. keep it as string.

platform/view/services/db/driver/sql/common/pagination.go
```diff
@@ -85,25 +87,140 @@ func (p *OffsetPagination) GoBack(numOfpages int) (driver.Pagination, error) {
...
+	}
+	if pageSize < 0 {
+		return nil, fmt.Errorf("page size shoud be grater than zero. pageSize: %d", pageSize)
+	}
+	return &KeysetPagination{
+		offset:      offset,
+		pageSize:    pageSize,
+		sqlIdName:   sqlIdName,
+		idFieldName: idFieldName,
+		firstId:     "",
```
alexandrosfilios, line 114: you don't need to set empty strings. you can omit them and they default to the empty string

alexandrosfilios (changes requested):
left some comments on the sql builder before proceeding.

platform/common/driver/vault.go
HayimShaul: Didn't fully understand:


1.
 .From(myTable.Join(otherTable, ci.CmpFields(myTable.Field("id"), otherTable.Field("id"), "=")))

do you mean to add a WHERE clause? then why not Where(...)? Isn't From(..) used for the table name?

2. poka yoke - I don't understand how your example checks for errors

3. you don't want to write  Select("id").Where(...).From().From().Limit().Where()   what you wrote at the top suggested that you did

HayimShaul:
Alex says:

1. Why do we need empty pagination?

2. write tests for the query that has to be generated every time depending on the pagination and after going back and forth to the pages.
you don't have to test it on a specific table
because that couples your tests
of course you can also have a test in the vault where you set the pagination

3. I have written some tests, but beware the query for the keyset is not correct

4. why do we need an offset when we have keyset pagination?

5. anyways try to write some meaningful tests that cover different cases, eg.
the key is string in the current case. What if it is an int? what if it is something non comparable? I already have these fields in the struct
 - What happens when we call next and prev on offset or keyset?
 - What if they return empty result sets? You dont need a database, I simulate the result there.
 - What happens if we have select *? or if we havent selected the ID in the existing query? We should add it to the select fields. But if we had already added it, we shouldnt add it again. what if we already had WHERE, ORDER BY, LIMIT, OFFSET? (i took care of those cases btw)
Try to think from a user perspective when you create your interfaces. The vault shouldnt know how to update the pagination.
 - write some cases and fix some cases and then i can re-review

PR #886, opened 2025-04-10
Update go version to 1.24.0

Vibhuair20:


mbrandenburger (changes requested):
Thanks! This looks almost ready. I think you only missed to update the go version in `./go.mod`.

adecaro:
Setting the version in the `go.mod` file imposes the minimum go version the project must be run against. 
Any reason why go1.24 was selected? Many are still using go1.23.

Vibhuair20:
Hello Angelo,
The main reason behind updating the go mod version is that with the latest
version the concurrency and the testing efficiency.
Which in case of large projects can be very beneficial.
and also parallel testing is optimized.

PR #891, opened 2025-04-14
Refactor db config (#953)

alexandrosfilios:
The target is to simplify the db configuration and make it re-usable, while enable default behavior.
This PR addresses the following:
* flatten the config
```
persistence:
 type: memory

persistence:
 type: sqlite
 opts:
  maxIdleConns: 3
  skipPragmas: false
 ...

persistence:
 type: postgres
 opts:
  maxIdleConns: 3
```
* decouple configs for different persistence types and remove unnecessary opts (before we used one unified config for postgres and sql, so we defined opts that were not used, e.g. skipPragmas.
* move configuration logic for db's to one point, so we can deduplicate the logic and easily replace it (see next step below).

The next step will be to:
* introduce a new section with named configs:
```
fsc.persistences:
 default:
  type: postgres
  opts:
   maxIdleConns: 3
 my_custom_postgres:
  type: postgres
  opts:
   maxIdleConns: 4
  my_custom_mem:
   type: memory
```
A service would read and provide this config:
```
type PersistenceName string
const DefaultPersistence PersistenceName = "default"

type DbConfigService interface {
 GetConfig(PersistenceName) (Config, error)
}
```
* The config for each component is simplified to pointing to the name of the persistence (or default to "default" if not specified")
```
kvs:
 persistence: my_custom_postgres
 cache:
  size: 30
```
or
```
kvs:
  cache:
   size: 30
```


platform/view/services/db/driver/driver.go
```diff
@@ -103,23 +105,34 @@ type Config interface {
...
+func (d NamedDrivers) GetDriver(t driver.PersistenceType) (Driver, error) {
+	for _, dr := range d {
+		if dr.Name == t {
+			return dr.Driver, nil
+		}
+	}
+	return nil, errors.Errorf("driver %s not found", t)
+}
+
 type Driver interface {
```
adecaro, line 119: Please, remove the databases that do not belong to the view-sdk

adecaro (approved):
I approve, but either you update `core-fabric.md` or you update the description of the PR.

PR #898, opened 2025-04-22
Modified config

alexandrosfilios:
Introduced a new section in core.yaml to define all persistences:

```
fsc.persistences:
 default:
  type: postgres
  opts:
   maxIdleConns: 3
 my_custom_postgres:
  type: postgres
  opts:
   maxIdleConns: 4
  my_custom_mem:
   type: memory
```

Now instead of defining all persistence configurations inline for each store (e.g. kvs), we reference them by their name
```
kvs:
 persistence: my_custom_postgres
 cache:
  size: 30
```

or we don't define them, and the `default` is selected:
```
kvs:
  cache:
   size: 30
```

This PR also changes the signature of the drivers to reduce the dependencies. Instead of passing the config service, we only pass the persistence name (e.g. `my_custom_postgres`, `default` or the empty name `` that defaults to `default`)
```
type Driver interface {
	// NewKVS returns a new UnversionedPersistence for the passed data source and config
	NewKVS(PersistenceName, ...string) (UnversionedPersistence, error)
       ...
}
```


docs/core-fabric.md
```diff
@@ -457,9 +478,8 @@ Simple:
 
 ```yaml
 persistence:
-  type: sql
+  type: sqlite
```
adecaro, line 481: is this accurate? should `persistence` be something else?

docs/db-driver.md
```diff
@@ -11,8 +11,8 @@ Although these methods have different names and return different stores, their s
 
 ```go
 type Driver interface {
-    NewBindingStore(string, Config) (BindingStore, error)
-    NewAuditInfoStore(string, Config) (AuditInfoStore, error)
+    NewBindingStore(name PersistenceName, params ...string) (BindingStore, error)
```
adecaro, line 14: what's params?

docs/db-driver.md
adecaro, line 1: Don't you have to update also the section `Extending the Driver`?

platform/view/sdk/dig/providers.go
```diff
@@ -7,59 +7,33 @@ SPDX-License-Identifier: Apache-2.0
...
-	return signerinfo.NewStore(in.Config, in.Drivers, "default")
-}
-
-func newAuditInfoStore(in struct {
-	dig.In
-	Config  vdriver.ConfigService
-	Drivers []dbdriver.NamedDriver `group:"db-drivers"`
-}) (driver.AuditInfoStore, error) {
-	return auditinfo.NewStore(in.Config, in.Drivers, "default")
+	return kvs.New(utils.MustGet(kvs2.NewStore(config, driver)), "_default", size)
```
adecaro, line 36: please, avoid the user of `MustGet` especially given that you can return an error here.

platform/view/services/db/driver/driver.go
```diff
@@ -95,6 +93,14 @@ type UnversionedWriteTransaction interface {
 	Discard() error
 }
 
+type PersistenceName string
```
adecaro, line 96: please, add comments to these types

adecaro (approved):
LGTM, modulo the comments

docs/db-driver.md
alexandrosfilios: Params is what will be added as a prefix for this table.

For fsc, you usually use `default`, apart from the case of the vault where you also have one per channel.

For the case of fts, you have one table per tms, so these params would be network, channel, namespace. This makes it more applicable for any application that builds on top to keep the same signature format for the store instantiation.

docs/db-driver.md
alexandrosfilios: Good catch!

docs/core-fabric.md
alexandrosfilios: In general it should refer to the name of the persistence config. But here it explains more what the format of each config is, that's why I left it like that.

PR #906, opened 2025-05-12
Create a logger that automatically extracts the package and uses it a…

SaidAltury-ibm:
…s a prefix #815

### DIFF
--- a/platform/common/services/logging/logger.go
+++ b/platform/common/services/logging/logger.go
@@ -8,11 +8,14 @@ package logging
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"net/http"
+	"runtime"
 	"strings"
 	"testing"
 
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-lib-go/common/flogging/floggingtest"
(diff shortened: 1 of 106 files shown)


platform/common/services/logging/logger.go
```diff
@@ -54,8 +57,45 @@ func Named(loggerName string) Option {
 	}
 }
 
-func MustGetLogger(loggerName string) Logger {
-	return &logger{FabricLogger: flogging.MustGetLogger(loggerName)}
+func MustGetLogger(params ...string) Logger {
```
alexandrosfilios, line 60: To separate the concerns better, I would first find the name of the package

```
func getPackageName() (string, error) {
	pc, _, _, ok := runtime.Caller(4)
	if !ok {
		return "", errors.New("failed to get caller package name")
	}
	fn := runtime.FuncForPC(pc)
	if fn == nil {
		return "", errors.New("failed to get caller package name")
	}
	fullFuncName := fn.Name()
	return fullFuncName[:strings.LastIndex(fullFuncName, ".")], nil
}
```

and given this name I would create the logger name by splitting the package path and appending the params:
```
func loggerName(fullPkgName string, replacements map[string]string, params ...string) string {
	nameParts := append(strings.Split(fullPkgName, "/"), params...)
	name := strings.Join(nameParts, ".")

	for old, newVal := range replacements {
		name = strings.ReplaceAll(name, old, newVal)
	}
	return name
}
```

Then the whole method would be:
```
func GetLoggerWithReplacements(replacements map[string]string, params []string) (Logger, error) {
	fullPkgName, err := getPackageName()
	if err != nil {
		return nil, err
	}
	return &logger{FabricLogger: flogging.MustGetLogger(loggerName(fullPkgName, replacements, params...))}, nil
}
```

alexandrosfilios (commented):
LGTM but check the comment

alexandrosfilios (approved):
LGTM

alexandrosfilios (approved):
LGTM

Issue #884, opened 2025-04-08
Potential deadlock with NetworkStreamSession with Websockets


mbrandenburger:
The current implementation of the [NetworkStreamSession](https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/view/services/comm/session.go#L73) acquires the `sessionsMutex` and [calls close](https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/view/services/comm/session.go#L98) on each stream when closing a session. 

With websockets, it may happen that the close call hangs when trying to write `c.reads <- streamEOF` within the `multiplexedBaseConn` implementation, see [here](https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/view/services/comm/host/rest/websocket/multiplexed_provider.go#L427). This may happen when the connection will closed before `c.reads` is consumed.

This causes the caller of close to block forever, thus, the `sessionMutex` in [NetworkStreamSession] is not released anymore.

One approach to overcome this issue is to perform a non-blocking write to `c.reads` and then complete the `close()` call.

```go
select {
case c.reads <- streamEOF:
default:
	logger.Warnf("cannot send EOF to close connection")
}
```

alexandrosfilios:
@mbrandenburger Currently we are writing tests for the comm layer on the branch `test-comm`. Let's try to reproduce and capture it in a unit test.

mbrandenburger:
https://github.com/hyperledger-labs/fabric-smart-client/pull/887 adds a test that triggers the experiences issue.

mbrandenburger:
Fixed with #909

PR #909, opened 2025-05-13
cleanup websockets multiplexing

mbrandenburger:
This PR improves the current websocket implementation and fixes the issue report in #884.

- Introduce connects killall to cleanup all goroutines
- Reduce number of goroutines for shovel messages between several layers
- Add testing

### DIFF
--- a/platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
+++ b/platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
@@ -10,6 +10,8 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	errors2 "errors"
+	"io"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -18,7 +20,6 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 	web2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/client/web"
 	host2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/host"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics"
@@ -34,13 +35,6 @@ const (
 
 type SubConnId = string
 
-var subConnId atomic.Uint64
-
-func NewSubConnId() SubConnId {
-	// Must be thread safe
-	return strconv.FormatUint(subConnId.Add(1), 10)
-}
-
 type MultiplexedMessage struct {
 	ID  SubConnId `json:"id"`
 	Msg []byte    `json:"msg"`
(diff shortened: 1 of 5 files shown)


platform/view/services/comm/host/rest/websocket/common.go
```diff
@@ -11,7 +11,7 @@ import (
 	host2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/host"
 )
 
-var logger = logging.MustGetLogger("view-sdk.services.comm.rest.host")
+var logger = logging.MustGetLogger("view-sdk.services.comm.host.rest.websocket")
```
alexandrosfilios, line 14: you don't need to pass any params anymore

platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
```diff
@@ -63,21 +77,38 @@ func NewMultiplexedProvider(tracerProvider trace.TracerProvider, metricsProvider
...
 		m: newMetrics(metricsProvider),
 	}
+
+	// spawn ActiveSubConn tracker
+	p.trackerDone = StartTracker(func() {
+		sum := 0
+		p.mu.RLock()
+		for _, c := range p.clients {
+			c.mu.RLock()
+			sum += len(c.subConns)
```
alexandrosfilios, line 87: I would avoid the locks here. You only need eventually consistent results, so I think it would be better to keep an atomic.Int for each client that keeps the length of the subConns. Then here you sum them. These results are just for tracking purposes (or for sanity checks in our tests), so using locks in production may add unnecessary overhead.

platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
```diff
@@ -150,7 +182,7 @@ func newClientConn(conn *websocket.Conn, tracer trace.Tracer, m *Metrics, onClos
 
 func (c *multiplexedClientConn) newClientSubConn(ctx context.Context, src host2.PeerID, info host2.StreamInfo) (*stream, error) {
 	c.mu.Lock()
-	sc := c.newSubConn(NewSubConnId())
+	sc := c.newSubConn(strconv.FormatUint(c.subConnId.Add(1), 10))
```
alexandrosfilios, line 185: at this point we could extract this to a method `c.newSubConnId() SubConnId` and define SubConnId as int64 instead of string to avoid the conversion. What do you think?

alexandrosfilios (commented):
Left some comments

platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
mbrandenburger: I did not touch the existing logic of the metric tracker; just moved it in order to be able to cancel the go routine - if needed. I would improve the actual tracking in another PR

platform/view/services/comm/host/rest/websocket/multiplexed_provider.go
mbrandenburger: yip, we could change this -do you think we should do this as part of this PR?
The main reason why I touched the subConIDs at all is to avoid the package variable and thus keep the scope unique per base connection.

alexandrosfilios (commented):
some comments

alexandrosfilios (approved):
LGTM but please check that it works with FTS first

mbrandenburger:
> LGTM but please check that it works with FTS first

Thanks @alexandrosfilios - I've tested this with FTS successfully. See https://github.com/hyperledger-labs/fabric-token-sdk/actions/runs/15136645966

mbrandenburger:
Let's create another issue to improve the websocket package further, for instance, by replacing mutexes by channels.

PR #926, opened 2025-05-26
Split collections functionality into multiple packages

alexandrosfilios:


platform/common/utils/collections/iterators/filter.go
```diff
@@ -0,0 +1,31 @@
...
+	next, err := it.Iterator.Next()
+	if err != nil {
+		return nil, err
+	}
+	if next == nil {
+		return nil, nil
+	}
+	if !it.filter(next) {
+		return it.Next()
+	}
```
HayimShaul, line 29: shouldn't this be a while that skips all filtered elements?
what if next is also filtered?

alexandrosfilios: and it's using tail recursion so there is no burden on the stack

HayimShaul (approved):
Looks good.
Left one comment and one question

platform/common/utils/collections/sets/sets.go
```diff
@@ -0,0 +1,76 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package sets
+
+import "github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/maps"
+
+type set[V comparable] map[V]struct{}
+
```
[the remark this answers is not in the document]
alexandrosfilios: exactly. `bool` is a bit. `struct` is none

PR #928, opened 2025-05-27
Add context to BindingStore

SaidAltury-ibm:


platform/view/endpoint.go
```diff
@@ -47,7 +49,7 @@ func NewEndpointService(es driver.EndpointService) *EndpointService {
 // If the passed identity does not have any endpoint set, the service checks
 // if the passed identity is bound to another identity that is returned together with its endpoints and public-key identifier.
 func (e *EndpointService) Resolve(party view.Identity) (view.Identity, map[PortName]string, []byte, error) {
```
alexandrosfilios, line 51: Can we add the ctx also here?

alexandrosfilios: Also for the other two methods of this struct

alexandrosfilios (approved):
LGTM modulo comments

PR #931, opened 2025-05-27
Add context to SignerInfoStore #929

SaidAltury-ibm:


platform/view/sig.go
```diff
@@ -88,12 +88,12 @@ func (s *SigService) RegisterSigner(identity view.Identity, signer Signer, verif
 
 // IsMe returns true if a signer was ever registered for the passed identity
 func (s *SigService) IsMe(identity view.Identity) bool {
-	return s.sigService.IsMe(identity)
+	return s.sigService.IsMe(context.Background(), identity)
```
alexandrosfilios, line 91: can you also add them in these two methods?

PR #932, opened 2025-05-28
F fix format

alexandrosfilios:
Removed interpreter parameter from Format method when it is not needed. We need a pagination interpreter only when we have pagination in the query.


platform/common/utils/collections/iterators/slice.go
```diff
@@ -0,0 +1,85 @@
...
+func (it *slice[T]) Next() (T, error) {
+	if !it.HasNext() {
+		return utils.Zero[T](), nil
+	}
+	item := it.items[it.i]
+	it.i++
+	return item, nil
+}
+
+func (it *slice[K]) HasNext() bool { return it.i < len(it.items) }
```
SaidAltury-ibm, line 52: slice[T] maybe?

SaidAltury-ibm (approved):
looks good, minor comments

SaidAltury-ibm (approved):
LGTM

PR #937, opened 2025-05-29
Remove protobuf

alexandrosfilios:


pkg/utils/proto/proto.go
```diff
@@ -8,7 +8,7 @@ package proto
 
 import (
 	//lint:ignore SA1019 Dependency to be updated to google.golang.org/protobuf/proto
-	protoV1 "github.com/golang/protobuf/proto" //nolint:all
+	protoV1 "google.golang.org/protobuf/proto" //nolint:all
```
mbrandenburger, line 11: I think we can also remove the comment

mbrandenburger: Don't we run into the issue that we must use `github.com/golang/protobuf/proto` for all `github.com/hyperledger/fabric-protos-go`?

adecaro:
@alexandrosfilios , what's the status of this?

PR #955, opened 2025-06-11
simpleKeyDataStore testing #948

SaidAltury-ibm:


platform/view/services/db/driver/sql/common/endorsetx.go
```diff
@@ -15,11 +15,11 @@ import (
 )
 
 func NewEndorseTxStore(writeDB WriteDB, readDB *sql.DB, table string, errorWrapper driver.SQLErrorWrapper, ci common.CondInterpreter) *EndorseTxStore {
-	return &EndorseTxStore{p: newSimpleKeyDataStore(writeDB, readDB, table, errorWrapper, ci)}
+	return &EndorseTxStore{p: NewSimpleKeyDataStore(writeDB, readDB, table, errorWrapper, ci)}
```
alexandrosfilios, line 18: First of all that's very thorough testing with all corner cases!

However, simplekeydatastore is just for deduplication. what we should test are the actual stores. so we need to have one test for endorsetx ,one for envelope and one for metadata. We could of course copy-paste the same code 3 times, but ideally the tests should also re-use the same code, because it doesn't make sense to duplicate everything. so here's what I propose:
1. you make this method and the key data store again private as they were
2. your simplekeydatatest will contain the test logic but not the tests. that is, the signatures will not be test signatures. for example:
```
type (
 getFunc func(*sql.DB, string) ([]byte, error)
 existsFunc func(*sql.DB, string) (bool, error)
 putFunc func(*sql.DB, string, []byte) error
)

func TestGetData(t *testing.T, get getFunc) {
	RegisterTestingT(t)

	db, mockDB, err := sqlmock.New()
	Expect(err).ToNot(HaveOccurred())

	key := "key1"
	data := []byte("value1")
	query := regexp.QuoteMeta("SELECT data FROM test_table WHERE key = $1")

	mockDB.
		ExpectQuery(query).
		WithArgs(key).
		WillReturnRows(mockDB.NewRows([]string{"data"}).AddRow(data))

	result, err := get(db, key)

	Expect(mockDB.ExpectationsWereMet()).To(Succeed())
	Expect(err).ToNot(HaveOccurred())
	Expect(result).To(Equal(data))
}
3. Then your actual test method in the file `endorsetx_test.go` will be
```
func TestEndorseTXGetData(t *testing.T) {
 TestGetData(t, func(db *sql.DB, key string) ([]byte, error) {return mockEndorseTXStore(db).GetEndorseTx(context.Background(), key)})
}
```


There's nothing inherently wrong with making a method public, but if this is only to accommodate a test, we should re-think why this method why was private in the first place. Was the method (simple key data store) falsely private or is the test too invasive?

PR #960, opened 2025-06-19
Move EndorserTx, Metadata, and Envelope stores under fabric where they belong #959

adecaro:
This PR includes also the following:
- NWO always generates a `default` persistence. No others.
-  sqlite also add a default configuration


alexandrosfilios (approved):
LGTM

PR #962, opened 2025-06-20
view platform: make it a pure service oriented platform #961

adecaro:
The goal of this PR is to make the `view platform` a full service oriented one. This means that we will not have the `platform/view/driver` folder as well as `platform/view/core`  and the go files in `platform/view`.
All the services will be available under `platform/view/services`. 

Moreover, the service locator pattern will only be used in the context of views and not by services to locate other services. The latter will be handled by DIG.


platform/view/sdk/dig/sdk.go
```diff
@@ -196,7 +197,7 @@ func (p *SDK) Start(ctx context.Context) error {
 		in.CommService.Start(ctx)
 
 		view3.InstallViewHandler(in.ViewManager, in.ViewService, in.TracerProvider)
-		go in.ViewManager2.Start(ctx)
+		go in.ViewManager.Start(ctx)
```
alexandrosfilios, line 200: Maybe this is the bug

mbrandenburger (approved):

Issue #967, opened 2025-07-01
chore(view_service): refactoring and clenaup


adecaro:
The goal of this Issue is to:
- Bring context.Context as first parameter of the functions getting it in input
- Introduce a way to set a timeout when running a view via `RunView`
- Remove the finality handler cause it is not needed anymore
- Streamline the use of local interfaces
- Remove unnecessary functions from `view.Manager`
- Introduce `LocalFactory` to create an instance of a view directly from a struct so to avoid any unnecessary struct marshaling.

Used by: https://github.com/hyperledger-labs/fabric-token-sdk/pull/1040

PR #968, opened 2025-07-01
view platform: cleanup view service #967

adecaro:
This PR does the following:
- Brings context.Context as first parameter of the functions getting it in input
- Introduces a way to set a timeout when running a view via `RunView`
- It removes the finality handler cause it is not needed anymore
- It streamlines the use of local interfaces
- Removes unnecessary functions from `view.Manager`
- Introduces `LocalFactory` to create an instance of a view directly from a struct so to avoid any unnecessary struct marshaling.

Used by: https://github.com/hyperledger-labs/fabric-token-sdk/pull/1040


adecaro:
This PR will be closed and the Issue will be addressed in a new PR.

Issue #970, opened 2025-07-02
Panic on close channel - delivery


thibaudgermain:
I get from time to time a panic (send on closed channel)  when stopping the delivery service:

```
panic: send on closed channel

goroutine 984 [running]:
github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/delivery.(*Delivery).Stop(...)
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/delivery/delivery.go:141
github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/delivery.(*Service).Stop(0xc001f06ea0)
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/delivery/service.go:102 +0xe0
github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic.(*Channel).Close(0xc001f06f70)
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/channel.go:62 +0x48
github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core.(*FSNProvider).Stop(0xc0021f2ba0)
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/provider.go:100 +0x180
github.com/hyperledger-labs/fabric-smart-client/platform/fabric/sdk/dig.(*SDK).PostStart.func1()
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/sdk/dig/sdk.go:135 +0x6c
created by github.com/hyperledger-labs/fabric-smart-client/platform/fabric/sdk/dig.(*SDK).PostStart in goroutine 838
	/go/src/github.com/hyperledger-labs/fabric-smart-client/platform/fabric/sdk/dig/sdk.go:133 +0x14c
```

It seems that in case of context done, we are sending a first error, then close the channel, then send a nil error in the channel. (to be confirmed)

I propose to not send anything to the err channel if err is nil:

```
func (d *Delivery) Stop(err error) {
	logger.Debugf("stop delivery with error [%v]", err)
	if err != nil {
		d.stop <- err
	}
	close(d.stop)
}
```

PR #972, opened 2025-07-02
fix: send on closed channel in delivery #970

thibaudgermain:
fixing an issue when sending nil values in channel in `delivery.go`


mbrandenburger (approved):
LGTM

Issue #981, opened 2025-07-11
Clean up logging


arner:
- Some log statements are at too high of a level. Examples:
  - `INFO [fsc.fabric.services.db.driver.sql.common] queryStatus -> SELECT tx_id, code, message FROM fsc_default__testchannel_vstatus WHERE tx_id = $1%!(EXTRA []interface {}=[configtx_0])`
  - `WARN [github.com.hyperledger-labs.fabric-token-sdk.token.services.logging.token-sdk.driver.fabtoken.default.testchannel.token-chaincode.membership.role.auditor.Auditor] newWalletService -> cannot set default identity, no identity available`
  - `WARN [fsc.view.services.view] runViewOn -> Start view CheckPublicParamsMatchView`
  - `INFO [fsc.view.sdk.dig] _ViewService_ProcessCommand_Handler -> unary call completed`
  - `WARN [fsc.fabric.services.db.driver.sql.common] convertStateRows -> Setting version of [9d6eb533296ffa4f9a45b8e03767c3b03bbc53c11a3d1916b3e3f0bef959137d0] to nil`
- Extra values (example: `queryStatus -> SELECT tx_id, code, message FROM fsc_default__testchannel_vstatus WHERE tx_id = $1%!(EXTRA []interface {}=[281440d942d51f4085e9fb1a425d556da1249ec78af2e70964eaf8b53d6e4c99])`)
- Printing bytes as %s.

mbrandenburger:
Let me add the following annoying logging to the list:
```
IgnoreError: <nil>
IgnoreError: attempt to clear closed buffer 
```

PR #994 addresses this one.

PR #986, opened 2025-07-15
Clean up logging

arner:
implements https://github.com/hyperledger-labs/fabric-smart-client/issues/981

### DIFF
--- a/platform/common/core/generic/committer/finality.go
+++ b/platform/common/core/generic/committer/finality.go
@@ -110,7 +110,7 @@ func (c *FinalityManager[V]) runStatusListener(ctx context.Context) {
 			c.logger.Debugf("check vault status for [%d] transactions", len(txIDs))
 			statuses, err := c.vault.Statuses(newCtx, txIDs...)
 			if err != nil {
-				c.logger.Errorf("error fetching statuses: %w", err)
+				c.logger.Errorf("error fetching statuses: %s", err.Error())
 				span.RecordError(err)
 				span.End()
 				continue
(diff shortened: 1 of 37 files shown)


platform/common/core/generic/committer/finality.go
```diff
@@ -110,7 +110,7 @@ func (c *FinalityManager[V]) runStatusListener(ctx context.Context) {
 			c.logger.Debugf("check vault status for [%d] transactions", len(txIDs))
 			statuses, err := c.vault.Statuses(newCtx, txIDs...)
 			if err != nil {
-				c.logger.Errorf("error fetching statuses: %w", err)
+				c.logger.Errorf("error fetching statuses: %s", err.Error())
```
adecaro, line 113: please, Arne, while updating these logs, check also if a context.Context is available and in case, use the context logger. In this case, it would be ErrorfContext(ctx,...

This, in addition to writing a log, will also generate an event in the traces.
Please, make sure that the string printed is `printable`. Arguments might need to sanitized depending on the cases.

platform/view/services/storage/driver/sql/common/simplekeydata.go
```diff
@@ -56,7 +56,7 @@ func (db *SimpleKeyDataStore) PutData(ctx context.Context, key string, data []by
 		Row(key, data).
 		OnConflictDoNothing().
 		Format()
-	logger.Debug(query, params)
+	logger.Debug(query, string(data))
```
adecaro, line 59: in this case, we don't need the DebugfContext because an event in the trace will be added by the call to ExecContext later.

adecaro (changes requested):
Hi Arne, this looks good to me modulo the comments. 
Let's take the change and use the context-based logs. 
Thanks.

Issue #338, opened 2022-08-02
NWO enable/disable node profiling


mbrandenburger:
Currently NWO starts the FSC nodes with profiling turned on by default. Profiling should be configurable via config and NWO topology.

Default: no profiling TBD

adecaro:
@mbrandenburger , can you take this on?

PR #989, opened 2025-07-15
NWO: disable profiling by default

mbrandenburger:
This PR addresses #338. NWO currently enforces profiling for every test. This PR let the user decided when to run with profiling enabled via the `FSCNODE_PROFILE=true` env var.

The profiling section in the readme explains how to do that.


adecaro (approved):
LGTM

PR #995, opened 2025-07-17
fabric-sdk: sequential events

adecaro:
This PR adds the possibility to instantiate the fabric-sdk event listener in such a way the events are fired sequentially.

This feature should avoid the glitches seen in the Token-SDK (https://github.com/hyperledger-labs/fabric-token-sdk/pull/1143). Indeed, what happens in the token sdk is the following:
- a listener is used to catch changes to the public parameters
- for each event a go routine is started
- therefore it might happen that a go routine for an event that happened later is processed before the go routine for an event that happened earlier causing misalignment.

### DIFF
--- a/platform/fabric/core/generic/events/listenermanager.go
+++ b/platform/fabric/core/generic/events/listenermanager.go
@@ -73,6 +73,7 @@ type ListenerManager[T EventInfo] struct {
 	delivery                   Delivery
 	blockProcessingParallelism int
 	ignoreBlockErrors          bool
+	sequential                 bool
 }
 
 func NewListenerManager[T EventInfo](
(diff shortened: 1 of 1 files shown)


arner:
Does this make it the default to process the events sequentially? Does that impact performance? Upgrading params is a very rare occurrence. If there is a noticeable impact on performance maybe there should be the option to override it in the config?

arner:
Ah good that you mention it, we ran into that one at some point, hard to find bug. The consumer launching a goroutine solves it indeed (if the application can handle asynchronous handlers, of course). A warning in the method doc comment could help prevent it for future users, or even to add "Sync" to its name.

PR #1003, opened 2025-07-24
Add support for extended logging in DbConfig

barvhaim:


platform/view/services/storage/driver/sql/postgres/sql_test.go
```diff
@@ -49,3 +55,86 @@ func TestPostgres(t *testing.T) {
 		return p.(*KeyValueStore).KeyValueStore
 	})
 }
+
+func verifyExtendedLoggingInCommand(t *testing.T, config *ContainerConfig, expectExtendedLogging bool) {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	require.NoError(t, err)
+
+	ctx := context.Background()
```
mbrandenburger, line 63: you can use `t.Context()` as this allows the test harness to cancel the context and all probably all consumers of the context.

platform/view/services/storage/driver/sql/postgres/sql_test.go
```diff
@@ -49,3 +55,86 @@ func TestPostgres(t *testing.T) {
...
+	require.NotNil(t, foundContainer, "Container not found")
+
+	// Inspect the container to get the command
+	inspect, err := cli.ContainerInspect(ctx, foundContainer.ID)
+	require.NoError(t, err)
+
+	// Check extended logging parameters based on expectation
+	cmd := inspect.Config.Cmd
+	if expectExtendedLogging {
+		assert.Contains(t, cmd, "log_destination=stderr")
```
mbrandenburger, line 86: I am wondering if there is a particular reason to use `assert` over `require` here. The difference is that `assert` does not call `FailNow`.

platform/view/services/storage/driver/sql/postgres/test_utils.go
```diff
@@ -181,7 +182,28 @@ func startPostgresWithLogger(c ContainerConfig, t Logger, printLogs bool) (func(
 			Timeout:  time.Second,
 			Retries:  10,
 		},
-	}, &container.HostConfig{
+	}
+
+	// Add extended logging configuration if enabled
+	if c.ExtendedLogging {
+		containerConfig.Cmd = []string{
```
mbrandenburger, line 189: The `ExtendedLogging` toggle probably makes sense for many cases and I can imagine that when debugging with postgres, sometimes more flexibility is needed when investigating an odd issue.

I think it would be nice to allow the `Cmd` be overwritten via env vars to let the user inject any postgres command / env vars via testing without touching this code all the time.

adecaro:
Closing this cause it is not needed anymore.

PR #1008, opened 2025-08-01
Interduce Cache Metrics and MetricMap

AkramBitar:
Implements CacheMetrics to measure cache usage, including put, get, delete, and update operations, as well as cache hits and misses. Introduces a MetricMap to monitor the usage of a basic key-value cache using a Go map, along with a CacheMetrics interface for tracking cache statistics


mbrandenburger:
@AkramBitar have you considered existing projects such as https://pkg.go.dev/github.com/dgraph-io/ristretto which comes with metrics out-of-the-box?

Issue #1010, opened 2025-08-04
Set Otel Logging level correctly


mbrandenburger:
The current code sets the Otel logging level to debug instead of using the parents log level
https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/common/services/logging/otel.go#L40

It seems that this was mistakenly introduced during development. However, setting it to debug by default may impact the overall fscnode performance during normal operation (i.e., log level = `info`).

mbrandenburger:
Closed with #1011

PR #1011, opened 2025-08-04
Set Otel log level based on parent logger

mbrandenburger:
This PR addresses the issue reported in #1010 

- fix hardcoded debug log level
- add log level check for expensive logging parms, such as, debug.Stack or ~base64 encoding~
- fix incorrect log caller for context based loggers

### DIFF
--- a/platform/common/services/logging/otel.go
+++ b/platform/common/services/logging/otel.go
@@ -15,7 +15,6 @@ import (
 	"go.opentelemetry.io/otel/log/noop"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 )
 
 const (
@@ -38,8 +37,7 @@ type otelLogger interface {
 func NewOtelLogger(zapLogger *zap.Logger) otelLogger {
 	return otelzap.New(zapLogger,
 		otelzap.WithLoggerProvider(newLoggerProvider(zapLogger.Name(), OtelSanitize())),
-		// otelzap.WithMinLevel(zapLogger.Level()),
-		otelzap.WithMinLevel(zapcore.DebugLevel),
+		otelzap.WithMinLevel(zapLogger.Level()),
 	).Sugar()
 }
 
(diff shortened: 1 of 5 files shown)


platform/view/services/comm/host/rest/websocket/stream.go
```diff
@@ -126,7 +126,9 @@
 		logger.Debugf("Wrote to [%s@%s], but message not ready yet.", s.info.RemotePeerID, s.info.RemotePeerAddress)
 		return n, nil
 	}
-	logger.Debugf("Ready to send to [%s@%s]: [%s]", s.info.RemotePeerID, s.info.RemotePeerAddress, logging.Base64(content))
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
```
adecaro, line 129: logging.Base64 return a struct whose String function returns the encoding. The String function is not evaluated if the log level is not debug. So, the additional check is redundant  here.

platform/common/services/logging/otel.go
```diff
@@ -36,8 +35,7 @@ type otelLogger interface {
 func NewOtelLogger(zapLogger *zap.Logger) otelLogger {
 	return otelzap.New(zapLogger,
 		otelzap.WithLoggerProvider(newLoggerProvider(zapLogger.Name())),
-		//otelzap.WithMinLevel(zapLogger.Level()),
-		otelzap.WithMinLevel(zapcore.DebugLevel),
+		otelzap.WithMinLevel(zapLogger.Level()),
```
adecaro, line 38: we could use the already existing funciton   
```go
func Init(c Config) {
flogging.Init(c)
}
```
to allow a further customization for the otel level. So that even if we have info for the terminal, we have debug for otel.

mbrandenburger: > Can we still add trace information to spans without also logging?

yes - you, you just need an instance of a span and then you create an event.

Example:
```
somespan.AddEvent("Read response")
```

arner (approved):
LGTM!

PR #1012, opened 2025-08-04
metrics harmonization: Add subsystem to metricsOpts and derive prams from path

SaidAltury-ibm:
- add subsystem
- automatically derive namespace and subsystem from path
- fix multiplexed_ws name
- namespace code cleanup


platform/view/services/tracing/config.go
```diff
@@ -21,12 +26,22 @@ type LabelName = string
...
 	Namespace  string
+	Subsystem  string
 	LabelNames []LabelName
 }
 
 func WithMetricsOpts(o MetricsOpts) trace.TracerOption {
+	subsystem := ""
+	o.Namespace, subsystem = parseFullPkgName(GetPackageName(), map[string]string{
+		"github.com_hyperledger-labs_fabric-smart-client_platform": "fsc",
+		"github.com_hyperledger-labs_fabric-token-sdk_token":       "fts",
```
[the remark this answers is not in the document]
adecaro: yes, I would leverage the package `init` function so that any project can inject replacements.

SaidAltury-ibm:
> @SaidAltury-ibm can you show how the namespace derived from the package path looks in the Jaeger UI? I was wondering if that might be a bit too verbose.
for example https://github.com/hyperledger-labs/fabric-smart-client/blob/980ad7df527b3adb4a7b1ab7a1ebd75ef9c48b91/platform/common/core/generic/vault/vault.go#L106
the namespace would be "fsc_common_core_genetic" 
the subsystem would be "vault"

integration/fabric/iou/commands.go
```diff
@@ -75,15 +75,15 @@ func CheckLocalMetrics(ii *integration.Infrastructure, user string, viewName str
 	gomega.Expect(metrics).NotTo(gomega.BeEmpty())
 
 	var sum float64
-	for _, m := range metrics["fsc_view_operations"].GetMetric() {
+	for _, m := range metrics["fsc_view_services_view_calls_operations"].GetMetric() {
```
adecaro, line 78: Please, introduce a constant for the value `"fsc_view_services_view_calls_operations"`

integration/reporting/prometheus/client.go
```diff
@@ -59,7 +59,7 @@ func (c *reporter) GetVector(metric model.Metric) (model.Vector, error) {
 }
 
 func (c *reporter) GetViewOperations(nodeName, viewName string) (int, error) {
-	m := model.Metric{"__name__": "fsc_view_operations"}
+	m := model.Metric{"__name__": "fsc_view_services_view_calls_operations"}
```
adecaro, line 62: reuse the constant here too

adecaro:
Hi @SaidAltury-ibm , we are on a very good path.

We need to do the same on the rest of the metrics. 

For example, in the code we have something like this.

```	orderedTransactions = metrics.CounterOpts{
		Namespace:    "ttx",
		Name:         "ordered_transactions",
		Help:         "The number of ordered transactions.",
		LabelNames:   []string{"network"},
		StatsdFormat: "%{#fqname}.%{network}",
	}
```

Also here, we need to make sure to have Namespace and Subsystem automatically set if not otherwise specified.
In the above case as well, the namespace is not set properly and should be set automatically. The subsystem though needs to be set to `ordering`.

platform/view/services/tracing/testtracing/config_test.go
```diff
@@ -0,0 +1,92 @@
+/*
```
adecaro, line 1: please move this file a package above, under tracing

adecaro (approved):
LGTM

Issue #1013, opened 2025-08-06
deprecate the use of github.com/pkg/errors


adecaro:
Replace it https://github.com/cockroachdb/errors by updating the package `pkg/utils/errors`.

PR #1014, opened 2025-08-06
deprecate the use of github.com/pkg/errors #1013

adecaro:


pkg/utils/errors/errors.go
```diff
@@ -38,6 +38,15 @@ func WithMessagef(err error, format string, args ...any) error {
 	return errors.WithMessagef(err, format, args...)
 }
 
+// WithMessage annotates err with the format specifier.
+func WithMessage(err error, format string) error {
+	return errors.WithMessage(err, format)
+}
```
mbrandenburger, line 44:
```suggestion
func WithMessage(err error, message string) error {
	return errors.WithMessage(err, message)
}
```

platform/view/services/storage/driver/common/base.go
```diff
@@ -50,7 +49,7 @@ func (db *BaseDB[T]) BeginUpdate() error {
 	tx, err := db.newTransaction()
 	if err != nil {
 		db.txLock.Unlock()
-		return errors2.Wrapf(err, "error starting db transaction")
+		return errors.Wrapf(err, "error starting db transaction")
```
mbrandenburger, line 52: While this was the wrong use of `errors.Wrapf` before .... Do you wanna use `errors.Wrap` here instead (and many other occurrences)?

mbrandenburger (approved):
Wonderful PR

Issue #908, opened 2025-05-13
configuration rework


adecaro:
This Github Issues asks the following:

- NWO should generated the leanest possible configuration
- the `persistences` section must contain a default
- if a section or a value is not specified it must default unless the configuration is strictly required

adecaro:
closing this as it will not be addressed in this form

PR #1025, opened 2025-08-12
add defaults, make tls root cert path relative

arner:
Part of https://github.com/hyperledger-labs/fabric-smart-client/issues/908.

- Adds defaults for keepalive (if not specified used to lead to "keep your calm" warnings in the peers)
- Only warns for no default fabric network if there are more than one
- Uses TranslatePath for relative paths for the root tls certificate for peers.


adecaro:
@arner , it looks good to me. Though, I have added @mbrandenburger and @ale-linux for additional review given that also some TLS configuration is involved.

mbrandenburger (approved):
LGTM

PR #1027, opened 2025-08-13
Remove span termination from function that should not be responsible for ending it

SaidAltury-ibm:
Previously, this function was responsible for ending the Hitman span,
even though it was not the function that created it. This caused the
span to close unexpectedly and potentially break trace continuity.
Removed the span termination here so it is handled only by the correct
owner.


adecaro (approved):
LGTM.

Is there a way to test this?
Let's wait also Marcus check.

mbrandenburger (approved):

Issue #1009, opened 2025-08-02
endpoint.Service: Bind multiple ephemeral identities at the same time


adecaro:
Currently, the endpoint service `platform/view/services/endpoint` allows us to one ephemeral identity at the time using the function
```go
Bind(ctx context.Context, longTerm view.Identity, ephemeral view.Identity) error
```
In the Token-SDK, we need to binds per pseudonym generated. This means two queries.
This Issues asks to modify the function to get in input multiple ephemeral identities in order to have only one DB query.
The suggested change is to
```go
Bind(ctx context.Context, longTerm view.Identity, ephemeralIDs ...view.Identity) error
```
This way we will have minimum disruption in the rest of the code.
The implementation of the function will check also that `len( ephemeralIDs) != 0` before proceeding.

adecaro:
@HayimShaul , once this is done, we have to bring it to the token-sdk.

PR #1033, opened 2025-08-26
feature(endpoint.Service)!: Bind multiple ephemeral identities at the same time #1009

adecaro:
The goal of this PR is to allow the binding of multiple identities at the same time so to same in the number of DB connections when binding multiple identities. This will be used in the token-sdk.

This PR is linked to FTS's PR https://github.com/hyperledger-labs/fabric-token-sdk/pull/1245

> [!NOTE]
> This PR contains a breaking change on the endpoint service API 
> `Bind(ctx context.Context, longTerm view.Identity, ephemeralIDs ...view.Identity) error`

### DIFF
--- a/platform/view/services/endpoint/service.go
+++ b/platform/view/services/endpoint/service.go
@@ -146,18 +146,20 @@ func (r *Service) GetResolver(ctx context.Context, id view.Identity) (*Resolver,
 	return r.resolver(ctx, id)
 }
 
-func (r *Service) Bind(ctx context.Context, longTerm view.Identity, ephemeral view.Identity) error {
-	if longTerm.Equal(ephemeral) {
-		logger.DebugfContext(ctx, "cannot bind [%s] to [%s], they are the same", longTerm, ephemeral)
+func (r *Service) Bind(ctx context.Context, longTerm view.Identity, ephemeralIDs ...view.Identity) error {
+	// filter out any identities equal to the longTerm identity
+	var toBind []view.Identity
+	for _, id := range ephemeralIDs {
+		if !longTerm.Equal(id) {
+			toBind = append(toBind, id)
+		}
+	}
+	if len(toBind) == 0 {
 		return nil
 	}
-
-	logger.DebugfContext(ctx, "bind [%s] to [%s]", ephemeral, longTerm)
-
-	if err := r.bindingKVS.PutBinding(ctx, ephemeral, longTerm); err != nil {
-		return errors.WithMessagef(err, "failed storing binding of [%s]  to [%s]", ephemeral.UniqueID(), longTerm.UniqueID())
+	if err := r.bindingKVS.PutBindings(ctx, longTerm, toBind...); err != nil {
+		return errors.WithMessagef(err, "failed storing bindings")
 	}
-
 	return nil
 }
 
--- a/platform/view/services/storage/driver/sql/common/binding.go
+++ b/platform/view/services/storage/driver/sql/common/binding.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils"
@@ -19,6 +20,8 @@ import (
 	cond2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/query/cond"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 )
+
+const BindingStoreMaxEphemerals = 1000
 
 func NewBindingStore(readDB *sql.DB, writeDB WriteDB, table string, errorWrapper driver.SQLErrorWrapper, ci common.CondInterpreter) *BindingStore {
 	return &BindingStore{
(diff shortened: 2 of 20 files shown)


platform/view/services/storage/driver/sql/postgres/binding_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	common3 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common"
+	common2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPutBindings_MultipleEphemerals(t *testing.T) {
```
mbrandenburger, line 18:
```suggestion
func TestPutBindingsMultipleEphemerals(t *testing.T) {
```

platform/view/services/storage/driver/sql/sqlite/binding_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	common3 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common"
+	common2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPutBindings_MultipleEphemerals(t *testing.T) {
```
mbrandenburger, line 18:
```suggestion
func TestPutBindingsMultipleEphemerals(t *testing.T) {
```

platform/view/services/storage/driver/sql/sqlite/binding.go
```diff
@@ -48,25 +56,31 @@ func (db *BindingStore) PutBinding(ctx context.Context, ephemeral, longTerm view
...
+	for i, eph := range ephemeral {
+		args = append(args, eph.UniqueID(), longTerm)
+		oneArgRef := fmt.Sprintf("($%d, $%d)", i*2+3, i*2+4)
+		argsReferences = append(argsReferences, oneArgRef)
+	}
+
+	query += strings.Join(argsReferences, ", ")
+	query += " ON CONFLICT DO NOTHING;"
+
+	logger.DebugfContext(ctx, "executing bulk insert: %s", query)
```
mbrandenburger, line 75: I am wondering if this should be really a debug message that also appears as tracing events. 

This is a general questions for all logger.InfofContext in this PR. I would add tracing events only if you would gain additional knowledge in order to follow / understand the timing in traces. I think the traces should not be used as general log.

mbrandenburger (commented):
Thanks again for addressing my comments.

platform/view/services/endpoint/service.go
```diff
@@ -146,18 +147,14 @@ func (r *Service) GetResolver(ctx context.Context, id view.Identity) (*Resolver,
...
 
-func (r *Service) Bind(ctx context.Context, longTerm view.Identity, ephemeral view.Identity) error {
-	if longTerm.Equal(ephemeral) {
-		logger.DebugfContext(ctx, "cannot bind [%s] to [%s], they are the same", longTerm, ephemeral)
-		return nil
+func (r *Service) Bind(ctx context.Context, longTerm view.Identity, ephemeralIDs ...view.Identity) error {
+	// filter out the ids that are equal to longTerm
+	toBind := collections.FilterSlice(ephemeralIDs, func(id view.Identity) bool {
+		return !longTerm.Equal(id)
+	})
```
mbrandenburger: if `toBind` is empty, should we `return` here early with a proper error message?

platform/view/services/storage/driver/sql/postgres/binding_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"testing"
+
+	testing2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common/testing"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPutBindingsMultipleEphemerals(t *testing.T) {
```
mbrandenburger:
```suggestion
func TestPutBindingsMultipleEphemeralsPostgres(t *testing.T) {
```

Let me suggest to add Postgres to the test name to indicate this test runs with Postgres. I am suggesting this as I am currently trying to "tidy" up the postgres tests a bit in #1069.

platform/view/services/storage/driver/sql/sqlite/binding.go
```diff
@@ -48,25 +56,31 @@ func (db *BindingStore) PutBinding(ctx context.Context, ephemeral, longTerm view
...
+	argsReferences := []string{"($1, $2)"}
+	args = append(args, longTerm.UniqueID(), longTerm)
+	for i, eph := range ephemeral {
+		args = append(args, eph.UniqueID(), longTerm)
+		oneArgRef := fmt.Sprintf("($%d, $%d)", i*2+3, i*2+4)
+		argsReferences = append(argsReferences, oneArgRef)
+	}
+
+	query += strings.Join(argsReferences, ", ")
+	query += " ON CONFLICT DO NOTHING;"
```
mbrandenburger: we have the same query in the postgres impl; could we consider moving it to `sql/common/`?

platform/view/services/storage/driver/sql/sqlite/binding.go
```diff
@@ -48,25 +56,31 @@ func (db *BindingStore) PutBinding(ctx context.Context, ephemeral, longTerm view
...
-		ON CONFLICT DO NOTHING;
-		COMMIT;
-		`, db.table, db.table)
 
-	logger.Debug(query, ephemeral.UniqueID(), longTerm.UniqueID())
-	_, err := db.writeDB.ExecContext(ctx, query, ephemeral.UniqueID(), longTerm, longTerm.UniqueID(), longTerm)
+	// Build single INSERT with multiple VALUES
+	query := fmt.Sprintf(`INSERT INTO %s (ephemeral_hash, long_term_id) VALUES `, db.table)
+
+	args := []interface{}{}
```
mbrandenburger:
```suggestion
	var args := []any
```

platform/view/services/storage/driver/sql/postgres/binding.go
```diff
@@ -38,29 +39,38 @@ func newBindingStore(readDB, writeDB *sql.DB, table string) *BindingStore {
...
+	argsReferences := []string{"($1, $2)"}
+	args = append(args, longTerm.UniqueID(), longTerm)
+	for i, eph := range ephemerals {
+		args = append(args, eph.UniqueID(), longTerm)
+		oneArgRef := fmt.Sprintf("($%d, $%d)", i*2+3, i*2+4)
+		argsReferences = append(argsReferences, oneArgRef)
+	}
+
+	query += strings.Join(argsReferences, ", ")
+	query += " ON CONFLICT DO NOTHING;"
```
mbrandenburger: this part is not trivial 🤯

platform/view/services/storage/driver/sql/postgres/binding.go
```diff
@@ -38,29 +39,38 @@ func newBindingStore(readDB, writeDB *sql.DB, table string) *BindingStore {
...
-	logger.DebugfContext(ctx, "Put binding for pair [%s:%s]", ephemeral.UniqueID(), longTerm.UniqueID())
-	if lt, err := db.GetLongTerm(ctx, longTerm); err != nil {
-		return err
-	} else if lt != nil && !lt.IsNone() {
-		logger.DebugfContext(ctx, "Replacing [%s] with long term [%s]", longTerm.UniqueID(), lt.UniqueID())
-		longTerm = lt
-	} else {
-		logger.DebugfContext(ctx, "Id [%s] is an unregistered long term ID", longTerm.UniqueID())
+
+func (db *BindingStore) PutBindings(ctx context.Context, longTerm view.Identity, ephemerals ...view.Identity) error {
```
mbrandenburger: is there a limit of max number of ephemerals we should consider here?

mbrandenburger (changes requested):
Thanks for this PR! It is in a good shape already. See below a few minor comments to address. Thanks again!

platform/view/services/storage/driver/sql/postgres/binding.go
HayimShaul: sounds right. Any suggestions for a limit?

platform/view/services/storage/driver/sql/sqlite/binding.go
mbrandenburger: sorry - this was a sloppy suggestion. it should be `var args []any`.

platform/view/services/endpoint/service.go
mbrandenburger: you are right, probably not, as we also just return nil if the ephemerals are empty.

platform/view/services/storage/driver/sql/common/binding.go
```diff
@@ -94,6 +94,9 @@ func (db *BindingStore) PutBindings(ctx context.Context, longTerm view.Identity,
 	if len(ephemerals) == 0 {
 		return nil
 	}
+	if len(ephemerals) > 1000 {
+		return errors.Errorf("Too many ephemerals (%d). Max allowed is 1000", len(ephemerals))
+	}
```
mbrandenburger: Please update the go docs for this function to clearly describe that here is a limit.

platform/view/services/storage/driver/sql/postgres/binding_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"testing"
+
+	testing2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common/testing"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPutBindingsMultipleEphemeralsPostgres(t *testing.T) {
```
mbrandenburger: What about adding one case to test our limit :)

AkramBitar (approved):

PR #1034, opened 2025-08-28
Fix comm stack resource consumption

mbrandenburger:
The current stack suffers from high resource utilization (active number of goroutines and increasing memory usage) when invoking views through the p2p comm layer.
 
This PR tames the resource consumption.

- set reasonable read buffer size
- cleanup subcon when they are closed
- more efficient sessionID computation
- remove delayed subcon close
- remove delayed context deletion after view execution

### DIFF
--- a/platform/view/services/comm/host/rest/host.go
+++ b/platform/view/services/comm/host/rest/host.go
@@ -9,8 +9,8 @@ package rest
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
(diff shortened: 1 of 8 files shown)


mbrandenburger:
This screenshot shows the memory consumption (memory leak) and exploding number of goroutines with and without the fix in this PR.

Screenshot 2025-08-29 at 09 07 10

adecaro:
@mbrandenburger , wow, super interesting. Only the rate of objects allocated got worse. Do I read it correctly?

mbrandenburger:
> @mbrandenburger , wow, super interesting. Only the rate of objects allocated got worse. Do I read it correctly?

No no, not worse. It increased was the FSC node was not able to process at a higher rate. Here is a screen shot of TPS/latency for the view invocation.

You can see, with the fix in this PR, the TPS ~doubled and the latency remains stable. Thus, this the node is processing more invocations, the object allocation rate is highter.

Screenshot 2025-08-29 at 09 07 00

ale-linux (approved):
Awesome stuff @mbrandenburger! Only issue I see is that getting rid of the debug printouts might bite is in the ass later... how about we leave them encased in an `if` that checks whether the debug level is enabled?

arner (approved):
Nice, looks like a big improvement!

PR #1045, opened 2025-09-08
fix(prometheus): adjust runtime.Caller depth in GetPackageName

### LINKED ISSUE #1044: GetPackageName returns wrong package name due to incorrect runtime.Caller depth
GetPackageName uses runtime.Caller(2) to retrieve the package name of the caller.
However, this depth is incorrect — it returns the package of the helper function itself rather than the actual caller.

This leads to wrong namespace/subsystem values being applied when creating Prometheus metrics.

- **Steps to Reproduce:**
Call Provider.NewCounter (or any metric constructor).
Observe the generated metric namespace/subsystem.
The values are not resolved to the caller’s package as intended.

- **Expected Behavior:**
GetPackageName should correctly return the caller’s package name so that metrics are labeled consistently.

- **Root Cause:**
runtime.Caller(2) skips only two stack frames, which still points to the Prometheus provider internals.
The correct depth is runtime.Caller(3) to reach the actual external caller.

- **Fix:**
Increase depth from 2 → 3 in GetPackageName.

SaidAltury-ibm:
The GetPackageName function was returning the wrong package name because
runtime.Caller was using depth=2. Increased it to depth=3 so that it correctly
resolves the caller package instead of the helper function itself.
fix #1044

### DIFF
--- a/platform/view/services/metrics/prometheus/provider.go
+++ b/platform/view/services/metrics/prometheus/provider.go
@@ -22,6 +22,15 @@ type Provider struct {
 	SkipRegisterErr bool
 }
 
+const (
+	// callerSkipFrames represents the number of stack frames to skip when determining
+	// the calling package name. This skips: (1) GetPackageName itself, (2) the method
+	// calling GetPackageName (e.g., NewCounter/NewGauge/NewHistogram), and (3) the
+	// applyNamespaceSubsystem method, to reach the actual caller that initiated the
+	// metrics creation.
+	callerSkipFrames = 3
+)
+
 var (
 	replacersMutex sync.RWMutex
 	replacers      = map[string]string{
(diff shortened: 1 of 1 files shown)


platform/view/services/metrics/prometheus/provider.go
```diff
@@ -98,7 +98,7 @@ func (p *Provider) NewHistogram(o metrics.HistogramOpts) metrics.Histogram {
 }
 
 func GetPackageName() string {
-	pc, _, _, ok := runtime.Caller(2)
+	pc, _, _, ok := runtime.Caller(3)
```
adecaro, line 101: please, replace 3 with a const and add a comment to it to justify the value. Thanks.

SaidAltury-ibm: Fixed - replaced 3 with callerSkipFrames constant. Added comment explaining that it skips 3 frames: GetPackageName itself, the metrics creation method, and applyNamespaceSubsystem to reach the actual caller's package.

adecaro (approved):
LGTM

PR #1048, opened 2025-10-01
Add fabric-x support

mbrandenburger:
This PR adds a new platform for Fabric-x


adecaro (approved):
LGTM. As agrred, we will restructure the documentation in another PR.

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/hyperledger-labs/fabric-smart-client
 
-go 1.24.2
-
-toolchain go1.24.3
+go 1.24.3
```
arner, line 3: Ok for FSC. Do you think we can keep FTS on go 1.24.2? For backwards compatibility with letting the peer build chaincode on Fabric 2 and 3? 😵 

```shell
docker run --rm hyperledger/fabric-ccenv:latest go version
go version go1.24.2 linux/arm64
```

mbrandenburger: Also note that the fsc integration tests build chaincode using `hyperledger/fabric-ccenv` and it seems to work without an issue. I guess you only run into an issue once the chaincode uses new language features which are not yet supported in fabric-ccenv (go1.24.2)

PR #1061, opened 2025-10-22
fabricx: remove decoration 

adecaro:
Currently, the FabricX SDK decorates the host provider and the config provider. Both of them are not necessary.
In order to remove the config provider decoration, NWO has been updated to inject a peer definition that can be used not only for delivery but for any other function of the fabric platform (query, finality, and so on).


mbrandenburger (approved):
LGTM

PR #1065, opened 2025-10-29
fix: createLockTag returns int64

mbrandenburger:
This PR fixes a bug introduced https://github.com/hyperledger-labs/fabric-smart-client/commit/ea5aa2095c1dab12b0130b08f6db325803df013c#diff-65733d2b46f196197a0740de95a8e1ba1ff8cc35fba42e7d967e8b52a92d8b3dR217. The `createLockTag` must return a `int64` value as `pg_advisory_xact_lock(bigint)` requires `bigint`. Passing a `uint64` may not fit into a signed 64-bit integer.


mbrandenburger:
Testing this one with the TokenSDK seems to resolve the issue https://github.com/mbrandenburger/fabric-token-sdk/actions/runs/18920031033/job/54013662950?pr=4

Issue #1066, opened 2025-10-30
Enabling postgres unit/integration tests on CI


mbrandenburger:
It seems that currently `platform/view/services/storage/driver/sql/postgres/sql_test.go` is not run by CI.

Task:
- ensure test works properly 
- create make target for "long" unit tests such as those which use postgres
- include new make target into GHA

PR #1069, opened 2025-10-31
test(postgres): enable postgres unit tests in CI

mbrandenburger:
This PR addresses #1066 

- create make target for postgres unit tests
- all unit tests run by default with -race; this can be locally controlled via GO_TEST_PARAMS if needed

Update:
With enabled unit tests, the current impl of postgres db seem to be a bit fragile and not align well with the sqlite impl.
The OG impl of the postgres db, @arner choose pkey to be BYTEA, later, this got changed to TEXT. It's not clear to me what was the reason for that. However, as a result extra key transformation/encoding/decoding had to happen to support `\0x00` runes and general non-UTF8s pkeys. This PR changes the pkey type back to BYTEA. Testing with the TokenSDK (see link below) reem to work ok.

Update:
Additional changes on the postgres notifier were needed to fix race issues. However, since this component seem to be used exclusively, we propose to deprecate it and move the it to the Token SDK.

### DIFF
--- a/platform/view/services/storage/driver/sql/common/bench.go
+++ b/platform/view/services/storage/driver/sql/common/bench.go
@@ -16,6 +16,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	namespace = "ns"
+	key       = "key"
+)
+
 var (
 	returnValue []byte
 	returnErr   error
--- a/platform/view/services/storage/driver/sql/common/helpers.go
+++ b/platform/view/services/storage/driver/sql/common/helpers.go
@@ -9,20 +9,16 @@ package common
 import (
 	"context"
 	"database/sql"
-	"encoding/binary"
 	"fmt"
-	"strings"
 	"sync"
 	"testing"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -108,20 +108,27 @@ testing-docker-images: ## Pull images for system testing
...
+	export FABRIC_LOGGING_SPEC=error; \
+	export FAB_BINS=$(FAB_BINS); \
+	go test $(GO_TEST_PARAMS) --skip '(Postgres)' $(GO_PACKAGES)
 
-.PHONY: unit-tests-race
-unit-tests-race: ## Run unit tests with race
-	@export GORACE=history_size=7; export FAB_BINS=$(FAB_BINS); go test -race -cover $(shell go list ./... | grep -v '/integration/')
-	cd integration/nwo/; export FAB_BINS=$(FAB_BINS); go test -race -cover ./...
+.PHONY: unit-tests-postgres
+unit-tests-postgres: ## Run unit tests for postgres (requires container images as defined in testing-docker-images)
```
AkramBitar: Which target to run SQLite unti-tests? 
Maybe we need to have one target in makefile that test both postgres and Sqlite

adecaro (approved):
LGTM. 
Thanks for pushing this.

PR #1074, opened 2025-11-07
refactor!: Remove consul sdk dep

mbrandenburger:
This PR refactors the postgres container code. In particular, it removes a consul sdk dep to pick a free port for the postgres instances we run within NWO and unit-tests. We let NWO select a free container port or let the container runtime select the port during unit-tests.

This PR contributes to #828 and #434

Overview:
- let NWO select the ports for the postgres containers
- in unit-tests, the container runtime selects the port
- introduce options for `postgres.DefaultConfig`

> [!NOTE]
> This PR changes the API of `postgres.DefaultConfig`, which is also used by the TokenSDK for testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.3
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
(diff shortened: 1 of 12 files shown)


SaidAltury-ibm (approved):
This is neat! great work LGTM

PR #1076, opened 2025-11-11
bug fix: view and broadcast context

adecaro:
This PR fixes the following two bugs:
- When calling `InitiateViewWithIdentity`, the passed context was ignored.
- When calling `Broadcast` on a fabric `AtomicBroadcastClient`, the passed context is request specific but the stream is shared. The context to obtain a broadcast stream should be request agnostic.

This PR is linked to https://github.com/hyperledger-labs/fabric-token-sdk/pull/1263


arner (approved):
LGTM

Issue #1077, opened 2025-11-12
fabric/services/endorser: better support for unit-testing


adecaro:

mbrandenburger:
Completed with https://github.com/hyperledger-labs/fabric-smart-client/pull/1078

PR #1078, opened 2025-11-12
refactor(fabric/services/endorser): better support for unit-testing #1077

adecaro:
The goal of this PR is to add constructors to allow the developer to construct mocks more easily.

This PR is connected to https://github.com/hyperledger-labs/fabric-token-sdk/pull/1265.

### DIFF
--- a/platform/fabric/chaincode.go
+++ b/platform/fabric/chaincode.go
@@ -15,36 +15,18 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 )
 
+// Envelope models a fabric envelope
 type Envelope struct {
-	e driver.Envelope
+	driver.Envelope
 }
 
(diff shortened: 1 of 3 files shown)


platform/fabric/services/endorser/builder.go
```diff
@@ -12,25 +12,24 @@ import (
...
+	sp services.Provider
 }
 
-func NewBuilder(context view.Context) *Builder {
+func NewBuilder(context services.Provider) *Builder {
 	if context == nil {
 		panic("context must be set")
 	}
 	return &Builder{sp: context}
 }
```
mbrandenburger:
```suggestion
func NewBuilder(sp services.Provider) *Builder {
if sp == nil {
panic("service provider must be set")
}
return &Builder{sp: sp}
}
```

PR #1085, opened 2025-11-23
refactor: ability to run a view with a different context

adecaro:
This PR introduces a new option for `Context#RunView`, WithContext. WithContext allows the developer to specify a `context.Context` to use when running the view. This is very helpful when the context needs to be enhanced with additional values, for example.

Tested in the token-sdk: https://github.com/hyperledger-labs/fabric-token-sdk/pull/1270

### DIFF
--- a/platform/view/services/view/view.go
+++ b/platform/view/services/view/view.go
@@ -79,7 +79,7 @@ func AsInitiatorView(context view.Context, initiator View) (interface{}, error) 
 
 // RunViewNow invokes the Call function of the given view.
 // The view context used to the run view is either ctx or the one extracted from the options (see view.WithContext), if present.
-func RunViewNow(ctx ParentContext, v View, opts ...view.RunViewOption) (res interface{}, err error) {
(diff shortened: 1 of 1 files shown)


platform/view/services/view/mock/runner.go
```diff
@@ -25,61 +26,61 @@ type Responders struct {
 	Channel       *session.LocalBidirectionalChannel
 }
 
-type MockContext struct {
+type DelegatedContext struct {
```
mbrandenburger: I am wondering if we can move the `DelegatedContext` into the integration test where it is exclusively used, and even make it private.

platform/view/view/context.go
```diff
@@ -18,6 +18,7 @@ type RunViewOptions struct {
 	AsInitiator bool
 	Call        func(Context) (interface{}, error)
 	SameContext bool
+	Context     context.Context
```
mbrandenburger:
```suggestion
	Ctx     context.Context
```

PR #1094, opened 2025-11-28
enhancement: implement multiton ListenerManager

SaidAltury-ibm:
- Introduced Provider that automatically manages ListenerManager lifecycle
- Reuse manager instances per network:channel combination
- listen() is now private and runs automatically in a background goroutine
- Automatic cleanup when context is canceled or stream errors occur
- increase test coverage


platform/fabricx/core/finality/listenermanager.go
```diff
@@ -64,9 +88,13 @@ func newNotifi(ctx context.Context, cfg config.ConfigService) (*notificationList
 	if err != nil {
 		return nil, err
 	}
+	// CRITICAL FIX: Use context.Background() for the stream lifecycle.
+	// This ensures the stream stays alive for the duration of the shared manager,
+	// regardless of when the individual view's context (ctx) is cancelled.
+	streamCtx := context.Background()
```
mbrandenburger: Good catch. We don't want our notification listener attached to the view's lifecycle. Instead, we want some context that is associated with the lifecycle of the fsc node. That is, when we shutdown the fsc node, the notification listener is stopped gracefully.

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -32,15 +34,28 @@ func NewListenerManagerProvider(fnsp *fabric.NetworkServiceProvider, configProvi
...
 		configProvider: configProvider,
+		managers:       make(map[string]ListenerManager),
 	}
 }
 
 type listenerManagerProvider struct {
 	fnsp           *fabric.NetworkServiceProvider
 	configProvider config.Provider
+	managers       map[string]ListenerManager // Cache: "network:channel" -> ListenerManager
+	lock           sync.Mutex
```
mbrandenburger:
```suggestion
	managersMu           sync.Mutex
```

platform/fabricx/core/finality/listenermanager.go
mbrandenburger: Maybe we provide implement a start/stop interface for `notificationListenerManager` instead of `Listen()`?

platform/fabricx/core/finality/nlm_test.go
```diff
@@ -56,26 +56,65 @@ func (m *mockListener) getStatus() (string, int) {
...
+// runManager starts the manager asynchronously and ensures Stop() is called on test cleanup.
+func runManager(t *testing.T, nlm *notificationListenerManager, managerCancel context.CancelFunc) {
+	t.Helper()
+
+	// Start the manager asynchronously
+	errs := make(chan error, 1)
+	go func() {
+		errs <- nlm.Start()
+	}()
+
```
mbrandenburger: that is an interesting test. I think it is not guaranteed that `nlm.Start()` is actually called. The go rountine might not start in time, so `t.Cleanup()` gets called before I believe!?!?

platform/fabricx/sdk/dig/sdk.go
```diff
@@ -66,5 +66,13 @@ func (p *SDK) Install() error {
 }
 
 func (p *SDK) Start(ctx context.Context) error {
+	err := p.Container().Invoke(func(provider finality.ListenerManagerProvider) error {
+		provider.Initialize(ctx)
+		return nil
+	})
```
mbrandenburger: this is nice!

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -40,14 +41,19 @@ type listenerManagerProvider struct {
...
+	managersMu     sync.Mutex
+	baseCtx        context.Context
 }
 
-func (p *listenerManagerProvider) NewManager(ctx context.Context, network, channel string) (ListenerManager, error) {
+func (p *listenerManagerProvider) Initialize(ctx context.Context) {
+	p.baseCtx = ctx
+}
+
+func (p *listenerManagerProvider) NewManager(network, channel string) (ListenerManager, error) {
```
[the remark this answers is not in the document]
mbrandenburger: Could we just check that `p.baseCtx != nil`, otherwise we panic as  this clearly a programming error? Not yet sure about it.

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -64,21 +70,27 @@ func (p *listenerManagerProvider) NewManager(ctx context.Context, network, chann
...
 	// Start listening in background
 	// Listen() is a blocking method
 	go func() {
 		if err := lm.listen(); err != nil && !errors.Is(err, context.Canceled) {
 			logger.Errorf("Notification listener stream terminated unexpectedly: %s", err)
 		}
+		// Clean up the manager from the map when listen exits
+		p.managersMu.Lock()
+		delete(p.managers, key)
+		p.managersMu.Unlock()
```
mbrandenburger: Maybe we should add a test so make sure that `p.managers` is eventually empty once we shutdown.

mbrandenburger (commented):
Hi @SaidAltury-ibm Great progress on the impl.

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -18,29 +19,63 @@ import (
...
+	initOnce       sync.Once
 }
 
-func (p *listenerManagerProvider) NewManager(ctx context.Context, network, channel string) (ListenerManager, error) {
+// Initialize sets the base context for the provider. This method is idempotent and should be
+// called once during setup.
+func (p *Provider) Initialize(ctx context.Context) {
+	p.initOnce.Do(func() {
+		p.baseCtx = ctx
+		logger.Infof("Provider initialized with base context")
```
mbrandenburger: Let's make this debug logging.

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -18,29 +19,63 @@ import (
...
+func NewListenerManagerProvider(fnsp *fabric.NetworkServiceProvider, configProvider config.Provider) *Provider {
+	return &Provider{
 		fnsp:           fnsp,
 		configProvider: configProvider,
+		managers:       make(map[string]ListenerManager),
 	}
 }
 
-type listenerManagerProvider struct {
+type Provider struct {
```
mbrandenburger: I believe it would be nice to describe the provider and how it is supposed to be used (call Initialize before NewManager) ...

platform/fabricx/core/finality/listenermanager.go
mbrandenburger: We could remove `notifyStream` and instead save `notifyClient`. In `listen(ctx context.Context)` we create the stream.

```go
notifyStream, err := notifyClient.OpenNotificationStream(ctx)
```

mbrandenburger (commented):
@SaidAltury-ibm Thanks for the latest updates. I am wondering if we can simplify the nlm by removing the `ctx` and `cancelF`. Please see my comments. THanks

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -32,35 +34,45 @@ type ListenerManagerProvider interface {
...
 	return &Provider{
 		fnsp:           fnsp,
 		configProvider: configProvider,
 		managers:       make(map[string]ListenerManager),
+		// Note: baseCtx and initOnce will be initialized in the Initialize method.
 	}
 }
 
+// Provider implements ListenerManagerProvider and manages ListenerManager instances.
+// IMPORTANT: Initialize method MUST be called once during service setup before calling NewManager mothod.
```
mbrandenburger: typo: method

platform/fabricx/core/finality/listenermanager.go
```diff
@@ -86,35 +100,44 @@ func (p *Provider) NewManager(network, channel string) (ListenerManager, error)
 		return nil, err
 	}
 
-	lm, err := newNotifi(p.baseCtx, cfg)
+	// Create a listen context that does not inherit the cancellation of the baseCtx.
+	// This is necessary because the baseCtx might be cancelled by a different component,
+	// but the listener should only stop if it's explicitly cancelled (e.g., during shutdown).
+	listenCtx := context.WithoutCancel(p.baseCtx)
```
mbrandenburger: this is not clear to me. I thought we use the `baseCtx` to shutdown all managers if we shutdown the FSC node?

platform/fabricx/core/finality/nlm_test.go
```diff
@@ -390,47 +392,48 @@ func TestNotificationListenerManager(t *testing.T) {
...
+			listenErr <- nlm.listen(ctx)
 		}()
 
 		// wait briefly to ensure all goroutines are up and blocking
 		time.Sleep(shortWait)
 
-		// Call Shutdown(), which cancels the context
-		nlm.Shutdown()
+		// Cancel the context (equivalent to Shutdown)
+		cancel()
```
mbrandenburger: I am wondering, if this is the only test where we manually call cancel() to test the shutdown behaviour, should we create the cancelable context just here in this test.

For example:

```go
t.Run("Shutdown_Graceful_Exit", func(t *testing.T) {
		t.Parallel()

		ctx, cancel := context.WithCancel(t.Context())
		nlm, fakeStream := setupTest(t, ctx)

		// mock Recv to block indefinitely on context
		fakeStream.RecvStub = func() (*protonotify.NotificationResponse, error) {
			<-ctx.Done()
			return nil, ctx.Err()
		}

		// Start listen() in a goroutine
		listenErr := make(chan error, 1)
		go func() {
			listenErr <- nlm.listen(ctx)
		}()

		// wait briefly to ensure all goroutines are up and blocking
		time.Sleep(shortWait)

		// Cancel the context (equivalent to Shutdown)
		cancel()
```

And in all other cases where we dont care about manually canceling ... we could just do ... 

```go
		const targetTxID = "tx_duplicate_listener"
		ctx := t.Context()
		nlm, fakeStream := setupTest(t, ctx)

		fakeStream.RecvStub = func() (*protonotify.NotificationResponse, error) {
			<-ctx.Done()
			return nil, ctx.Err()
		}

		runManager(t, nlm, ctx)

		ml := &mockListener{}
```

WDYT?

platform/fabricx/core/finality/provider.go
```diff
@@ -41,14 +41,16 @@ func NewListenerManagerProvider(fnsp *fabric.NetworkServiceProvider, configProvi
 		fnsp:           fnsp,
 		configProvider: configProvider,
 		managers:       make(map[string]ListenerManager),
+		initOnce:       sync.Once{},
```
mbrandenburger: this is not needed ... the zero-value of `sync.Once` is valid. Note that is similar to a `sync.Mutex`.

platform/fabricx/core/finality/provider_test.go
```diff
@@ -0,0 +1,460 @@
...
+		
+		manager3, err3 := provider.NewManager("network2", "channel1")
+		require.NoError(t, err3)
+		
+		// All should be different instances
+		assert.NotSame(t, manager1, manager2, "Different channels should have different managers")
+		assert.NotSame(t, manager1, manager3, "Different networks should have different managers")
+		assert.NotSame(t, manager2, manager3, "Different combinations should have different managers")
+		
+		assert.Equal(t, int32(3), createCount.Load(), "Should create 3 different managers")
```
mbrandenburger: Would it be useful to also check that `len(provider.managers) == 3`?

mbrandenburger (commented):
Beautiful! @SaidAltury-ibm A few nits ... I think we are almost done. Thanks

PR #1101, opened 2026-01-03
feat: Add TokenX integration and fix FabricX transaction mismatch

gutama:


gutama:
DCO check passed, thank you

adecaro:
I'm wondering if the TokenX app should be better positioned in the fabric-samples rather than here.
@gutama @mbrandenburger , what do you think?

integration/fabricx/tokenx/SPECIFICATION.md
```diff
@@ -0,0 +1,1008 @@
+# TokenX - Token Management System Specification
```
mbrandenburger: I think this a great spec! However, I have the feeling it's a bit beyond the scope of an integration test. I could think about create a dedicated lab to continue development of this application, or work towards a "fabric-x" sample. What do you think?

mbrandenburger:
> I'm wondering if the TokenX app should be better positioned in the fabric-samples rather than here. @gutama @mbrandenburger , what do you think?

I agree, this should be placed eventually in fabric-x-samples. Closing.

PR #1110, opened 2026-01-08
Refactor fabric imports

mbrandenburger:
This PR reduces FSC’s dependency on `github.com/hyperledger/fabric` by refactoring and moving the required Fabric code into `platform/fabric`. Fabric is not designed or versioned as a reusable Go library, and depending on its internal packages introduces instability, misleading dependency alerts, and maintenance challenges.

By internalizing the required functionality, FSC gains fine-grained control over the code it depends on, at the cost of needing to explicitly track and maintain compatibility with Fabric protocol changes. This is a deliberate trade-off.

Additionally, removing direct Fabric dependencies eliminates false-positive Dependabot alerts caused by Fabric’s non-semantic Go module versioning, which incorrectly associates current Fabric releases with older versions such as 1.4 or 2.2.

TODOs:
- [x] github.com/hyperledger/fabric/core/chaincode/platforms/util
- [x] github.com/hyperledger/fabric/protoutil
- [x] github.com/hyperledger/fabric/discovery/client
- [x] github.com/hyperledger/fabric/core/common/ccprovider
- [x] github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/rwsetutil
- [x] github.com/hyperledger/fabric/common/configtx
- [x] github.com/hyperledger/fabric/common/channelconfig
- [x] github.com/hyperledger/fabric/msp
- [x] github.com/hyperledger/fabric/core/ledger/pvtdatastorage
- [x] github.com/hyperledger/fabric/core/ledger/kvledger/txmgmt/statedb/statecouchdb

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/gorilla/websocket v1.5.3
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
-	github.com/hyperledger/fabric v1.4.0-rc1.0.20250510200036-435a7f1a780a
 	github.com/hyperledger/fabric-chaincode-go/v2 v2.3.0
 	github.com/hyperledger/fabric-config v0.3.0
 	github.com/hyperledger/fabric-contract-api-go/v2 v2.2.0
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

mbrandenburger:
> One suggestion: Could we move `platform/fabric/msp` and `platform/fabric/protoutils` under `plaftorm/fabric/core`? The rational being that these packages are used by the `core` package alone and we don't recommend the developers to access them directly.

@adecaro good idea - I moved the packages as you suggested.

AkramBitar (commented):
So far so good :-)
LGTM

PR #1121, opened 2026-01-20
Benchmarks: add multiGRPC connection benchmark, add ECDSA on GRPC benchmarks and improve Plots

SaidAltury-ibm:
This PR improves the benchmark suite to better reflect real throughput limits and isolate different bottlenecks:

1. Multi-connection gRPC benchmark:

      - Adds a variant where each benchmark worker uses its own gRPC connection
      - Removes the single-connection bottleneck and allows the server to scale with available CPUs

1. PURE ECDSA signing benchmark over gRPC

      - Adds a Pure ECDSA benchmark that exercises real CPU and allocation pressure

1. Improved plots

      - Updated plots to better compare CPU scaling and GOGC sweeps
      - Catch /direct and /grpc benchmarks

### DIFF
--- a/integration/benchmark/node/bench_test.go
+++ b/integration/benchmark/node/bench_test.go
@@ -29,7 +29,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Benchmark(b *testing.B) {
+func BenchmarkNode(b *testing.B) {
 	benchmarks := []struct {
 		name    string
 		factory viewregistry.Factory
@@ -92,7 +92,7 @@ func Benchmark(b *testing.B) {
 		})
 
 		// run all benchmarks via grpc view API
-		b.Run(fmt.Sprintf("grpc/%s", bm.name), func(b *testing.B) {
+		b.Run(fmt.Sprintf("grpcSingleConnection/%s", bm.name), func(b *testing.B) {
 			n, err := setupNode(b, nodeConfPath, namedFactory{
 				name:    bm.name,
 				factory: bm.factory,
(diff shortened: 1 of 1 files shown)


integration/benchmark/grpc/grpc_multi_connection_bench_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+		for pb.Next() {
+			resp, err := cli.CallViewWithContext(b.Context(), "fid", nil)
+			require.NoError(b, err)
+			require.NotNil(b, resp)
+		}
+	})
+
+	benchmark.ReportTPS(b)
+}
```
mbrandenburger: I think it would be nice to move this benchmark closer to the other case where we use a single connection https://github.com/hyperledger-labs/fabric-smart-client/blob/main/integration/benchmark/grpc/grpc_bench_test.go#L41

integration/benchmark/grpc/grpc_bench_test.go
```diff
@@ -40,7 +40,60 @@ func BenchmarkGRPC(b *testing.B) {
 	benchmark.ReportTPS(b)
 }
 
-func setupServer(tb testing.TB) string {
+func BenchmarkGRPCMultiConnectionCPU(b *testing.B) {
+	srvEndpoint := setupServer(b, "cpu")
+
+	b.RunParallel(func(pb *testing.PB) {
```
mbrandenburger:
```suggestion
    b.ResetTimer()
	b.RunParallel(func(pb *testing.PB) {
```

Should be also reset the timer here as in the other benchmarks?

PR #1127, opened 2026-02-10
fix(web): allow TLS without client authentication

### LINKED ISSUE #1111: [BUG] `ClientCACertFiles` required even when `fsc.web.tls.clientAuthRequired: false`
## Problem description

As reported in the title, there's a potential bug in FSC regarding the `TLS/mTLS` setup. Specifically, if I run an FSC with a section as:

```yaml
fsc:
  web:
    enabled: true
    # HTTP server listener address
    address: 0.0.0.0:10021
    tls:
      enabled: true
      # X.509 certificate used for TLS server
      cert:
        file: /target/fsc/fsc.example.com/peers/issuer.fsc.example.com/tls/server.crt
      # Private key used for TLS server        
      key:
        file: /target/fsc/fsc.example.com/peers/issuer.fsc.example.com/tls/server.key
      # Require client certificates / mutual TLS for inbound connections.
      # Note that clients that are not configured to use a certificate will
      # fail to connect to the node.
      clientAuthRequired: false
```

The FSC node fails with the error:

```
2026-01-13 09:07:10.200 UTC 0060 INFO [fsc.platform.view.services.grpc] NewGRPCServer -> New GRPC Server at [0.0.0.0:10022], TLS [true], RequireClientCert [false]
2026-01-13 09:07:10.202 UTC 0061 INFO [fsc.platform.view.sdk.dig] func2 -> Starting WEB server...
2026-01-13 09:07:10.203 UTC 0062 INFO [fsc.platform.view.sdk.dig] func3 -> Starting operations system...
2026-01-13 09:07:10.203 UTC 0063 INFO [fsc.platform.view.sdk.dig] func1 -> Starting GRPC server...
2026-01-13 09:07:10.203 UTC 0064 FATA [fsc.platform.view.sdk.dig] func2 -> Failed starting WEB server: client TLS CA certificate pool must not be empty
```

My feeling is that it shouldn't happen. If I don't enable `mTLS` through the `clientAuthRequired` flag, I don't need to provide any client CA.

mbrandenburger:
Fixes #1111 by making ClientCACertFiles optional when clientAuthRequired is false. Server now starts successfully with one-way TLS without requiring client CA certificates.

### DIFF
--- a/platform/view/services/web/server/server.go
+++ b/platform/view/services/web/server/server.go
@@ -38,25 +38,16 @@ func (t TLS) Config() (*tls.Config, error) {
 	var tlsConfig *tls.Config
 
 	if !t.Enabled {
+		// no TLS
 		return tlsConfig, nil
 	}
+
+	// TLS setup
 	cert, err := tls.LoadX509KeyPair(t.CertFile, t.KeyFile)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(t.ClientCACertFiles) == 0 {
-		return nil, errors.Errorf("client TLS CA certificate pool must not be empty")
-	}
-
-	caCertPool := x509.NewCertPool()
-	for _, caPath := range t.ClientCACertFiles {
-		caPem, err := os.ReadFile(caPath)
-		if err != nil {
-			return nil, err
-		}
-		caCertPool.AppendCertsFromPEM(caPem)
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM. Thanks Macurs.

PR #1136, opened 2026-02-14
refactor: omit unnecessary reassignment

moveyield:
The new version of Go has been optimized, and variables do not need to be reassigned.

For more info: https://tip.golang.org/wiki/LoopvarExperiment#does-this-mean-i-dont-have-to-write-x--x-in-my-loops-anymore


mbrandenburger:
Thanks @moveyield for creating this PR. Certainly a nice catch and good progress to modernize our codebase. 

I am wondering if there are more spaces in the code where the reassignment is not "so obvious". How did you approach this and find this? Is there more to find?

Issue #1100, opened 2025-12-17
enhancement: notification service handler timeout


mbrandenburger:
The current implementation of the notification service callback/handler invocation may suffer from deadlocks if the handler impl blocks.

Handler invocation: `platform/fabricx/core/finality/nlm.go`

We want some mechanisms to handle this situation with a timeout or other appropriate means.

PR #1141, opened 2026-02-16
enhancement: notification service handler timeout #1100

SaidAltury-ibm:
## Async handler invocation with timeout detection in notification dispatcher

This PR addresses Issue #1100

### Problem

The notification dispatcher in `nlm.go` invoked `OnStatus` handler callbacks synchronously while holding a write lock. This created three risks:

1. **Head-of-line blocking** — a slow or misbehaving handler would block all subsequent notification processing for the entire dispatcher.
2. **No timeout enforcement** — the existing `TODO` comment acknowledged this gap. A rogue handler could block indefinitely with no detection or recovery.

### Solution

Each handler is now invoked in its own goroutine directly inside the dispatcher's section.

Each handler goroutine wraps the `OnStatus` call with a `context.WithTimeout` (default 5s) and a watcher pattern:

- An inner goroutine calls `OnStatus` and closes a `done` channel on completion.
- The outer goroutine `select`s on `done` vs the timeout context.
- If the handler exceeds the timeout, a warning is logged.

This is the same fire-and-forget pattern used by the existing `events` package (`ListenerManager.onBlock`), with the addition of timeout detection and logging.

### Tradeoff

Go has no mechanism to forcefully kill a goroutine. If a handler ignores the context and never returns, the inner goroutine will leak. However, the dispatcher remains fully unblocked and continues processing new notifications. The leaked goroutine is contained and logged.

### Changes

- **`nlm.go`**
  - Added `DefaultHandlerTimeout` constant (5s) and `handlerTimeout` field to `notificationListenerManager`.
  - Replaced synchronous handler invocation with async goroutine-per-handler pattern.
  - Added timeout detection with warning logs for handlers that exceed the deadline.

- **`provider.go`**
  - Set `handlerTimeout: DefaultHandlerTimeout` in the `notificationListenerManager` constructor.

### DIFF
--- a/platform/fabricx/core/finality/nlm.go
+++ b/platform/fabricx/core/finality/nlm.go
@@ -25,6 +25,10 @@ import (
 
 var logger = logging.MustGetLogger()
 
+// DefaultHandlerTimeout is the maximum time allowed for a finality listener
+// handler to complete. If a handler exceeds this timeout, a warning is logged
+// and the handler is abandoned. Note: Handlers that ignore context cancellation
+// will leak goroutines, but this is preferable to blocking the dispatcher.
 const DefaultHandlerTimeout = 5 * time.Second
 
 type notificationListenerManager struct {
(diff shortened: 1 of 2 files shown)


mbrandenburger (approved):
LGTM

PR #1184, opened 2026-03-03
test(deployment): add new integration test for endorser policy

SaidAltury-ibm:
new: added the fabricx/deployment test

fix: namespace polling, interceptor race, and approver finality hang
network.go:
- Fix namespace deployment polling: use tryListInstalledNames() inside gomega.Eventually with a proper closure so the list refreshes on each retry, instead of passing a static slice that never updates
- Add 30s propagation timeout for namespace availability

interceptor.go:
- Cache marshalled bytes atomically to handle concurrent closure
- Fall back to cached bytes when interceptor is closed between IsClosed() check and query executor access, preventing post-failure transaction hangs

iou_test.go:
  - remove the namespace update code

### DIFF
--- a/integration/fabricx/deployment/commands_test.go
+++ b/integration/fabricx/deployment/commands_test.go
@@ -23,15 +23,15 @@ const (
 	defaultViewTimeout = 2 * time.Minute
 )
 
-func CreateDeployment(ii *integration.Infrastructure, identityLabel string, amount uint, approver string) (string, error) {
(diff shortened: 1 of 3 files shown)


mbrandenburger (changes requested):
Thanks Said @SaidAltury-ibm this is nice PR already. I have a few change requests to make this PR even stronger. Please see below. WDYT?

Issue #885, opened 2025-04-09
GRPC stack does not return correct error message


ale-linux:
[This](https://github.ibm.com/decentralized-trust-research/fscx/blob/net_test/integration/fabricx/net/net_test.go) test case tests the behaviour of the net stack when the server has a certificate with a correct and incorrect SANS extension, against both GRPC and webrpc. In both cases the client will refuse to speak to the server, but in the case of webrpc, [the error message is meaningful](https://github.ibm.com/decentralized-trust-research/fscx/blob/net_test/integration/fabricx/net/net_test.go#L87) whereas in the case of GRPC we [just hit the timeout](https://github.ibm.com/decentralized-trust-research/fscx/blob/net_test/integration/fabricx/net/net_test.go#L81)

PR #1189, opened 2026-03-10
return meaningful GRPC error messages

HayimShaul:
return meaningful GRPC error messages.

Solves issue #885


SaidAltury-ibm (approved):
thank you @HayimShaul , LGTM

PR #1229, opened 2026-04-08
refactor: cleanup testify

mbrandenburger:
- use require if possible
- use built-in such as require.Len(t, x, 3) instead of require.Equals(t, len(x), 3)
- replace common/utils/assert with testify where possible


SaidAltury-ibm (approved):
Thank you @mbrandenburger this looks great and aligns the project with clear policy

PR #1231, opened 2026-04-09
testing: consistent use of t.Helper()

mbrandenburger:
This PR introduces consistent use of `t.Helper()` function to improve readability or error messages during tests.


adecaro (approved):
LGTM. As a suggestion, you could enable the linter that checks for the helper and see if it gives the okay.

PR #1232, opened 2026-04-09
endpoint service enhancements

adecaro:
This PR does the following:
- It renames LookupIPv4 to LookupIP and make sure it can properly handle IPv4 and IPv6 formats. Indeed, it failed to handler endpoint of the form `[::]:8080`.
- Improve test coverage above 80%
- Add GoDocs


platform/view/services/endpoint/service_test.go
```diff
@@ -178,3 +169,291 @@ func TestGetIdentity(t *testing.T) {
...
+
+	found := false
+	for _, valid := range validResults {
+		if result == valid {
+			found = true
+			break
+		}
+	}
+
+	assert.True(t, found, "localhost should resolve to a loopback address, got: %s", result)
```
mbrandenburger: could this test be shortened by using `require.Contains()`?

platform/view/services/endpoint/pki_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/endpoint"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefaultPublicKeyIDSynthesizer_PublicKeyID(t *testing.T) {
+	synthesizer := endpoint.DefaultPublicKeyIDSynthesizer{}
+
+	t.Run("RSA public key", func(t *testing.T) {
```
mbrandenburger, line 29: I believe we should add `t.Parallel()` if possible or `//nolint:paralleltest` (if we cannot run the test in parallel). Please check the new tests you add here and existing one that are modified.

platform/view/services/endpoint/service_test.go
```diff
@@ -178,3 +169,291 @@ func TestGetIdentity(t *testing.T) {
...
+		{"IPv6 standard port", "[::1]:8080", "8080"},
+		{"IPv6 low port", "[::1]:80", "80"},
+		{"IPv6 high port", "[::1]:65535", "65535"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := endpoint.LookupIP(tt.endpoint)
+			// Extract port from result
+			_, port, err := splitHostPort(result)
```
mbrandenburger: should we better use `net.SplitHostPort` instead of the new helper function below?

platform/view/services/endpoint/service_test.go
```diff
@@ -178,3 +169,291 @@ func TestGetIdentity(t *testing.T) {
...
+			checkFunc: func(t *testing.T, input, output string) {
+				assert.Equal(t, input, output, "IPv4 any address should be returned as-is")
+			},
+		},
+		{
+			name:     "IPv6 multicast",
+			endpoint: "[ff02::1]:8080",
+			checkFunc: func(t *testing.T, input, output string) {
+				assert.Equal(t, input, output, "IPv6 multicast should be returned as-is")
+			},
```
mbrandenburger, line 391: I would add `t.Helper()` or `//nolint:thelper` to these check functions. Once we "enable" golangci-linter this will come up. Just as a preparation.

mbrandenburger (changes requested):
Thank you @adecaro for this PR! Please check my comments below.

platform/view/services/endpoint/service_test.go
mbrandenburger: just noticing it as I have this PR currently open https://github.com/hyperledger-labs/fabric-smart-client/pull/1231

platform/view/services/endpoint/service_test.go
adecaro: good catch, actually.

platform/view/services/endpoint/service_test.go
adecaro: but it to assert that found is true. It is not related to the string that is just the error message in the case the assertion fails.

platform/view/services/endpoint/pki_test.go
adecaro: okay

platform/view/services/endpoint/service_test.go
adecaro: okay

adecaro:
@mbrandenburger , thanks for the review. Addressed all your comments. 
Please, have another look.

platform/view/services/endpoint/service_test.go
mbrandenburger: ok, I understand. I thought you could simply do 
```go
require.Containsf(t, validResults, result, "localhost should resolve to a loopback address, got: %s", result)
``` 
and get rid of for loop.

adecaro: sorry for the misunderstanding then. I'll check that solution 😅

adecaro:
@mbrandenburger , updated the test as suggested.

mbrandenburger (approved):
Thank you @adecaro

PR #1256, opened 2026-04-20
fix(fabricx): restore endpoint parser for id=<group>,<type>,<host:port>

tinystone007:
## Problem

The fabric-x official ansible collection (`LF-Decentralized-Trust-labs/fabric-x-ansible-collection`) and the fabric-x-tool `config-builder` emit orderer endpoints in channel config as:

- `id=<group>,broadcast,<host>:<port>` for routers
- `id=<group>,deliver,<host>:<port>` for assemblers

See [ansible collection configtx template](https://github.com/LF-Decentralized-Trust-labs/fabric-x-ansible-collection/blob/main/roles/configtxgen/templates/configtx.yaml.j2).

FSC v0.8.2 parsed this format in `platform/fabricx/core/membership/endpoint.go` and filtered to broadcast-only endpoints in `Service.OrdererConfig` (https://github.com/hyperledger-labs/fabric-smart-client/blob/v0.8.2/platform/fabricx/core/membership/membership.go#L170-L207). The parser was removed during the v0.9/v0.10 membership refactor, which now treats every endpoint (including deliver-typed assembler ports) as broadcast.

## Impact

With the official configtx template, FSC clients attempt to `Broadcast` to assemblers (which only serve `Deliver`) and get:

```
failed to get status after broadcast to [host:7x53]: rpc error: code = Unknown desc = should not be used
```

Any FSC-based token-sdk / fabricx application deployed on top of the ansible-provisioned fabric-x infrastructure hits this regression on v0.9+.

## Fix

This PR ports the v0.8.2 `endpoint.go` and reintroduces the filter in `OrdererConfig`:

1. Parses `id=<group>,<type>,<host:port>` when present
2. Skips `deliver`-typed endpoints (clients only broadcast here; deliver is sourced from the local sidecar / committer)
3. Falls back to treating plain `<host:port>` as broadcast for backward compatibility with non-fabric-x callers

## Sibling PR

Related upstream regression patch: #1249 (arma -> BFT broadcaster alias, also from the same v0.9/v0.10 membership refactor).

## Testing

Verified end-to-end on a fabric-x v0.1.9 / orderer v0.0.24 deployment using the ansible collection configtx format:
- TMS deploy finality succeeds
- IssueToken / Transfer / Redeem full token lifecycle passes
- No more "should not be used" broadcast errors on assembler ports

### DIFF
--- a/platform/fabricx/core/membership/endpoint.go
+++ b/platform/fabricx/core/membership/endpoint.go
@@ -1,0 +1,51 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package membership
+
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+)
+
+const (
+	OrdererBroadcastType = "broadcast"
+	OrdererDeliverType   = "deliver"
+)
+
+var (
+	endpointRegex            = regexp.MustCompile(`id=(\d+),(` + OrdererBroadcastType + `|` + OrdererDeliverType + `),(.*)`)
+	ErrInvalidEndpointFormat = errors.New("invalid endpoint format")
+)
+
+type endpoint struct {
+	ID       int
+	Type     string
+	Endpoint string
+}
+
+// parseEndpoint parses the fabric-x channel-config endpoint format
+// `id=<group>,(broadcast|deliver),<host>:<port>` as emitted by the official
+// fabric-x ansible collection and fabric-x-tool config-builder. A plain
+// `<host>:<port>` value is accepted and treated as broadcast with ID=0 for
+// backward compatibility.
+func parseEndpoint(str string) (*endpoint, error) {
+	match := endpointRegex.FindStringSubmatch(str)
+	if len(match) == 4 {
+		id, err := strconv.Atoi(match[1])
+		if err != nil {
+			return nil, errors.Wrap(err, "invalid endpoint id")
+		}
+		return &endpoint{ID: id, Type: match[2], Endpoint: match[3]}, nil
+	}
+	// Fallback: plain host:port. Treat as broadcast.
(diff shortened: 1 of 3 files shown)


platform/fabricx/core/membership/endpoint.go
```diff
@@ -0,0 +1,51 @@
...
+	Type     string
+	Endpoint string
+}
+
+// parseEndpoint parses the fabric-x channel-config endpoint format
+// `id=<group>,(broadcast|deliver),<host>:<port>` as emitted by the official
+// fabric-x ansible collection and fabric-x-tool config-builder. A plain
+// `<host>:<port>` value is accepted and treated as broadcast with ID=0 for
+// backward compatibility.
+func parseEndpoint(str string) (*endpoint, error) {
```
adecaro, line 37: please, add unit-tests to make sure this function behave as expected especially in the corner cases.

adecaro (changes requested):
Hi @tinystone007 , great catch and great effort.

Please, add unit-tests and we are good with this PR.

platform/fabricx/core/membership/endpoint.go
tinystone007: Added unit tests in `platform/fabricx/core/membership/endpoint_test.go` (commit eff7d970) covering:

- `id=<n>,broadcast,host:port` and `id=<n>,deliver,host:port` happy paths
- multi-digit ID
- endpoint with embedded colons (IPv6-style `[::1]:7050`) round-tripping through the `.*` capture
- plain `host:port` falling through to the backward-compat fallback (`ID=0`, `Type=broadcast`)
- empty string → `ErrInvalidEndpointFormat` (asserted via `errors.Is`)
- ID overflowing `int` → wrapped `strconv` error

Let me know if there are other corner cases you'd like covered.

mbrandenburger (approved):
Thank you @tinystone007

adecaro (approved):
LGTM. Thanks a lot 🙏

PR #1296, opened 2026-04-24
fix: consistent use of TLS12/TLS13

### LINKED ISSUE #1297: Inconsistent TLS configuration
The current code suffers from inconsistent TLS configurations which may lead to handshake failures.

To follow go's [current defaults](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L791) we should use the following TLS configuration.

```go
MinVersion: tls.VersionTLS12,
MaxVersion: tls.VersionTLS13,
```

mbrandenburger:
This PR fixes inconsistent use of min/max tls version in TLS configurations which may cause handshake failures.

As aligned with go [current defaults](https://github.com/golang/go/blob/master/src/crypto/tls/common.go#L791) we use the following TLS configuration.

```go
MinVersion: tls.VersionTLS12,
MaxVersion: tls.VersionTLS13,
```

Closes #1297

### DIFF
--- a/platform/fabricx/core/committer/grpc/grpc.go
+++ b/platform/fabricx/core/committer/grpc/grpc.go
@@ -118,7 +118,8 @@ func TransportCredentials(tlsConfig *config.TLSConfig) (credentials.TransportCre
 	}
 
 	t := &tls.Config{
-		MinVersion: tls.VersionTLS13,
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

mbrandenburger:
Blocked until #1298 is completed.

PR #1235, opened 2026-04-10
fix(storage): migrate SimpleKeyDataStore and KeyValueStore to squirrel

sid200727:
Part of #1159

Migrates `SimpleKeyDataStore` and `KeyValueStore` from the deprecated `CondInterpreter` query builder to squirrel's `PlaceholderFormat`, following the same pattern established for `SignerInfoStore` (#1199) and `BindingStore`.

## Changes
- `common/simplekeydata.go` : replace `CondInterpreter` with `sq.PlaceholderFormat`
- `common/unversioned.go` : replace `CondInterpreter` with `sq.PlaceholderFormat`
- `sqlite/unversioned.go` : pass `sq.Question`
- `postgres/keyvalue.go` : pass `sq.Dollar`
- `fabric/sql/common` : update `MetadataStore`, `EndorseTxStore`, `EnvelopeStore` constructors
- Update all affected test files to match squirrel SQL format

## Test results
All locally runnable tests pass:
- `platform/fabric/services/db/driver/sql/common` 
- `platform/view/services/storage/driver/sql/common` 
- `platform/view/services/storage/driver/sql/sqlite` 
- All query builder tests 

Postgres failures are pre-existing (require Docker daemon).

PR #1301, opened 2026-04-25
feat(vault): migrate VaultStore SELECT queries to squirrel (#1272)

### LINKED ISSUE #1272: Migrate VaultStore to squirrel
> Following the merges of PR #1199, #1235, and #1258, the remaining store is VaultStore in platform/fabric/services/db/driver/sql/common/vault.go. 
> Unlike the previous stores, it uses both CondInterpreter and PagInterpreter. @mbrandenburger, could you advise on the best approach for migrating PagInterpreter to squirrel before I start? 
> Should it follow the same pattern or does it need a different strategy? 

 _Originally posted by @sid200727 in [#1159](https://github.com/hyperledger-labs/fabric-smart-client/issues/1159#issuecomment-4289351447)_

sid200727:
Closes #1272

Migrates VaultStore SELECT queries from the legacy query builder to squirrel, following the same pattern as the SignerInfoStore migration (PR #1235).

## Changes
-> `platform/view/services/storage/driver/sql/query/pagination/squirrel.go` - new `ApplyToSquirrel` helper that translates `driver.Pagination` into squirrel `.Limit()/.Offset()` calls
-> `platform/fabric/services/db/driver/sql/common/vault.go` - `NewVaultStore` now takes `sq.PlaceholderFormat` instead of `PagInterpreter`; all SELECT queries migrated to squirrel
-> `platform/fabric/services/db/driver/sql/sqlite/vault.go` - passes `sq.Question`
-> `platform/fabric/services/db/driver/sql/postgres/vault.go` - passes `sq.Dollar`

## Design (per @mbrandenburger)
Built a squirrel-based `ApplyToSquirrel` helper in the pagination package rather than bypassing `PagInterpreter` entirely, preserving existing pagination semantics while removing the dependency on the legacy query builder.

All existing vault tests pass.

### DIFF
--- a/platform/fabric/services/db/driver/sql/common/vault.go
+++ b/platform/fabric/services/db/driver/sql/common/vault.go
@@ -215,6 +215,7 @@ func (db *VaultStore) convertStateRows(writes driver.Writes, metaWrites driver.M
 				return nil, errors.Wrapf(err, "failed to marshal metadata for [%s:%s]", ns, pkey)
 			}
 			if len(val.Raw) == 0 {
+				// Deleting value
 				logger.Debugf("setting version of [%s] to nil", pkey)
 				val.Version = nil
 			}
@@ -456,7 +457,7 @@ func (db *vaultReader) GetStateMetadata(ctx context.Context, namespace driver.Na
 func (db *vaultReader) GetLast(ctx context.Context) (*driver.TxStatus, error) {
 	// Select the row with the highest pos that is not Busy
 	it, err := db.queryStatus(ctx,
-		sq.Expr("pos=(SELECT max(pos) FROM "+db.tables.StatusTable+" WHERE code!=?)", driver.Busy),
+		sq.Expr(fmt.Sprintf("pos=(SELECT max(pos) FROM %s WHERE code!=?)", db.tables.StatusTable), driver.Busy),
 		pagination.None())
 	if err != nil {
 		return nil, err
--- a/platform/fabric/services/db/driver/sql/common/vault.go
+++ b/platform/fabric/services/db/driver/sql/common/vault.go
@@ -160,6 +160,10 @@ func (db *VaultStore) SetStatuses(ctx context.Context, code driver.TxStatusCode,
 	return nil
 }
 
+// SetStatusesBusy builds an INSERT ... ON CONFLICT DO UPDATE statement that marks all
+// transactions in txIDs as Busy in the status table. It returns the SQL query string
+// and the corresponding argument slice ready for execution, or an error if the query
+// cannot be built.
 func (db *VaultStore) SetStatusesBusy(txIDs []driver.TxID) (string, []any, error) {
 	ib := db.sb.Insert(db.tables.StatusTable).
 		Columns("tx_id", "code")
--- a/platform/fabric/services/db/driver/sql/common/vault.go
+++ b/platform/fabric/services/db/driver/sql/common/vault.go
@@ -145,6 +145,7 @@ func (db *VaultStore) SetStatuses(ctx context.Context, code driver.TxStatusCode,
 	for _, txID := range txIDs {
 		ib = ib.Values(txID, code, message)
 	}
+
 	query, params, err := ib.
 		Suffix("ON CONFLICT (tx_id) DO UPDATE SET code = EXCLUDED.code, message = EXCLUDED.message").
 		ToSql()
@@ -170,6 +171,7 @@ func (db *VaultStore) SetStatusesBusy(txIDs []driver.TxID) (string, []any, error
 	for _, txID := range txIDs {
 		ib = ib.Values(txID, driver.Busy)
 	}
+
 	return ib.
 		Suffix("ON CONFLICT (tx_id) DO UPDATE SET code = EXCLUDED.code").
 		ToSql()
@@ -242,6 +244,7 @@ func (db *VaultStore) convertStateRows(writes driver.Writes, metaWrites driver.M
 		}
 		for pkey, metaVal := range metaWrite {
 			if _, ok = write[pkey]; ok {
+				// MetaWrites already written in the previous section
 				continue
 			}
 			metadata, err := marshallMetadata(metaVal.Metadata)
@@ -257,6 +260,7 @@ func (db *VaultStore) convertStateRows(writes driver.Writes, metaWrites driver.M
 				return nil, err
 			}
 			states = append(states, common4.Tuple{ns, pkey, []byte{}, metaVal.Version, metadata})
+
 		}
 	}
 	return states, nil
@@ -455,6 +459,7 @@ func (db *vaultReader) GetStateMetadata(ctx context.Context, namespace driver.Na
 	if err != nil {
 		return meta, nil, fmt.Errorf("error decoding metadata: %w", err)
 	}
+
 	return meta, kversion, err
 }
 
@@ -488,6 +493,7 @@ func (db *vaultReader) GetAllTxStatuses(ctx context.Context, p driver.Pagination
 	if p == nil {
 		return nil, fmt.Errorf("invalid input pagination: %+v", p)
 	}
+
 	txStatusIterator, err := db.queryStatus(ctx, nil, p)
 	if err != nil {
 		return nil, err
--- a/platform/fabric/services/db/driver/sql/common/vault.go
+++ b/platform/fabric/services/db/driver/sql/common/vault.go
@@ -111,7 +111,7 @@ func (db *VaultStore) newTxLockVaultReader(ctx context.Context, isolationLevel d
 	}
 	return &vaultReader{
 		readDB:    tx,
-		sb:        db.vaultReader.sb,
+		sb:        db.sb,
 		sanitizer: db.sanitizer,
 		tables:    db.tables,
 	}, tx.Commit, nil
@@ -131,7 +131,7 @@ func (db *VaultStore) newGlobalLockVaultReader() (*vaultReader, releaseFunc, err
 	}
 	return &vaultReader{
 		readDB:    db.readDB,
-		sb:        db.vaultReader.sb,
+		sb:        db.sb,
 		sanitizer: db.sanitizer,
 		tables:    db.tables,
 	}, release, nil
--- a/platform/fabric/services/db/driver/sql/common/vault.go
+++ b/platform/fabric/services/db/driver/sql/common/vault.go
@@ -15,6 +15,8 @@ import (
 	"fmt"
 	"sync"
 
+	sq "github.com/Masterminds/squirrel"
+
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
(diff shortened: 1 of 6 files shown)


sid200727:
Hi @mbrandenburger ! PR #1301 is up for the VaultStore squirrel migration - lint and tests are clean. Would love your feedback whenever you get a chance. 🙏

mbrandenburger (changes requested):
Thank you @sid200727 for working on this. This PR makes good progress. In particular, the squirrel-based PagInterpreter alternative is much easier than I thought. 

Why don't we update the "write-path" in this PR as well so we could get rid of `ci common4.CondInterpreter` entirely here? Is there any blocker?

Could you please add unit test for the new helper functions (`ApplyToSquirrel`, etc ...).

sid200727:
Thank you @mbrandenburger! No blocker on migrating the write-path, I'll look at the write operations in vault.go and migrate those to squirrel as well to remove ci entirely. Will also add unit tests for ApplyToSquirrel.

platform/view/services/storage/driver/sql/query/pagination/squirrel_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func TestApplyToSquirrel_None(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+
+	q, args, err := pagination.ApplyToSquirrel(pagination.None(), baseQuery()).ToSql()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(q).To(Equal("SELECT tx_id, code FROM status"))
+	Expect(args).To(BeEmpty())
+}
```
mbrandenburger: Do you think it would be possible to rewrite the tests using `testify/require` instead of `onsi/gomega`. This would allow us to run the tests in parallel easily.

platform/view/services/storage/driver/sql/query/pagination/squirrel.go
```diff
@@ -0,0 +1,58 @@
...
+		return sb
+
+	case *keyset[string, any]:
+		return applyKeysetSquirrel(pg, sb)
+
+	case *keyset[int, any]:
+		return applyKeysetSquirrel(pg, sb)
+
+	case *empty:
+		return sb.Limit(0)
```
mbrandenburger, line 42: Do we cover these in the unit tests?

platform/fabric/services/db/driver/sql/common/vault.go
```diff
@@ -376,29 +367,44 @@ func (db *vaultReader) GetState(ctx context.Context, namespace driver.Namespace,
...
+func betweenStrings(col, start, end string) sq.Sqlizer {
+	var parts sq.And
+	if start != "" {
+		parts = append(parts, sq.GtOrEq{col: start})
+	}
+	if end != "" {
+		parts = append(parts, sq.Lt{col: end})
+	}
+	return parts
+}
```
mbrandenburger, line 401: We should have a unit test for this one as well. WDYT?

platform/fabric/services/db/driver/sql/common/vault.go
```diff
@@ -221,7 +215,6 @@ func (db *VaultStore) convertStateRows(writes driver.Writes, metaWrites driver.M
 				return nil, errors.Wrapf(err, "failed to marshal metadata for [%s:%s]", ns, pkey)
 			}
 			if len(val.Raw) == 0 {
-				// Deleting value
```
mbrandenburger, line 224: Can you please keep these comments as they provide additional context for readers - even if the sometimes seem trivial.

platform/fabric/services/db/driver/sql/common/vault.go
```diff
@@ -441,35 +450,22 @@ func (db *vaultReader) GetStateMetadata(ctx context.Context, namespace driver.Na
...
 	}
-
 	return meta, kversion, err
 }
 
 func (db *vaultReader) GetLast(ctx context.Context) (*driver.TxStatus, error) {
-	it, err := db.queryStatus(ctx, IsLast(common4.TableName(db.tables.StatusTable)), pagination.None())
+	// Select the row with the highest pos that is not Busy
+	it, err := db.queryStatus(ctx,
+		sq.Expr("pos=(SELECT max(pos) FROM "+db.tables.StatusTable+" WHERE code!=?)", driver.Busy),
```
mbrandenburger:
```suggestion
 sq.Expr(
      fmt.Sprintf("pos=(SELECT max(pos) FROM %s WHERE code!=?)", db.tables.StatusTable),
      driver.Busy,
  )
```

platform/fabric/services/db/driver/sql/sqlite/vault.go
```diff
@@ -56,39 +51,58 @@ func (db *VaultStore) Store(ctx context.Context, txIDs []driver.TxID, writes dri
...
-		WriteString(`
-		BEGIN;
-`)
+	db.GlobalLock.RLock()
+	defer db.GlobalLock.RUnlock()
+
+	tx, err := db.writeDB.Begin()
+	if err != nil {
+		return errors.Wrapf(err, "failed to initiate db transaction")
+	}
```
mbrandenburger, line 59: Could we add a `defer func() { _ = tx.Rollback() }` instead of manually calling rollback on the error cases below?

platform/fabric/services/db/driver/sql/common/vault.go
```diff
@@ -162,41 +160,37 @@ func (db *VaultStore) SetStatuses(ctx context.Context, code driver.TxStatusCode,
...
 
-func (db *VaultStore) SetStatusesBusy(txIDs []driver.TxID, sb common4.Builder) {
-	rows := make([]common4.Tuple, len(txIDs))
-	for i, txID := range txIDs {
-		rows[i] = common4.Tuple{txID, driver.Busy}
+func (db *VaultStore) SetStatusesBusy(txIDs []driver.TxID) (string, []any, error) {
+	ib := db.sb.Insert(db.tables.StatusTable).
+		Columns("tx_id", "code")
+	for _, txID := range txIDs {
+		ib = ib.Values(txID, driver.Busy)
```
mbrandenburger, line 172: If I understand this change correctly, we are previously where executing a query whereas with the change we just create the query and execute it later using execOrRollback. Is this correct?

mbrandenburger (changes requested):
Thank you @sid200727 for working on this.

platform/view/services/storage/driver/sql/query/pagination/squirrel_test.go
sid200727: Sure, will switch to testify/require and makes it consistent with the rest of the codebase and enables parallel runs.

platform/view/services/storage/driver/sql/query/pagination/squirrel.go
sid200727: Good catch, will add tests for those branches too

platform/fabric/services/db/driver/sql/common/vault.go
sid200727: Agreed, will add a unit test for betweenStrings as well.

platform/fabric/services/db/driver/sql/common/vault.go
sid200727: Sorry about that, will restore it as those context comments definitely help readers understand the logic.

platform/fabric/services/db/driver/sql/common/vault.go
sid200727: Makes sense, accepting the suggestion!

platform/fabric/services/db/driver/sql/common/vault.go
sid200727: Yes, that's correct @mbrandenburger, the old code built and executed the query inline, while the new version builds the squirrel insert builder and passes it to execOrRollback which handles execution and rollback in one place.

platform/fabric/services/db/driver/sql/sqlite/vault.go
sid200727: Good point, using defer is much cleaner than scattering manual rollback calls. Will apply that. 🙏

platform/fabric/services/db/driver/sql/common/vault.go
mbrandenburger: Actually I was wrong - the old code modified the `sb common4.Builder` parameter. That is, the new API makes much more sense. Can you update the godocs and describe what `SetStatusesBusy` is doing and what the inputs/outputs are.

platform/fabric/services/db/driver/sql/sqlite/vault.go
mbrandenburger: Should we also use the same concept in `postgres/vault.go` and get rid of `execOrRollback`? That way we would have a consistent way to rollback in the sqlite and postgres impl. WDYT?

platform/fabric/services/db/driver/sql/common/vault.go
sid200727: Sure ! that makes more sense! Will add godoc for SetStatusesBusy describing the inputs/outputs. 👍

platform/fabric/services/db/driver/sql/sqlite/vault.go
sid200727: That makes sense, using defer func() { _ = tx.Rollback() } consistently in both sqlite and postgres vault implementations would be cleaner and removes the need for execOrRollback entirely. 
Happy to apply that pattern to postgres/vault.go as well 😃.

sid200727:
Added godoc comments to all functions with changed or new signatures, NewVaultStore and newVaultStore in both postgres and sqlite drivers, queryState, queryStatus, betweenStrings in common vault, and ApplyToSquirrel in the pagination package.

platform/view/services/storage/driver/sql/query/pagination/squirrel.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"fmt"
+
+	sq "github.com/Masterminds/squirrel"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
+)
+
+// ApplyToSquirrel translates a driver.Pagination into squirrel limit/offset
+// clauses applied to sb and returns the modified SelectBuilder.
```
mbrandenburger:
```suggestion
```

platform/view/services/storage/driver/sql/query/pagination/squirrel_test.go
```diff
@@ -0,0 +1,103 @@
...
+func baseQuery() sq.SelectBuilder {
+	return sq.Select("tx_id", "code").From("status")
+}
+
+func TestApplyToSquirrel_Nil(t *testing.T) { //nolint:paralleltest
+	q, args, err := pagination.ApplyToSquirrel(nil, baseQuery()).ToSql()
+	require.NoError(t, err)
+	require.Equal(t, "SELECT tx_id, code FROM status", q)
+	require.Empty(t, args)
+}
```
mbrandenburger: I think we could run these tests now with `t.Parralel()`. WDYT?

mbrandenburger (changes requested):
Thank you @sid200727. A few little things and then this is ready to merge.

platform/view/services/storage/driver/sql/query/pagination/squirrel_test.go
sid200727: Agreed, will add t.Parallel() to all tests.

sid200727: Sure!

mbrandenburger:
@arner could I please lend once more your eagle eyes on this one please :)

platform/fabric/services/db/driver/sql/postgres/vault.go
```diff
@@ -57,54 +55,40 @@ func (db *VaultStore) Store(ctx context.Context, txIDs []driver.TxID, writes dri
...
 	if len(txIDs) > 0 {
-		sb := common5.NewBuilder()
-		db.SetStatusesBusy(txIDs, sb)
-		query, args := sb.Build()
-		if err := execOrRollback(ctx, tx, query, args); err != nil {
+		query, params, err := db.SetStatusesBusy(txIDs)
+		if err != nil {
+			return errors.Wrapf(err, "failed building busy query")
+		}
+		logger.Debug(query, len(params), params)
```
arner, line 65: I don't know how the vault works exactly, but have a hunch that this could become quite big if we set a list of tx ids as busy, wdyt @mbrandenburger ?

arner: (same for other places in this file)

arner (commented):
@mbrandenburger I'm by no means an expert in this, did my best to spot issues but haven't found any :)

platform/fabric/services/db/driver/sql/postgres/vault.go
mbrandenburger: I agree that these logs could be pretty big - the previous code were also logging all queries - thus this particular PR does not change the logging situation. Thus, I would keep the detailed logging.

mbrandenburger (approved):

Issue #798, opened 2025-03-21
Code maintainability


alexandrosfilios:
Architectural changes, Cleanups, Refactorings towards a more maintainable codebase

Issue #209, opened 2022-01-13
Provide a configuration validation tool


mffrench:
We do a lot of big fingers some time and unfortunately many time from these big fingers on configuration we have panic on fsc node which doesn't help to find source of errors.

In that perspective, a configuration validator would help a lot.

malsomesh9:
I opened a first-slice PR for this here: https://github.com/hyperledger-labs/fabric-smart-client/pull/1416

What it adds:
- a new `fsccli validate config` command for preflight checking of `core.yaml`
- validation for node loading, fabric network config, gRPC/web server config, and tracing config
- focused tests around the new command

This is intended to make common configuration mistakes easier to catch before starting a node, and it also partially helps with the error-reporting concerns in #410.

malsomesh9:
I already opened a PR for this here: #1416
https://github.com/hyperledger-labs/fabric-smart-client/pull/1416

I also added a summary comment above describing what it implements. I’ll continue from that PR rather than opening another one.

malsomesh9:
@mbrandenburger Plz have a look

PR #1327, opened 2026-04-26
node: add config validation command

malsomesh9:
## Summary
- add a `peer validate-config` command that loads and validates the FSC node configuration before startup
- fail with a clear error when the config is missing `fsc.id`
- return the concrete config file path via `ConfigFileUsed()` so the validator can report exactly what it checked

## Why
Today, configuration mistakes are often only discovered later during node startup. This adds a lightweight validation tool that helps contributors and operators catch obvious config issues earlier and see which `core.yaml` was actually loaded.

## Testing
- `go test ./node/validate`
- `go test ./platform/view/services/config`
- `go test ./node/...`

Refs #209
Refs #798

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/hyperledger-labs/fabric-smart-client/node/start"
+	"github.com/hyperledger-labs/fabric-smart-client/node/validate"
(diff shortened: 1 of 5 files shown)


malsomesh9:
@SaidAltury-ibm @mbrandenburger Plz take a look at this.

PR #1329, opened 2026-04-26
fix(lazy): add read lock to lazyProvider.Length

### LINKED ISSUE #1324: lazyProvider.Length() is not concurrency-safe: missing read lock
## Summary

`lazyProvider.Length()` in `platform/common/utils/lazy/provider.go` reads from `cache` without holding `cacheLock`, making it unsafe for concurrent use. All other methods that access `cache` — `Get`, `Peek`, `Update`, `Delete` — correctly acquire the lock before accessing the map.

## Current behavior

```go
func (v *lazyProvider[I, K, V]) Length() int {
    return len(v.cache)  // no lock held
}
```

Calling `Length()` concurrently with any write operation (`Get`, `Update`, `Delete`) constitutes a data race on the underlying map. The Go runtime will detect this under `-race` and panic in production.

## Expected behavior

```go
func (v *lazyProvider[I, K, V]) Length() int {
    v.cacheLock.RLock()
    defer v.cacheLock.RUnlock()
    return len(v.cache)
}
```

## Steps to reproduce

Run any test that calls `Length()` concurrently with `Get()` under the race detector:

```
go test -race ./platform/common/utils/lazy/...
```

## Additional context

The existing test `TestProviderMethods` calls `Length()` but does not exercise it concurrently alongside writes, so the race is not currently caught. A concurrent test should be added alongside the fix.

mvanhorn:
Resolves #1324.

\`lazyProvider.Length()\` in \`platform/common/utils/lazy/provider.go\` read \`len(v.cache)\` without holding \`cacheLock\`. Every other cache reader (\`Get\`, \`Peek\`, \`Update\`, \`Delete\`) acquires the lock first, so calling \`Length\` concurrently with any of those is a data race that the Go runtime detects under \`-race\` and panics on in production.

## Fix

Adds the symmetric read-lock acquisition the issue calls out:

\`\`\`go
func (v *lazyProvider[I, K, V]) Length() int {
    v.cacheLock.RLock()
    defer v.cacheLock.RUnlock()
    return len(v.cache)
}
\`\`\`

## Test

Added \`TestLengthConcurrentWithWrites\` in \`provider_test.go\` that drives 200 \`Update\` calls, 200 \`Delete\` calls, and 200 \`Length\` calls from three goroutines in parallel. Without the RLock fix this would trip the race detector or panic with "concurrent map iteration and map write".

\`go test -race ./platform/common/utils/lazy/...\` passes locally with the fix.

### DIFF
--- a/platform/common/utils/lazy/provider.go
+++ b/platform/common/utils/lazy/provider.go
@@ -114,5 +114,7 @@ func (v *lazyProvider[I, K, V]) Delete(input I) (V, bool) {
 }
 
 func (v *lazyProvider[I, K, V]) Length() int {
+	v.cacheLock.RLock()
+	defer v.cacheLock.RUnlock()
 	return len(v.cache)
 }
--- a/platform/common/utils/lazy/provider_test.go
+++ b/platform/common/utils/lazy/provider_test.go
@@ -170,6 +170,42 @@ func TestParallel(t *testing.T) {
 	require.Len(t, values, 1, "we always got one value back (the one we first set)")
 }
 
+// TestLengthConcurrentWithWrites exercises Length alongside Update / Delete
+// to expose the missing read lock. Without the RLock fix on Length, this
+// test panics under `go test -race` with "concurrent map iteration and map
+// write" or trips the race detector.
+func TestLengthConcurrentWithWrites(t *testing.T) {
+	t.Parallel()
+	const iterations = 200
+	cache := newTestCache()
+
+	var wg sync.WaitGroup
+	wg.Add(3)
+
+	go func() {
+		defer wg.Done()
+		for i := 0; i < iterations; i++ {
+			_, _, _ = cache.Update(entry{key: fmt.Sprintf("k%d", i), value: fmt.Sprintf("v%d", i)})
+		}
+	}()
+
+	go func() {
+		defer wg.Done()
+		for i := 0; i < iterations; i++ {
+			_, _ = cache.Delete(entry{key: fmt.Sprintf("k%d", i)})
+		}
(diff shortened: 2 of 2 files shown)


SaidAltury-ibm:
@mvanhorn all commits should be Signed-off

mvanhorn:
Done — amended cfd1a6e with `Signed-off-by`, force-pushed as ff48072. DCO check now passing. Verified: `go test -race` on TestLengthConcurrentWithWrites still passes.

SaidAltury-ibm (approved):
Thank you @mvanhorn, LGTM!

SaidAltury-ibm:
@mvanhorn still there's an issue with the commit after rebase.

mvanhorn:
@SaidAltury-ibm thanks for the rebase. The DCO failure was an author/committer email mismatch with the sign-off line. Amended (d0a8001) so author, committer, and Signed-off-by all use mvanhorn@gmail.com. DCO check is green now.

SaidAltury-ibm (approved):

SaidAltury-ibm (approved):

mvanhorn:
Appreciate the quick turnaround @SaidAltury-ibm. The lazyProvider race was subtle (only fires under -race plus concurrent Get/Length) so good to have the lock symmetry in place.

SaidAltury-ibm:
Thanks @mvanhorn ! Agreed it's the kind of thing that only surfaces under the right conditions, which is exactly why having lock symmetry as a baseline rule makes it easier to spot. thank you for adding a concurrent test for `Length()` in the same PR that's useful.

PR #1333, opened 2026-04-26
fabric: apply shared TLS root cert defaults (#1112)

malsomesh9:
## What changed
- apply `fabric.<network>.tls.rootCert.file` as the default outbound TLS root CA for configured Fabric peers and orderers
- preserve per-endpoint `tlsRootCertFile` precedence when a peer or orderer provides its own root CA
- keep compatibility with the legacy flat `fabric.<network>.tls.rootCertFile` key while the TLS model is being aligned
- update the Fabric config model/docs/examples to describe the `tls` section as client-side connection defaults

## Why
The `#1112` TLS refactor discussion calls out that the Fabric network TLS section should describe client-side connection settings clearly. Before this change, the shared TLS root certificate was documented but never actually applied to peer/orderer client connections, so operators had to repeat the same `tlsRootCertFile` value on every endpoint.

## Impact
Configured Fabric peers and orderers now inherit the shared outbound root CA from `fabric.<network>.tls.rootCert.file` unless they override it explicitly. Existing per-endpoint configs keep the same behavior, and older `rootCertFile` configs continue to work.

## Validation

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -340,12 +340,16 @@ fabric:
       enabled:  true
       # Specifies whether the fabric network requires mutualTLS
       clientAuthRequired: false
-      # The client tls certificate if mutualTLS is required
+      # The client TLS certificate if mutualTLS is required
       clientCert:
         file: /path/to/client.crt
-      # The client tls key if mutualTLS is required
+      # The client TLS key if mutualTLS is required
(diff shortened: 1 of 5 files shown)


mbrandenburger:
Superseded by #1728

PR #1334, opened 2026-04-26
fabric: align shared TLS server name override (#1112)

malsomesh9:
## What changed
- accept `fabric.<network>.tls.serverNameOverride` as the shared Fabric client-side TLS hostname override
- keep the legacy `fabric.<network>.tls.serverhostoverride` key as a fallback for compatibility
- document the shared override in the Fabric configuration guide and test fixture

## Why
Issue `#1112` is aligning FSC TLS configuration around clearer, more consistent names. The shared Fabric client-side override still used the older `serverhostoverride` spelling even though endpoint-level configs and newer docs use `serverNameOverride`.

## Impact
Operators can now configure the shared outbound hostname override with the same `serverNameOverride` spelling used elsewhere in FSC, without breaking existing configs that still use the legacy key.

## Validation

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -340,6 +340,9 @@ fabric:
       enabled:  true
       # Specifies whether the fabric network requires mutualTLS
       clientAuthRequired: false
+      # Optional default server name override for outbound TLS hostname verification.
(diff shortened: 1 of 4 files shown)


mbrandenburger:
Superseded by #1728

PR #1336, opened 2026-04-26
fabric: apply ordering TLS defaults to configured orderers (#1112)

malsomesh9:
## What changed
- apply `ordering.tlsEnabled` to configured Fabric `orderers:` entries instead of only discovered channel orderers
- apply `ordering.tlsClientAuthRequired` to configured `orderers:` entries as the default client-auth setting
- preserve explicit per-orderer `tlsEnabled`, `tlsDisabled`, and `tlsClientSideAuth` opt-ins

## Why
Issue `#1112` is aligning FSC TLS behavior with the documented configuration model. Before this change, the `ordering` TLS overrides were only used when deriving orderer connections from channel config, while statically configured `fabric.<network>.orderers` silently ignored them.

## Impact
Configured orderers now follow the same TLS defaulting rules documented for the ordering service. Users can disable or enable TLS for ordering separately from peer traffic, and can enable ordering mTLS centrally without repeating the setting on every configured orderer.

## Validation

### DIFF
--- a/platform/fabric/core/generic/config/service.go
+++ b/platform/fabric/core/generic/config/service.go
@@ -79,13 +79,23 @@ func NewService(configService Configuration, name string, defaultConfig bool) (*
 	}
 
 	tlsEnabled := configService.GetBool(fmt.Sprintf("fabric.%stls.enabled", prefix))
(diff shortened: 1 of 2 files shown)


mbrandenburger:
Superseded by #1728

PR #1337, opened 2026-04-26
comm: add transport round-trip benchmarks (#797)

### LINKED ISSUE #797: Comm layer performance
Currently the main implementation used for communication is based on web sockets. Previously we used libp2p but we transitioned in order to support custom routing (necessary for node replication), and multi-node tracing (transfer context between nodes is not supported for libp2p out of the box).

However, the most important criterion would be the actual performance of the communication layer. The applications currently support both modes (all integration tests test against both ways), but if we want to decommission libp2p, we will need to have some performance tests to test that web sockets don't have a negative impact on performance.

malsomesh9:
## What changed
- add websocket P2P transport microbenchmarks for new-session and reused-session round trips
- add libp2p P2P transport microbenchmarks for the same session patterns
- widen existing transport test setup helpers from `*testing.T` to `testing.TB` so the same node setups can be reused by benchmarks

## Why
Issue `#797` asks for communication-layer performance testing so FSC can compare transports with actual measurements instead of assumptions. We already had functional transport tests, but no runnable microbenchmarks in the transport packages themselves.

## Impact
Developers can now run apples-to-apples local benchmarks for websocket and libp2p session round trips and use those numbers as a baseline when changing the comm layer.

## Validation
- `go test -run '^$' -bench BenchmarkSessionRoundTrip -benchtime=1x ./platform/view/services/comm/host/websocket`
- `go test -run '^$' -bench BenchmarkSessionRoundTrip -benchtime=1x ./platform/view/services/comm/host/libp2p`

Fixes #797

### DIFF
--- a/platform/view/services/comm/host/libp2p/p2p_bench_test.go
+++ b/platform/view/services/comm/host/libp2p/p2p_bench_test.go
@@ -1,0 +1,132 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package libp2p
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+)
+
+func BenchmarkSessionRoundTrip(b *testing.B) {
+	benchmarkSessionRoundTrip(b, setupTwoNodes)
+}
+
+func benchmarkSessionRoundTrip(b *testing.B, setup func(testing.TB) (*comm.HostNode, *comm.HostNode)) {
+	b.Helper()
+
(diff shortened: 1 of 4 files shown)


malsomesh9:
I refreshed this branch on top of current `main`, resolved the merge conflict, and collapsed the old unsigned stack into a single GPG-signed commit.

I also updated the PR body to use `Fixes #797` and re-ran the benchmark validation locally:
- `go test -run ^ -bench BenchmarkSessionRoundTrip -benchtime=1x ./platform/view/services/comm/host/websocket`
- `go test -run ^ -bench BenchmarkSessionRoundTrip -benchtime=1x ./platform/view/services/comm/host/libp2p`

One caveat from the bot flow: `/assign` on `#797` is currently rejected because the issue does not have `status: ready for dev`, so the issue-link gate may still remain blocked until that issue becomes assignable.

malsomesh9:
@mbrandenburger @SaidAltury-ibm

The issue has been resolved from my side and the changes are ready for review.

Could you please assign it to me so I can proceed with the final review workflow?

PR #1344, opened 2026-04-27
bug(config service): yaml tags are not respected #1343

### LINKED ISSUE #1343: bug(config service): yaml tags are not respected
If we try to unmarshal the following data structure,

```
// TLSConfig specifies TLS settings for secure communication.
// Supports three modes: no TLS, server TLS (rootCerts only), and mutual TLS (all fields).
type TLSConfig struct {
	// Enabled indicates whether TLS is enabled for this endpoint.
	Enabled bool `yaml:"enabled"`
	// ClientKeyPath is the path to TLS client private key
	ClientKeyPath string `yaml:"clientKey,omitempty"`
	// ClientCertPath is the path to TLS client certificate
	ClientCertPath string `yaml:"clientCert,omitempty"`
	// RootCertPaths are the paths to TLS root certificates.
	RootCertPaths []string `yaml:"rootCerts,omitempty"`
	// ServerNameOverride is the server name to use for TLS hostname verification.
	ServerNameOverride string `yaml:"serverNameOverride,omitempty"`
}
```

`RootCertPaths` will not be filled property because the config service ignores the tag.

adecaro:
This PR fix the bug identified by #1343 by using the following config 

```go
koanf.UnmarshalConf{
		DecoderConfig: config,
		Tag:           "yaml",
	}
```

This way `koanf` will process the `yaml` field tags.

### DIFF
--- a/docs/platform/view/services/config-service.md
+++ b/docs/platform/view/services/config-service.md
@@ -97,3 +97,14 @@ func (s *MyService) UpdateConfig(config *config.Provider, rawYaml []byte) {
 ## Implementation Details
 
 Currently, the configuration service is based on [github.com/knadh/koanf](https://github.com/knadh/koanf), a light-weight, extensible configuration management library for Go. While the `config.Provider` interface abstracts most of these details, the underlying implementation utilizes `koanf` for its powerful parsing, merging, and environment variable substitution capabilities.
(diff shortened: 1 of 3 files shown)


SaidAltury-ibm (approved):
Thank you @adecaro, LGTM

PR #1345, opened 2026-04-27
update(endpoint/service): no ip resolution

### LINKED ISSUE #1348: update(endpoint/service): no ip resolution
Currently, the endpoint service resolves a resolver's address so an IP address.
This causes the following situation: Alice tries to connect to Bob via hostname:port. Alice replaces hostname with the IP. When the connection is establish, the comm layer will enforce that Bob's certificate contains the IP and not the hostname causing a failure in the connection.

adecaro:
This PR does the following: It removes the IP resolution on the resolver's address so that the actual address is used. 
This avoids the following situation: Alice tries to connect to Bob via <hostname>:<port>. Alice replaces hostname with the IP. When the connection is establish, the comm layer will enforce that Bob's certificate contains the IP and not the hostname causing a failure in the connection.

### DIFF
--- a/integration/nwo/fsc/node/crypto_template.go
+++ b/integration/nwo/fsc/node/crypto_template.go
@@ -21,6 +21,7 @@ PeerOrgs:{{ range .PeerOrgs }}
     - 127.0.0.1
     - ::1
     - "::"
+    - host.docker.internal
   {{- end }}
   {{- end }}
   Users:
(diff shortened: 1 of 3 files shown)


SaidAltury-ibm (approved):
Thank you @adecaro, LGTM. can you just link it to an issue so we can track it?

Issue #1358, opened 2026-04-29
testing: extend x509 helper coverage with functional public-key verification


malsomesh9:
Part of #1233

## Summary
Add follow-up unit coverage for `platform/view/services/id/x509` to verify that keys extracted from certificates and PEM payloads are functionally correct for cryptographic operations, not only structurally equal.

## Motivation
PR #1307 added strong baseline coverage for the x509 identity helper package, but review noted that some assertions still stop at comparing key coordinates. We should also prove that the extracted public key can successfully verify signatures produced by the matching private key.

## Scope
- add tests that sign a message with the generated private key
- verify signatures with public keys extracted from certificates and PEM public-key material
- keep the tests focused on x509 helper behavior and verifier integration
- preserve the Stage 2 coverage goal for #1233

## Acceptance Criteria
- `platform/view/services/id/x509` includes a test that validates cryptographic use of the extracted public key
- the new coverage complements the structural equality assertions already added in PR #1307
- `go test ./platform/view/services/id/x509` passes

## Context
Follow-up from review on PR #1307.

PR #1360, opened 2026-04-29
testing: extend x509 helper coverage with functional verification

malsomesh9:
## What changed
Added functional x509 helper coverage in `platform/view/services/id/x509` to prove that extracted public keys are usable for cryptographic verification, not just structurally equivalent after decoding.

## Why
Issue #1358 follows review feedback on the earlier x509 helper coverage: we already checked key structure, but we also need to demonstrate that keys extracted from certificate and public-key PEM material can verify signatures from the matching private key.

## Impact
This closes the testing gap around `PemDecodeKey` and `Deserializer.DeserializeVerifier` by validating real signature verification behavior with low-S ECDSA signatures.

## Validation
- `go test ./platform/view/services/id/x509`

### DIFF
--- a/platform/view/services/id/x509/ecdsa_test.go
+++ b/platform/view/services/id/x509/ecdsa_test.go
@@ -1,0 +1,110 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package x509
+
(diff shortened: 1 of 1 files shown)


malsomesh9:
@mbrandenburger @adecaro @SaidAltury-ibm Plz have a look.

PR #1417, opened 2026-05-09
fabric: tolerate partial peer failures on query

### LINKED ISSUE #474: Allow successful queries if one of the configured peers is down
When doing a fabric.ChaincodeQuery with WithEndorsersFromMyOrg, the FSC queries all configured peers of the own organization. On _any_ query, it validates if all responses match:

https://github.com/hyperledger-labs/fabric-smart-client/blob/110f6327f420c5dc08dbaceea3fd000ed1055115/platform/fabric/core/generic/chaincode/invoke.go#L119

In some situations this may be important (i.e. querying other orgs, not trusting all own peers). But sometimes availibility should be considered more important than the integrity. For example: I have three peers in my organization and I want to update one. The client should still be able to query (and invoke).

Perhaps it would be worthwhile to have a flag to 'trust any response' and it would just take the first of all the responses that has a valid payload?

malsomesh9:
## Summary
- allow query collection to keep successful proposal responses when some peers fail
- preserve the stricter fail-on-any-error behavior for endorsement and submit paths
- add regression tests for partial peer failure and all-peers-down cases

## Testing
- go test ./platform/fabric ./platform/fabric/core/generic/chaincode -count=1

Fixes #474

### DIFF
--- a/platform/fabric/core/generic/chaincode/invoke.go
+++ b/platform/fabric/core/generic/chaincode/invoke.go
@@ -9,6 +9,7 @@ package chaincode
 import (
 	"bytes"
 	"context"
+	stderrors "errors"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


malsomesh9:
@mbrandenburger @SaidAltury-ibm Plz have a look.

PR #1458, opened 2026-05-13
 test: add unit coverage for core generic rwset

### LINKED ISSUE #1441: Unit test for platform/fabric/core/generic/rwset

Ayush3941:
closes #1441
 ## Description
  This PR adds focused unit tests for `platform/fabric/core/generic/rwset` to satisfy Stage 2 coverage requirements.

  ## What Changed
  - Added unit tests for composite key helpers:
    - `CreateCompositeKey`
    - `CreateRangeKeysForPartialCompositeKey`
    - `SplitCompositeKey`
    - `validateCompositeKeyAttribute`
  - Added envelope unpacking tests:
    - success path for valid endorser transaction envelopes
    - error paths for invalid bytes/payload/header/signature/transaction data
  - Added handler and loader tests:
    - `NewEndorserTransactionHandler` and `endorserTransactionReader.Read`
    - `Loader.AddHandlerProvider`, `GetRWSetFromEvn`, `GetRWSetFromETx`, `GetInspectingRWSetFromEvn`
    - success and failure scenarios using test-local fakes/mocks
  - Added processor manager tests:
    - request ID behavior
    - processor registration/default processor behavior
    - `ProcessByID` envelope path, transaction path, and error propagation


platform/fabric/core/generic/rwset/composite_test.go
```diff
@@ -0,0 +1,182 @@
...
+			name:        "invalid utf8 attribute",
+			objectType:  "asset",
+			attributes:  []string{"\xff\xfe"},
+			expectError: true,
+			errorMsg:    "not a valid utf8 string",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
mbrandenburger: This loop-variable capture is unnecessary with go version > 1.22. Please check all your tests.

platform/fabric/core/generic/rwset/handler_loader_test.go
```diff
@@ -0,0 +1,401 @@
...
+			nil,
+			nil,
+			&fakeInspector{
+				inspectFunc: func(_ context.Context, _ []byte, _ ...cdriver.Namespace) (fdriver.RWSet, error) {
+					return nil, stderrors.New("inspect-failed")
+				},
+			},
+		)
+		_, _, err := loader.GetInspectingRWSetFromEvn(context.Background(), cdriver.TxID(chdr.TxId), mustMarshalProto(t, env))
+		require.EqualError(t, err, "inspect-failed")
```
mbrandenburger: I suggest to use `require.ErrorContains` over `require.EqualError` to be consistent. Please check all tests.

platform/fabric/core/generic/rwset/handler_loader_test.go
```diff
@@ -0,0 +1,401 @@
...
+			nil,
+			nil,
+			&fakeInspector{
+				inspectFunc: func(_ context.Context, rwsetBytes []byte, _ ...cdriver.Namespace) (fdriver.RWSet, error) {
+					require.Equal(t, results, rwsetBytes)
+					return expectedRWS, nil
+				},
+			},
+		)
+		rws, tx, err := loader.GetInspectingRWSetFromEvn(context.Background(), cdriver.TxID(chdr.TxId), mustMarshalProto(t, env))
```
mbrandenburger: please use `t.Context()` instead of `context.Background()`. Check all tests.

platform/fabric/core/generic/rwset/handler_loader_test.go
```diff
@@ -0,0 +1,401 @@
...
+		require.NoError(t, err)
+		require.Same(t, expectedRWS, rws)
+		require.Equal(t, chdr.TxId, tx.ID())
+		require.Equal(t, "network", tx.Network())
+		require.Equal(t, "mychannel", tx.Channel())
```
mbrandenburger, line 385: This is a bit of a surprise .... the loader above takes "channel" as argument instead of "mychannel". Can you please check why this check succeeds? Did we discover a bug here?

platform/fabric/core/generic/rwset/handler_loader_test.go
```diff
@@ -0,0 +1,401 @@
...
+			"network",
+			"channel",
+			&fakeEnvelopeService{existsFn: func(_ context.Context, _ string) bool { return false }},
+			nil,
+			nil,
+			nil,
+		)
+		_, _, err := loader.GetRWSetFromEvn(context.Background(), "tx1")
+		require.Error(t, err)
+		require.ErrorContains(t, err, "envelope does not exists")
```
mbrandenburger: I think we found a typo in the code here. We should use the chance to fix it.

platform/fabric/core/generic/rwset/handler_loader_test.go
```diff
@@ -0,0 +1,401 @@
...
+				existsFn: func(_ context.Context, _ string) bool { return true },
+				loadFn:   func(_ context.Context, _ string) ([]byte, error) { return mustMarshalProto(t, env), nil },
+			},
+			nil,
+			nil,
+			nil,
+		)
+		_, _, err := loader.GetRWSetFromEvn(context.Background(), cdriver.TxID(chdr.TxId))
+		require.Error(t, err)
+		require.ErrorContains(t, err, "header type not support")
```
mbrandenburger: I think we found a typo in the code. Let's use the chance to fix it.

mbrandenburger (changes requested):
Thank you @Ayush3941 for creating this PR. Please see my comments below.

platform/fabric/core/generic/rwset/handler_loader_test.go
Ayush3941: Thanks, I checked this more closely and yes your suspicion was right

In this path the returned transaction channel comes from the envelope header not from the loader’s configured channel.

Specifically what happened was:

* `GetInspectingRWSetFromEvn` unpacks the raw envelope and returns `upe` as the `ProcessTransaction`:
  https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/rwset/loader.go#L118-L138

* `UnpackedEnvelope` stores `Ch: chdr.ChannelId`, and `Channel()` returns `u.Ch`:
  https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/rwset/envelope.go#L123-L151

* In the test, the loader is created with `"channel"`, while the test envelope helper builds the envelope with `"mychannel"`:
  https://github.com/Ayush3941/fabric-smart-client/blob/test/core-generic-rwset-coverage-1441/platform/fabric/core/generic/rwset/handler_loader_test.go#L381-L399
  https://github.com/Ayush3941/fabric-smart-client/blob/test/core-generic-rwset-coverage-1441/platform/fabric/core/generic/rwset/helpers_test.go#L235

So the current behavior is:


tx.Network()  ---> comes from loader context
tx.Channel()  ---> comes from envelope struct
tx.ID()             --->comes from envelope struct


 The fix is for the loader to validate that the envelope belongs to the loader’s configured channel.

Something like:

```go
if chdr.ChannelId != c.Channel {
	return nil, nil, errors.Errorf(
		"channel mismatch, expected [%s], got [%s]",
		c.Channel,
		chdr.ChannelId,
	)
}
```
after 
https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/rwset/loader.go#L126-L129


I also noticed the same invariant gap in the other envelope-loading path.

In `GetRWSetFromEvn`, the loader unmarshals the envelope, extracts the channel header, and then directly dispatches to the handler:

https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/core/generic/rwset/loader.go#L73-L93

At that point the loader has both values available:

```go
c.Channel      // loader configured channel
chdr.ChannelId // envelope header channel
```

but it does not currently validate that they match
So I think the similar guard check should be added there too

since its a minor change should i fix this in same PR or create a new PR? @mbrandenburger

mbrandenburger: Good analizes. I think we can fix the "issues" here and add a channel-mismatch guard as you suggested.

Ayush3941: fixed the channel mismatch issue in eea2464 by adding the guard in both envelope loading paths, as discussed.

also updated the test doubles usage to follow the latest docs/dev/mocks.md guidance.

mbrandenburger (approved):

PR #1463, opened 2026-05-15
fix(fabric/config): add missing GetInt prefix wrapper

### LINKED ISSUE #1465: Missing GetInt prefix wrapper in fabric config Service silently zeroes ordering.connectionPoolSize and ordering.numRetries
## Bug

`platform/fabric/core/generic/config/Service` exposes prefixed accessors (`GetString`, `GetBool`, `IsSet`, `GetDuration`, `GetPath`, `UnmarshalKey`) that wrap the embedded `Configuration` with `"fabric.<prefix>."` for all keys.

The **`GetInt` wrapper is missing**, so callers fall through Go's embedded-interface method promotion to `Configuration.GetInt`, which does **not** prepend the prefix.

## Affected callers

- `Service.OrdererConnectionPoolSize()`: `IsSet` branch reports the key present (prefixed correctly), then immediately calls `s.GetInt(...)` which misses (unprefixed) and returns `0`.
- `Service.BroadcastNumRetries()`: same pattern — reads `ordering.numRetries` unprefixed.

## Reproducer

In the fabric config yaml:

```yaml
fabric:
  default:
    ordering:
      connectionPoolSize: 20
```

Expected: `OrdererConnectionPoolSize() == 20`
Actual: returns `0`.

Worse, this is silently **inferior** to omitting the key entirely (which falls back to the `defaultOrderingConnectionPoolSize = 10`).

## Downstream impact

In a production deployment (CBDC project on fabric-x), this caused `BFTBroadcaster.poolSize == 0`, exhausting the slow path on every broadcast and producing 100% submit failures under load. Diagnosis required reading the source — there is no warning logged.

## Fix

Add the missing `GetInt` wrapper that mirrors the existing accessors. PR will follow.

/cc maintainers — would like to self-assign but lack the org permission. Happy to take this.

EvanYan1024:
Fixes #1465

## Summary

`Service` exposes prefixed accessors (`GetString`, `GetBool`, `IsSet`, `GetDuration`, `GetPath`, `UnmarshalKey`) that wrap the embedded `Configuration` with `"fabric.<prefix>."` for all keys. The `GetInt` wrapper was missing, so callers like `OrdererConnectionPoolSize` and `BroadcastNumRetries` silently fell through Go's embedded-interface method promotion to `Configuration.GetInt`, which does **not** prepend the prefix.

The `IsSet` branch correctly reported the key as present (because `IsSet` is prefixed), but the immediately following value lookup missed and returned `0`.

Practical effect: setting `fabric.default.ordering.connectionPoolSize: 20` in yaml made `OrdererConnectionPoolSize()` return `0` instead of `20` — silently worse than leaving the key unset (which falls back to the `10` default). The same bug applies to `ordering.numRetries` via `BroadcastNumRetries`.

## Fix

Add the missing 4-line `GetInt` wrapper between `GetString` and `GetDuration`, following the exact pattern used by the other accessors.

## Test plan

- [x] Extended `service_test.go` to capture `GetIntArgsForCall(...)` and assert the prefixed key (`fabric.mynet.ordering.numRetries` / `fabric.mynet.ordering.connectionPoolSize`) is what gets passed down to `Configuration`.
- [x] `go test ./platform/fabric/core/generic/config/...` passes.
- [x] Without the wrapper (verified via `git stash`), the new assertions fail with `expected "fabric.mynet.ordering.numRetries" actual "ordering.numRetries"` — confirms the new tests catch the bug.

### DIFF
--- a/platform/fabric/core/generic/config/service.go
+++ b/platform/fabric/core/generic/config/service.go
@@ -262,6 +262,10 @@ func (s *Service) GetString(key string) string {
 	return s.Configuration.GetString("fabric." + s.prefix + key)
 }
 
+func (s *Service) GetInt(key string) int {
+	return s.Configuration.GetInt("fabric." + s.prefix + key)
+}
+
 func (s *Service) GetDuration(key string) time.Duration {
 	return s.Configuration.GetDuration("fabric." + s.prefix + key)
 }
--- a/platform/fabric/core/generic/config/service_test.go
+++ b/platform/fabric/core/generic/config/service_test.go
@@ -269,9 +269,11 @@ func TestServiceGetters(t *testing.T) {
 	require.Equal(t, 3, svc.MSPCacheSize())
 
 	// Ordering retries
-	m.GetIntReturnsOnCall(0, 5)
+	getIntCallsBefore := m.GetIntCallCount()
+	m.GetIntReturnsOnCall(getIntCallsBefore, 5)
 	require.Equal(t, 5, svc.BroadcastNumRetries())
-	m.GetIntReturnsOnCall(1, 0)
+	require.Equal(t, "fabric.mynet.ordering.numRetries", m.GetIntArgsForCall(getIntCallsBefore))
+	m.GetIntReturnsOnCall(getIntCallsBefore+1, 0)
 	require.Equal(t, 3, svc.BroadcastNumRetries())
 
 	m.IsSetReturns(true)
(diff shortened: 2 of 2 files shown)


SaidAltury-ibm (approved):
Thank you @EvanYan1024 this is a good catch!, LGTM. please rebase your branch on main

EvanYan1024:
> Thank you @EvanYan1024 this is a good catch!, LGTM. please rebase your branch on main

done

PR #1483, opened 2026-05-27
fix(grpc): cache *grpc.ClientConn per (service, network) in ClientProvider

### LINKED ISSUE #1482: Connection/goroutine leak: fabricx committer ClientProvider dials a new grpc.ClientConn on every call
### What happened?

`ClientProvider.NotificationServiceClient` and `QueryServiceClient`
(`platform/fabricx/core/committer/grpc/grpc.go`) call `grpc.NewClient` on every
invocation and return a fresh `*grpc.ClientConn` that the caller never closes.
Each `ClientConn` carries several long-lived goroutines (http2 reader,
loopyWriter, callback serializers, resolver watcher, balancer). Downstream
callers invoke these methods on the hot path — e.g. fabric-token-sdk's
`qe.Executor` calls the query-service client on every state query — so both the
goroutine count and the number of open connections grow linearly with
throughput and are never reclaimed.

In a multi-hour load test on a fabricx-based deployment, a single process
accumulated ~104k goroutines, dominated by grpc per-connection workers:

| goroutine | count |
|---|---|
| grpcsync.CallbackSerializer.run | 69,447 |
| http2Client.reader | 11,591 |
| loopyWriter.run | 11,592 |
| dnsResolver.watcher | 11,589 |

A heap profile (`inuse_objects`, focus `NewClient`) traced the allocation chain to:

    qe.Executor.QueryState
      → queryservice.RemoteQueryServiceProvider.Get
        → committer/grpc.ClientProvider.QueryServiceClient
          → ClientConn → grpc.NewClient   ← a new conn on every call

### How can we reproduce it?

Call `ClientProvider.QueryServiceClient(network)` (or `NotificationServiceClient`)
repeatedly for the same network and issue an RPC on each returned connection (as
`qe.Executor` does per state query). Each call dials a brand-new connection;
goroutines grow by a connection's worth per call and are never released.

A self-contained reproduction is included in the linked PR: against one
in-process gRPC server with identical usage, the dial-per-call path leaks ~8
goroutines per call (240 over 30 calls) while a cached provider adds ~0.

### Environment

- fabric-smart-client: `main` (also present in v0.11.x)
- Affected file: `platform/fabricx/core/committer/grpc/grpc.go`
- grpc-go: as pinned by go.mod

### Additional Information

Proposed fix: cache `*grpc.ClientConn` per (service, network) so each connection
is dialed once and reused. PR linked below.

EvanYan1024:
## Summary

`ClientProvider.NotificationServiceClient` and `QueryServiceClient` called
`grpc.NewClient` on every invocation. `grpc.NewClient` spawns several long-lived
goroutines per `ClientConn` (resolver watcher, callback serializers, http2
reader/writer, balancer); when downstream callers invoke these on the hot path —
e.g. fabric-token-sdk's `qe.Executor` calls `QueryServiceClient` on every state
query — the goroutine count grows linearly with throughput and the underlying
connections leak.

This caches `*grpc.ClientConn` per (service, network) using two `sync.Map`s. The
first call dials and stores; subsequent calls return the cached connection. Under
a benign cold-start race two callers may both dial; the loser closes its
connection and returns the winner's via `LoadOrStore`.

Fixes #1482

## Evidence

In a multi-hour load test on a fabricx-based deployment, a single process
accumulated ~104k goroutines, dominated by grpc per-connection workers (69,447
`CallbackSerializer.run`, ~11.6k each of `http2Client.reader` / `loopyWriter.run`
/ `dnsResolver.watcher`). A heap profile (`inuse_objects`, focus `NewClient`)
traced the chain to `qe.Executor.QueryState → RemoteQueryServiceProvider.Get →
ClientProvider.QueryServiceClient → grpc.NewClient`.

## Test plan

- Added `TestClientProvider_CachingPreventsPerCallGoroutineLeak`: drives the
  cached provider vs. the previous dial-per-call behavior against one real
  in-process server with identical usage, asserting the cached path adds ~no
  goroutines while dialing per call leaks a connection's worth each time (0 vs
  240 goroutines over 30 calls locally). Doubles as a regression guard (removing
  the cache fails `require.Same` and the goroutine-delta assertion).
- Extended the `NotificationServiceClient`/`QueryServiceClient` unit tests to
  assert per-network caching (same `*ClientConn` returned, config loaded once).
- `go test -race ./platform/fabricx/core/committer/grpc/...` passes.

## Notes

- Closing the losing duplicate before any RPC is safe per `grpc.ClientConn.Close`
  semantics.
- gRPC's automatic reconnect handles transient disconnects for an unchanged
  target; a target/TLS change would not be picked up while the cached conn lives
  — acceptable for a process-lifetime provider.
- Connections live until process exit; happy to add a `Close()` hook if preferred.

### DIFF
--- a/platform/fabricx/core/committer/grpc/grpc.go
+++ b/platform/fabricx/core/committer/grpc/grpc.go
@@ -37,11 +37,7 @@ type ServiceConfigProvider interface {
 
 // ClientProvider provides gRPC client connections for a given network.
 //
-// Connections are cached per (service, network). grpc.NewClient spawns ~6
-// long-lived goroutines per ClientConn (resolver watcher, callback serializers,
-// http2 reader/writer, balancer); callers like fabric-token-sdk's qe.Executor
-// invoke these methods on every state query, so without caching the goroutine
-// count grows linearly with throughput and leaks the underlying connections.
+// Connections are cached per (service, network) to prevent goroutine groth.
 type ClientProvider struct {
 	// configProvider is used to retrieve the configuration for a network.
 	configProvider ServiceConfigProvider
--- a/platform/fabricx/core/committer/grpc/grpc.go
+++ b/platform/fabricx/core/committer/grpc/grpc.go
@@ -10,6 +10,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"os"
+	"sync"
 	"time"
 
 	"google.golang.org/grpc"
@@ -35,9 +36,18 @@ type ServiceConfigProvider interface {
 }
 
 // ClientProvider provides gRPC client connections for a given network.
+//
+// Connections are cached per (service, network). grpc.NewClient spawns ~6
+// long-lived goroutines per ClientConn (resolver watcher, callback serializers,
+// http2 reader/writer, balancer); callers like fabric-token-sdk's qe.Executor
+// invoke these methods on every state query, so without caching the goroutine
+// count grows linearly with throughput and leaks the underlying connections.
 type ClientProvider struct {
 	// configProvider is used to retrieve the configuration for a network.
 	configProvider ServiceConfigProvider
+
+	notificationConn sync.Map // network -> *grpc.ClientConn
+	queryConn        sync.Map // network -> *grpc.ClientConn
 }
 
 // NewClientProvider returns a new ClientProvider instance.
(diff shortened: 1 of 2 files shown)


adecaro:
good catch @EvanYan1024 , thanks a lot 🙏

platform/fabricx/core/committer/grpc/grpc.go
```diff
@@ -35,9 +36,18 @@ type ServiceConfigProvider interface {
 }
 
 // ClientProvider provides gRPC client connections for a given network.
+//
+// Connections are cached per (service, network). grpc.NewClient spawns ~6
+// long-lived goroutines per ClientConn (resolver watcher, callback serializers,
+// http2 reader/writer, balancer); callers like fabric-token-sdk's qe.Executor
+// invoke these methods on every state query, so without caching the goroutine
+// count grows linearly with throughput and leaks the underlying connections.
```
mbrandenburger:
```suggestion
// Connections are cached per (service, network) to prevent goroutine groth.
```

mbrandenburger (approved):
Thank you @EvanYan1024 for submitting this PR. Good catch!

PR #1485, opened 2026-05-28
fix(fabric/ordering): always discard failed connections in BFT broadcast

### LINKED ISSUE #1484: Connection/semaphore leak: BFT broadcast skips cleanup of failed connections on success-with-partial-failure
### What happened?

`platform/fabric/core/generic/ordering/bft.go`'s `BFTBroadcaster.Broadcast()`
only runs its `usedConnections` cleanup on the overall-failure path (when
`counter < threshold`). When a broadcast partially fails but still clears the
BFT success threshold — e.g. 3 of 4 orderers ack SUCCESS and the 4th errors on
`Send` / `Recv` or returns a non-`SUCCESS` status — the failed connections are
appended to `usedConnections` but never get discarded. Their `connSem`
semaphore units stay permanently held and their gRPC streams stay open for the
lifetime of the process.

Under sustained partial-failure conditions (one slow/flaky orderer that
intermittently errors on `Recv`), the leak compounds: each broadcast leaks
one semaphore unit until `connSem` is fully exhausted, at which point
`getConnection` falls back to its slow path (waiting on the pool channel) and
overall ordering throughput collapses.

### How can we reproduce it?

Configure four orderers with one that intermittently errors on `Recv` (e.g.
slow/flaky network, process restart mid-stream). After enough partial-success
broadcasts, `connSem` fully drains and `getConnection`'s `Acquire` calls start
to block until pool conns are returned, throttling broadcast throughput.

The linked PR adds a unit test that reproduces the leak deterministically:
four fake orderers, threshold 3, three ack `SUCCESS` and the fourth errors on
`Recv`; pre-fix the failed connection's semaphore unit is never released
(asserted via `connSem.TryAcquire(1)` returning false after `Broadcast`
returns nil).

### Environment

- fabric-smart-client: `main` (also present on v0.10.x / v0.11.x — the buggy
  pattern predates recent ordering refactors)
- Affected file: `platform/fabric/core/generic/ordering/bft.go`
  (function `BFTBroadcaster.Broadcast`)

### Additional Information

Proposed fix: move the `usedConnections` cleanup loop above the
`counter >= threshold` check so it runs on every path. Successful connections
are already returned to their pools via `releaseConnection` (which does not
touch the semaphore), so the change only affects the previously-leaking
failed-connection set. PR linked below.

EvanYan1024:
## Summary

`BFTBroadcaster.Broadcast()` only ran its `usedConnections` cleanup on the
overall-failure path (`counter < threshold`). When a broadcast partially failed
but still cleared the BFT success threshold — e.g. 3 of 4 orderers acked, the
4th errored on `Send` / `Recv` or returned a non-`SUCCESS` status — the failed
connections never got discarded. Their `connSem` semaphore units stayed held
and their gRPC streams stayed open for the lifetime of the process.

This moves the cleanup loop above the threshold check so it runs on every
path. Successful connections are already returned to their pools by
`releaseConnection` (which does not touch the semaphore), so the change only
affects the previously-leaking failed-connection set.

Closes #1484

## Impact

Under sustained partial-failure conditions (one slow/flaky orderer that
intermittently errors on `Recv`), the leak compounds: each broadcast leaks
one semaphore unit until `connSem` is fully exhausted, at which point
`getConnection` falls back to its slow path and overall ordering throughput
collapses. This matches the symptom of "pool gradually fills with phantom
in-use slots over hours" observed under sustained load in our deployment.

## Test plan

- Added `TestBFTBroadcaster_DiscardsFailedConnectionsOnPartialFailureSuccess`:
  drives `Broadcast()` against four fake orderers (threshold = 3) where three
  ack `SUCCESS` and the fourth errors on `Recv`; asserts the broadcast
  returns nil **and** exactly one `connSem` unit is released (via
  `TryAcquire(1)` returning true once, then false). Pre-fix the test fails
  (no unit released); post-fix it passes. The test pre-fills the per-orderer
  connection pools so `getConnection` never touches `ClientFactory`, and uses
  a `fakeBroadcastStream` (the `Broadcast` interface from `client.go`) so each
  orderer's outcome is independently controllable.
- `go vet`, `gofmt -l` clean; `go test -race ./platform/fabric/core/generic/ordering/...` passes.

## Notes

- The fix is a control-flow refactor (move one `for` loop), with no API change
  and no behavior change for the no-failure success path (`usedConnections`
  empty, loop is a no-op) or the overall-failure path (same cleanup runs in
  the same place relative to the threshold check that now precedes it).

### DIFF
--- a/platform/fabric/core/generic/ordering/bft_test.go
+++ b/platform/fabric/core/generic/ordering/bft_test.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package ordering
 
 import (
-	"context"
 	"errors"
 	"testing"
 	"time"
@@ -68,7 +67,7 @@ func TestBFTBroadcaster_DiscardsFailedConnectionsOnPartialFailureSuccess(t *test
 
 	// Pre-fill the pools and hold the matching semaphore units, so
 	// getConnection bypasses ClientFactory and discard accounting balances.
-	require.NoError(t, b.connSem.Acquire(context.Background(), int64(len(orderers))))
+	require.NoError(t, b.connSem.Acquire(t.Context(), int64(len(orderers))))
 	streams := map[string]*fakeBroadcastStream{
 		"o1": {status: common.Status_SUCCESS},
 		"o2": {status: common.Status_SUCCESS},
@@ -79,7 +78,7 @@ func TestBFTBroadcaster_DiscardsFailedConnectionsOnPartialFailureSuccess(t *test
 		b.connectionPool(o.Address) <- &Connection{Stream: streams[o.Address]}
 	}
 
-	err := b.Broadcast(context.Background(), &common.Envelope{})
+	err := b.Broadcast(t.Context(), &common.Envelope{})
 	require.NoError(t, err)
 
 	// Exactly one unit should be free: o4's discarded connection.
--- a/platform/fabric/core/generic/ordering/bft_test.go
+++ b/platform/fabric/core/generic/ordering/bft_test.go
@@ -1,0 +1,88 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ordering
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	ab "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc"
+)
+
+// fakeBroadcastStream is a Broadcast-interface stub with per-instance outcomes.
+type fakeBroadcastStream struct {
+	sendErr error
+	recvErr error
+	status  common.Status
+}
+
+func (f *fakeBroadcastStream) Send(*common.Envelope) error { return f.sendErr }
+
+func (f *fakeBroadcastStream) Recv() (*ab.BroadcastResponse, error) {
+	if f.recvErr != nil {
+		return nil, f.recvErr
+	}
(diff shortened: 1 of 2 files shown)


platform/fabric/core/generic/ordering/bft_test.go
```diff
@@ -0,0 +1,88 @@
...
+	cfg := &fakeConfigService{
+		poolSize: len(orderers),
+		retries:  1,
+		orderers: orderers,
+	}
+	b := NewBFTBroadcaster(cfg, nil, nil)
+
+	// Pre-fill the pools and hold the matching semaphore units, so
+	// getConnection bypasses ClientFactory and discard accounting balances.
+	require.NoError(t, b.connSem.Acquire(context.Background(), int64(len(orderers))))
```
mbrandenburger:
```suggestion
	require.NoError(t, b.connSem.Acquire(t.Context(), int64(len(orderers))))
```

platform/fabric/core/generic/ordering/bft_test.go
```diff
@@ -0,0 +1,88 @@
...
+		"o1": {status: common.Status_SUCCESS},
+		"o2": {status: common.Status_SUCCESS},
+		"o3": {status: common.Status_SUCCESS},
+		"o4": {recvErr: errors.New("induced Recv failure")},
+	}
+	for _, o := range orderers {
+		b.connectionPool(o.Address) <- &Connection{Stream: streams[o.Address]}
+	}
+
+	err := b.Broadcast(context.Background(), &common.Envelope{})
```
mbrandenburger:
```suggestion
	err := b.Broadcast(t.Context(), &common.Envelope{})
```

mbrandenburger (approved):
Thank you @EvanYan1024 for working on this!

platform/fabric/core/generic/ordering/bft_test.go
```diff
@@ -0,0 +1,88 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package ordering
+
+import (
+	"context"
```
mbrandenburger:
```suggestion
```

mbrandenburger (approved):

PR #1490, opened 2026-05-29
perf(bft): per-orderer channel-slot pool to fix semaphore mutex starvation

### LINKED ISSUE #1489: BFT broadcaster getConnection serializes all callers on a shared semaphore mutex and busy-spins on cancellation
### What happened?

`BFTBroadcaster.getConnection` (`platform/fabric/core/generic/ordering/bft.go`) bounds orderer connections with a single shared `golang.org/x/sync/semaphore.Weighted`; when the pool is empty it retries `Acquire` in a `for` loop with a 1s timeout. Under high broadcast concurrency this has three failure modes:

1. **Shared-mutex starvation** — every caller contends on the one semaphore's internal `sync.Mutex` regardless of which orderer it dials, so the semaphore (not the network or the orderers) becomes the bottleneck.
2. **Busy-spin on cancellation** — the acquire loop has no `ctx.Done()` case, so a caller whose context is cancelled while no connection is available loops `Acquire` → immediate error → retry, pinning a CPU.
3. **Cross-orderer starvation** — because the semaphore is global, connections to the first orderers can exhaust the shared permits and permanently starve callers pinned to the remaining orderers.

In our deployment under sustained high broadcast concurrency (c≈1000), broadcast submit p99 reached ~6s with goroutines parked in `semaphore.(*Weighted).Acquire`.

### How can we reproduce it?

A microbenchmark over `getConnection` (connection creation mocked, callers ≫ pool size, ~1ms simulated send/recv hold) reproduces it deterministically:

1. **Single orderer**, 48 callers, pool 8 → `getConnection` p99 ≈ 5s (whole-second acquire timeouts). The distribution is bimodal: a few callers hit the reuse fast path at ~0 while the rest starve.
2. **Four orderers**, 64 callers, pool 8 → livelock: callers pinned to orderers whose permits were taken by others never acquire a connection.
3. **Cancelled context**: with the pool exhausted, a `getConnection` call with an already-cancelled context does not return (busy-spins).

### Environment

- FSC version: latest main (`4544a0cd`)
- Go Version: go1.26.3
- Operating system: Linux

### Additional Information

Fix in progress — replace the global semaphore with per-orderer channel-slot acquire (three-stage, non-spinning, with a `ctx.Done()` exit). A PR will reference this issue.

EvanYan1024:
Fixes #1489

## Problem

`BFTBroadcaster.getConnection` bounds orderer connections with a single shared `golang.org/x/sync/semaphore.Weighted` and spins on a 1s-timeout `Acquire` retry when the pool is empty. Under high broadcast concurrency this serializes every caller on the one semaphore mutex regardless of target orderer; a cancelled caller busy-spins (no `ctx.Done()` exit); and because the semaphore is global, connections to some orderers can starve callers pinned to others.

## Change

Per-orderer state replaces the global semaphore:

- a fixed-capacity `slots chan struct{}` that bounds connections per orderer without a mutex shared across orderers, and
- the existing buffered pool of idle `*Connection` for reuse.

`getConnection` becomes a three-stage, non-spinning acquire: try pool → try slot (create) → block on `select { pool / slot / ctx.Done() }`. `SendAndRecv` wraps the send+recv round-trip in a derived context so a hung `Recv` is cancelled and torn down instead of leaking.

## Tests

- Adapt `TestBFTBroadcaster_DiscardsFailedConnectionsOnPartialFailureSuccess` to the per-orderer slot accounting.
- Add `TestBFTBroadcaster_GetConnectionHonorsContextCancellation` — returns promptly on a cancelled context (previously busy-spun).

## Microbenchmark

`getConnection` wait latency, semaphore vs channel-slot (connection creation mocked, callers ≫ pool size, 1ms simulated send/recv hold). Single orderer, 48 callers, pool 8 (same cap on both builds):

| `getConnection` wait | semaphore | channel-slot |
| --- | ---: | ---: |
| p50 | 0s | 5.7ms |
| p99 | 5.00s | 5.8ms |
| max | 5.00s | 5.8ms |

The semaphore build is bimodal — a few callers hit the reuse fast-path while the rest starve on whole-second acquire timeouts; the channel-slot build queues callers fairly (~865× lower p99). Four orderers, 64 callers, pool 8: the semaphore build livelocks (callers pinned to permit-starved orderers never acquire); the channel-slot build completes (p99 ~1.2ms).

## Verified

- `go build ./...`, `go test ./platform/fabric/core/generic/ordering/`
- Our deployment (c≈1000 stress): submit p99 6.18s → 1.59s, 678 TPS sustained.

### DIFF
--- a/platform/fabric/core/generic/ordering/bft.go
+++ b/platform/fabric/core/generic/ordering/bft.go
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	common2 "github.com/hyperledger/fabric-protos-go-apiv2/common"
-	"golang.org/x/sync/semaphore"
 	"google.golang.org/grpc/status"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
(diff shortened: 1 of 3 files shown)


mbrandenburger (approved):
Thank you @EvanYan1024 for working on this! Great stuff.

PR #1499, opened 2026-06-01
refactor!: replace base58 encoding with base64

### LINKED ISSUE #1498: Remove github.com/mr-tron/base58 dep
### Description

The websocket impl uses github.com/mr-tron/base58 for serializing sha256 hashes of ecdsa pub keys. base58 is used mainly in libp2p wheres in FSC we consistently use base64 encoding. This task is about to clean this up.

### Implementation Steps

- replace base58 with base64
- test it

### Additional Information

_No response_

mbrandenburger:
Closes #1498 

Breaking change: While change is not directly exposed to the user, it's a breaking change and requires all nodes in the network to upgrade to that version.

### DIFF
--- a/platform/view/services/comm/host/websocket/pki.go
+++ b/platform/view/services/comm/host/websocket/pki.go
@@ -10,8 +10,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/sha256"
 	"crypto/x509"
-
-	"github.com/mr-tron/base58/base58"
+	"encoding/base64"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 )
@@ -40,5 +39,7 @@ func ecdsaPubKeyID(key *ecdsa.PublicKey) ([]byte, error) {
 	}
 
 	h := sha256.Sum256(marshaledPubKey)
-	return []byte(base58.Encode(h[:])), nil
+	buf := make([]byte, base64.StdEncoding.EncodedLen(len(h[:])))
+	base64.StdEncoding.Encode(buf, h[:])
+	return buf, nil
 }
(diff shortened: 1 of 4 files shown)


platform/view/services/comm/host/websocket/pki.go
```diff
@@ -40,5 +39,7 @@ func ecdsaPubKeyID(key *ecdsa.PublicKey) ([]byte, error) {
 	}
 
 	h := sha256.Sum256(marshaledPubKey)
-	return []byte(base58.Encode(h[:])), nil
+	buf := make([]byte, base64.StdEncoding.EncodedLen(len(h[:])))
+	base64.StdEncoding.Encode(buf, h[:])
+	return buf, nil
```
SaidAltury-ibm, line 44: can we consider replacing with: return []byte(base64.StdEncoding.EncodeToString(h[:])), nil
like you did in auth.go, WYT?

SaidAltury-ibm (changes requested):
Thank you @mbrandenburger, i think we should point out that this is or cloud be a breaking change WYT?. also left a minor comment.

platform/view/services/comm/host/websocket/pki.go
mbrandenburger: Note that `ecdsaPubKeyID` returns `[]byte`. Using `EncodeToString` and then convert to `[]byte` just adds two extra conversions `[]byte` -> `string` -> `[]byte`.

mbrandenburger (commented):
Good catch! While this change is not directly exposed to the user, it's a breaking change and requires all nodes in the network to upgrade to that version.

SaidAltury-ibm (approved):
LGTM

PR #1521, opened 2026-06-08
test: Unit test for platform/view/services/grpc  #1421

### LINKED ISSUE #1421: Unit test for platform/view/services/grpc

SaidAltury-ibm:
- added new test file for fields
 - renamed the mock folder to follow the mocks guide and be filtered by the coverage report
 
- test coverage is now 85.76%
 
 closes  #1421


platform/view/services/grpc/generate_test.go
```diff
@@ -4,6 +4,6 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-//go:generate protoc --proto_path=testdata/grpc --go_out=plugins=grpc,paths=source_relative:testpb testdata/grpc/test.proto
+//go:generate protoc --proto_path=testdata/grpc --go_out=paths=source_relative:testpb --go-grpc_out=paths=source_relative:testpb testdata/grpc/test.proto
```
mbrandenburger, line 7: Can you please check that this proto generate directive is compatible with the `make` target to "generate the protos? When looking at this, I think we should generate the protos via go:generate or via make target. WDYT?

platform/view/services/grpc/serverstatshandler_test.go
```diff
@@ -19,7 +24,7 @@ import (
 
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils"
 	grpc3 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc/metricsfakes"
+	metricsfakes "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc/mock"
```
mbrandenburger: Should we better use the actual package name in the code below rather than renaming the import, which is not needed.

mbrandenburger (commented):
Thank you @SaidAltury-ibm for creating this PR. Please see my comments below.

platform/view/services/grpc/generate_test.go
SaidAltury-ibm: will handle this in another issue

mbrandenburger (approved):

PR #1540, opened 2026-06-15
test(tracing): add unit tests for platform/view/services/tracing

### LINKED ISSUE #1424: Unit test for platform/view/services/tracing

sid200727:
Closes #1424

## Summary
Adds unit tests for `platform/view/services/tracing` increasing coverage from **27.1% to 82.6%**.

## Coverage breakdown
-> `span.go` - NewLabels, Append, ToLabels, End, AddEvent, SetAttributes: 100%
-> `config.go` - RegisterReplacer, Replacers, extractMetricsOpts: 90%+
-> `provider.go` - NewProviderWithBackingProvider, NewProviderWithNodeName, Tracer: 100%
- >`impl.go` - NewProviderFromConfig (None, Default, Console, File, Otlp error paths)
-> `tracer.go` - Start: 100%

## Testing
Uses `disabled.Provider` as a no-op metrics backend and `noop.TracerProvider` from OpenTelemetry to avoid external dependencies.


SaidAltury-ibm (dismissed):
Thank you, @sid200727, for this PR. I noticed that the tests for `span.go`, `config.go`, `tracer.go`, and `provider.go` are all grouped in `tracing_test.go`. Would it make sense to split them into separate test files so that each source file has its own corresponding test file? WYT?

sid200727:
@SaidAltury-ibm yes! That makes a lot of sense by keeping tests co-located with their source files makes it much easier to navigate and maintain coverage as the codebase evolves. 
I'll split them into separate test files (span_test.go, provider_test.go, impl_test.go, tracer_test.go) and update the PR shortly. 
Thanks for the suggestion!

platform/view/services/tracing/config_additions_test.go
```diff
@@ -0,0 +1,28 @@
...
+	Expect(func() {
+		tracing.RegisterReplacer("github.com_hyperledger-labs_fabric-smart-client_platform", "fsc")
+	}).To(Panic())
+}
+
+func TestReplacers_ContainsDefault(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	r := tracing.Replacers()
+	Expect(r).To(HaveKey("github.com_hyperledger-labs_fabric-smart-client_platform"))
+}
```
mbrandenburger: I think these test cases should be moved into `config_test.go`.

platform/view/services/tracing/span_test.go
```diff
@@ -0,0 +1,54 @@
...
+	Expect(labels).To(ContainElements("color", "blue"))
+}
+
+func TestLabels_Append_InvalidKey(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	l := tracing.NewLabels([]string{"color"})
+	l.Append(attribute.String("", "ignored"))
+	labels := l.ToLabels()
+	Expect(labels).To(ContainElements("color", ""))
+}
```
mbrandenburger: I believe this test is misleading. We want to make sure that the `Append` is a no-op if the label is invalid. Maybe we compare the labels before and after the `Append` operation?

platform/view/services/tracing/tracer_test.go
```diff
@@ -0,0 +1,64 @@
...
+	bp := noop.NewTracerProvider()
+	p := tracing.NewProviderWithBackingProvider(bp, mp)
+
+	tracer := p.Tracer("test-tracer", tracing.WithMetricsOpts(tracing.MetricsOpts{
+		LabelNames: []string{"op"},
+	}))
+
+	_, span := tracer.Start(context.Background(), "my-span")
+	span.SetAttributes(tracing.String("op", "write"))
+	span.End()
```
mbrandenburger, line 178: Shouldn't we test that `SetAttributes` actually does something?

platform/view/services/tracing/tracer_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestTracer_Start_AddEvent(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	mp := &disabled.Provider{}
+	bp := noop.NewTracerProvider()
+	p := tracing.NewProviderWithBackingProvider(bp, mp)
+
+	tracer := p.Tracer("test-tracer", tracing.WithMetricsOpts(tracing.MetricsOpts{}))
+	_, span := tracer.Start(context.Background(), "my-span")
+	span.AddEvent("my-event")
+	span.End()
```
mbrandenburger, line 206: Shouldn't we test that `AddEvent` actually does something?

platform/view/services/tracing/span_test.go
```diff
@@ -0,0 +1,54 @@
...
+	RegisterTestingT(t)
+	l := tracing.NewLabels([]string{})
+	Expect(l.ToLabels()).To(BeEmpty())
+}
+
+func TestNewLabels_WithKeys(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	l := tracing.NewLabels([]string{"key1", "key2"})
+	labels := l.ToLabels()
+	Expect(labels).To(HaveLen(4))
```
mbrandenburger: Instead of checking the length - we should test for `ConsistOf("key1", "", "key2", "")`

platform/view/services/tracing/impl_test.go
```diff
@@ -0,0 +1,74 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(p).NotTo(BeNil())
+}
+
+func TestNewProviderFromConfig_File(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	f, err := os.CreateTemp("", "tracing-test-*.json")
+	Expect(err).NotTo(HaveOccurred())
+	Expect(f.Close()).To(Succeed())
+	defer func() { Expect(os.Remove(f.Name())).To(Succeed()) }()
```
mbrandenburger: this part appears fragile. I suggest that we use `t.TempDir()` and create the file in the test folder. The test harness will take care of cleaning up.

platform/view/services/tracing/config_additions_test.go
```diff
@@ -0,0 +1,28 @@
...
+	RegisterTestingT(t)
+	Expect(func() {
+		tracing.RegisterReplacer("github.com_hyperledger-labs_fabric-smart-client_platform", "fsc")
+	}).To(Panic())
+}
+
+func TestReplacers_ContainsDefault(t *testing.T) { //nolint:paralleltest
+	RegisterTestingT(t)
+	r := tracing.Replacers()
+	Expect(r).To(HaveKey("github.com_hyperledger-labs_fabric-smart-client_platform"))
```
mbrandenburger: I think this test reveals an issue with the actual production code. The `Replacers()` function guards the underlying map structure using a mutex.

See the impl:
```go
func Replacers() map[string]string {
	replacersMutex.RLock()
	defer replacersMutex.RUnlock()
	return replacers
}
```

However, once returned, the users can read and modify the map without any protection.

Instead, `Replacers()` should return a copy of the map. We could use the chance to fix this issue here and test expected behavior accordingly. WDYT?

mbrandenburger (changes requested):
Thank you @sid200727 for working on this. Really good progress for our coverage. 

I think we should not use gomega asserts for these unit tests. I know that some tests exist using gomega, however, I think we could easily rewrite them and use `testify/require` instead. This would also allow us to run tests using `t.Parallel()` (in many cases). Do you think it is reasonable to change the tests?

See below a few more comments.

@SaidAltury-ibm we should document in the development guidelines that we "prefer" `testify` assertions for unit tests.

sid200727:
Thank you  @mbrandenburger 🙏 All the suggestions make sense, especially the behavioral assertions (checking that Append/SetAttributes/AddEvent actually change state, not just that they run) and the Replacers() copy-safety test, that's a good way to document the contract. 
I'll work through the threads and push an updated version soon.

mbrandenburger (approved):

PR #1543, opened 2026-06-18
feat: add TLS support for PostgreSQL connections

### LINKED ISSUE #1542: Feature: Add TLS support for PostgreSQL connections
### Problem Description

Currently, the PostgreSQL storage driver in the Fabric Smart Client (FSC) does not natively support advanced TLS (Transport Layer Security) configurations. 

This feature was originally proposed and implemented in the Fabric Token SDK (see hyperledger-labs/fabric-token-sdk#1771). However, after discussing with the maintainers (@adecaro, @mbrandenburger, and @AkramBitar), we agreed that the optimal approach is to implement this capability directly at the FSC layer. By doing so, the Token SDK (and any other dependent projects) will inherit the TLS support "for free".

### **Why is this needed?**
- **Security Compliance:** Encrypted communication with PostgreSQL is a strict requirement for many enterprise deployments.
- **Centralization:** Handling this at the FSC level prevents code duplication across higher-level SDKs.

### **Proposed Solution**
Implement full TLS support for the PostgreSQL storage driver within FSC. The implementation should include:
- **`tlsConfigProvider`:** A mechanism to parse TLS options (e.g., from `fsc.persistences.<name>.opts.tls`) and decorate the existing DB provider.
- **SSL Modes:** Support for standard Postgres SSL modes (`disable`, `require`, `verify-ca`, `verify-full`).
- **Certificate Handling:** Automatic loading of CA certificates, client certificates, and client keys when the `tls` block is present in the configuration.
- **Backward Compatibility:** Existing configurations without a `tls` block should continue to work seamlessly via fallback logic.

### **Next Steps**
I will be opening a Pull Request shortly with the proposed implementation ported over from my previous work in the Token SDK.

SurbhiAgarwal1:
### Description
This PR introduces full TLS (Transport Layer Security) support for the PostgreSQL storage driver in the Fabric Smart Client.

As discussed with @adecaro, @mbrandenburger, and @AkramBitar in hyperledger-labs/fabric-token-sdk#1771, adding TLS support directly to FSC is the preferred approach. This centralization ensures that any dependent project, including the Fabric Token SDK, inherits this crucial security capability "for free."

#### What I Added
- **`tls.go`**: Implements `TLSConfig`, certificate loading, validation of standard PostgreSQL `sslmode` strings, and custom connection string registration using `pgx/v5/stdlib`.
- **`config.go` Updates**: Updated the core `ConfigProvider` to seamlessly parse `tls` blocks from `fsc.persistences.<name>.opts.tls`. If enabled, the configuration generates and registers a customized TLS connection string transparently.
- **`tls_test.go`**: A comprehensive unit test suite covering:
  - All supported `sslmode` values (`disable`, `require`, `verify-full`, `verify-ca`).
  - Missing certificate files and invalid key paths handling.
  - Proper fallback logic for when a persistence-specific TLS block is absent.
  - Self-signed certificate generation to ensure robust test isolation.

#### How it Works
The configuration parser reads TLS options on initialization. If TLS is enabled, the code generates a customized `pgx.ConnConfig` mapping the `sslmode` strictly to Go's standard `crypto/tls` behaviors. It then registers this custom config to the `stdlib` driver with a dynamically generated name. The system then connects using this secure connection, without fundamentally altering how FSC accesses the underlying DB. 

#### Verification
Running the driver test suite:
`go test -v ./platform/view/services/storage/driver/sql/postgres -run "TestTLS|TestRegisterTLS"`
All tests pass cleanly.

#### Checklist
- [x] Code builds cleanly (`go build ./...`)
- [x] All unit tests pass
- [x] Documentation comments added for new exported symbols
- [x] DCO Signed-off

Closes #1542

### DIFF
--- a/platform/view/services/storage/driver/common/config.go
+++ b/platform/view/services/storage/driver/common/config.go
@@ -60,11 +60,3 @@ func (c *Config) unmarshalPersistenceKey(name driver2.PersistenceName, key strin
 	}
 	return c.configProvider.UnmarshalKey(fmt.Sprintf("fsc.persistences.%s.%s", name, key), v)
 }
-
-func (c *Config) IsSet(key string) bool {
-	return c.configProvider.IsSet(key)
-}
-
-func (c *Config) UnmarshalKey(key string, rawVal any) error {
-	return c.configProvider.UnmarshalKey(key, rawVal)
-}
--- a/platform/view/services/storage/driver/common/config.go
+++ b/platform/view/services/storage/driver/common/config.go
@@ -60,3 +60,11 @@ func (c *Config) unmarshalPersistenceKey(name driver2.PersistenceName, key strin
 	}
 	return c.configProvider.UnmarshalKey(fmt.Sprintf("fsc.persistences.%s.%s", name, key), v)
 }
+
+func (c *Config) IsSet(key string) bool {
+	return c.configProvider.IsSet(key)
+}
+
+func (c *Config) UnmarshalKey(key string, rawVal any) error {
+	return c.configProvider.UnmarshalKey(key, rawVal)
+}
--- a/platform/view/services/storage/driver/sql/postgres/config.go
+++ b/platform/view/services/storage/driver/sql/postgres/config.go
@@ -64,7 +64,7 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string) 
 	if o.TLSConfig.Enabled {
 		registeredConnStr, err := RegisterTLSConnection(o.DataSource, *o.TLSConfig)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to register TLS connection config")
+			return nil, errors.Wrap(err, "failed to register TLS connection config")
 		}
 		o.DataSource = registeredConnStr
 	}
--- a/platform/view/services/storage/driver/sql/postgres/config.go
+++ b/platform/view/services/storage/driver/sql/postgres/config.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package postgres
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
@@ -19,8 +18,6 @@ import (
 // config models the DB configuration
 type config interface {
 	UnmarshalDriverOpts(name driver.PersistenceName, v any) error
-	IsSet(key string) bool
-	UnmarshalKey(key string, rawVal interface{}) error
 }
 
 type Config struct {
@@ -32,6 +29,7 @@ type Config struct {
 	SkipCreateTable bool
 	TableNameParams []string
 	Tracing         *common2.TracingConfig
+	TLSConfig       *TLSConfig
 }
 
 func NewConfigProvider(config config) *ConfigProvider {
@@ -59,25 +57,14 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string) 
 	o.TableNameParams = params
 	o.Tracing = &common2.TracingConfig{}
 
-	var tlsConfig TLSConfig
-	tlsKey := fmt.Sprintf("fsc.persistences.%s.opts.tls", name)
-	if r.config.IsSet(tlsKey) {
-		if err := r.config.UnmarshalKey(tlsKey, &tlsConfig); err != nil {
-			return nil, fmt.Errorf("failed to unmarshal database TLS config: %w", err)
-		}
-	} else {
-		defaultTlsKey := "fsc.persistences.default.opts.tls"
-		if r.config.IsSet(defaultTlsKey) {
-			if err := r.config.UnmarshalKey(defaultTlsKey, &tlsConfig); err != nil {
-				return nil, fmt.Errorf("failed to unmarshal default database TLS config: %w", err)
-			}
-		}
+	if o.TLSConfig == nil {
+		o.TLSConfig = &TLSConfig{}
 	}
 
-	if tlsConfig.Enabled {
-		registeredConnStr, err := RegisterTLSConnection(o.DataSource, tlsConfig)
+	if o.TLSConfig.Enabled {
+		registeredConnStr, err := RegisterTLSConnection(o.DataSource, *o.TLSConfig)
 		if err != nil {
-			return nil, fmt.Errorf("failed to register TLS connection config: %w", err)
+			return nil, errors.Wrapf(err, "failed to register TLS connection config")
 		}
 		o.DataSource = registeredConnStr
 	}
--- a/platform/view/services/storage/driver/sql/postgres/config.go
+++ b/platform/view/services/storage/driver/sql/postgres/config.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package postgres
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
@@ -18,6 +19,8 @@ import (
 // config models the DB configuration
 type config interface {
 	UnmarshalDriverOpts(name driver.PersistenceName, v any) error
+	IsSet(key string) bool
+	UnmarshalKey(key string, rawVal interface{}) error
 }
 
 type Config struct {
@@ -55,5 +58,29 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string) 
 	}
 	o.TableNameParams = params
 	o.Tracing = &common2.TracingConfig{}
+
+	var tlsConfig TLSConfig
+	tlsKey := fmt.Sprintf("fsc.persistences.%s.opts.tls", name)
+	if r.config.IsSet(tlsKey) {
+		if err := r.config.UnmarshalKey(tlsKey, &tlsConfig); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal database TLS config: %w", err)
+		}
+	} else {
+		defaultTlsKey := "fsc.persistences.default.opts.tls"
+		if r.config.IsSet(defaultTlsKey) {
+			if err := r.config.UnmarshalKey(defaultTlsKey, &tlsConfig); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal default database TLS config: %w", err)
+			}
+		}
+	}
+
+	if tlsConfig.Enabled {
+		registeredConnStr, err := RegisterTLSConnection(o.DataSource, tlsConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to register TLS connection config: %w", err)
+		}
+		o.DataSource = registeredConnStr
+	}
+
 	return o, nil
 }
--- a/platform/view/services/storage/driver/sql/postgres/config.go
+++ b/platform/view/services/storage/driver/sql/postgres/config.go
@@ -29,7 +29,7 @@ type Config struct {
 	SkipCreateTable bool
 	TableNameParams []string
 	Tracing         *common2.TracingConfig
-	TLSConfig       *TLSConfig
+	TLSConfig       *TLSConfig `yaml:"tls" mapstructure:"tls"`
 }
 
 func NewConfigProvider(config config) *ConfigProvider {
@@ -57,11 +57,7 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string) 
 	o.TableNameParams = params
 	o.Tracing = &common2.TracingConfig{}
 
-	if o.TLSConfig == nil {
-		o.TLSConfig = &TLSConfig{}
-	}
-
-	if o.TLSConfig.Enabled {
+	if o.TLSConfig != nil && o.TLSConfig.Enabled {
 		registeredConnStr, err := RegisterTLSConnection(o.DataSource, *o.TLSConfig)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to register TLS connection config")
--- a/platform/view/services/storage/driver/sql/postgres/tls.go
+++ b/platform/view/services/storage/driver/sql/postgres/tls.go
@@ -28,7 +28,6 @@ type TLSConfig struct {
 	SSLMode      string `json:"ssl_mode"       mapstructure:"ssl_mode"       yaml:"ssl_mode"`
 }
 
-
 // createTLSConnConfig parses the datasource string and configures standard Go TLS.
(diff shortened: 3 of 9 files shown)


platform/view/services/storage/driver/sql/postgres/config.go
```diff
@@ -18,6 +19,8 @@ import (
 // config models the DB configuration
 type config interface {
 	UnmarshalDriverOpts(name driver.PersistenceName, v any) error
+	IsSet(key string) bool
+	UnmarshalKey(key string, rawVal interface{}) error
```
mbrandenburger: I believe this is not needed. What we could do instead is to just add `TLSConfig *TLSConfig` to the `Config`. This should "unmarshal" into the TLSConfig struct automatically if present in the config.

platform/view/services/storage/driver/common/config.go
```diff
@@ -60,3 +60,11 @@ func (c *Config) unmarshalPersistenceKey(name driver2.PersistenceName, key strin
...
 	return c.configProvider.UnmarshalKey(fmt.Sprintf("fsc.persistences.%s.%s", name, key), v)
 }
+
+func (c *Config) IsSet(key string) bool {
+	return c.configProvider.IsSet(key)
+}
+
+func (c *Config) UnmarshalKey(key string, rawVal any) error {
+	return c.configProvider.UnmarshalKey(key, rawVal)
+}
```
mbrandenburger: I believe this is not needed. See my other comment.

platform/view/services/storage/driver/sql/postgres/config.go
```diff
@@ -55,5 +58,29 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string)
...
+			return nil, fmt.Errorf("failed to unmarshal database TLS config: %w", err)
+		}
+	} else {
+		defaultTlsKey := "fsc.persistences.default.opts.tls"
+		if r.config.IsSet(defaultTlsKey) {
+			if err := r.config.UnmarshalKey(defaultTlsKey, &tlsConfig); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal default database TLS config: %w", err)
+			}
+		}
+	}
```
mbrandenburger: Could we replace this with something like ...

```go
if o.TLSConfig == nil {
  o.TLSConfig = &TLSConfig{}
}
```

platform/view/services/storage/driver/sql/postgres/config.go
```diff
@@ -55,5 +58,29 @@ func (r *ConfigProvider) GetOpts(name driver.PersistenceName, params ...string)
...
+
+	if tlsConfig.Enabled {
+		registeredConnStr, err := RegisterTLSConnection(o.DataSource, tlsConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to register TLS connection config: %w", err)
```
mbrandenburger: I suggest to use `errors.Wrapf` here using our `github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors` package

platform/view/services/storage/driver/sql/postgres/tls.go
```diff
@@ -0,0 +1,122 @@
...
+// TLSConfig defines the configuration parameters for securing database connections.
+type TLSConfig struct {
+	Enabled      bool   `json:"enabled"        mapstructure:"enabled"        yaml:"enabled"`
+	ServerName   string `json:"server_name"    mapstructure:"server_name"    yaml:"server_name"`
+	CertPath     string `json:"cert_path"      mapstructure:"cert_path"      yaml:"cert_path"`
+	KeyPath      string `json:"key_path"       mapstructure:"key_path"       yaml:"key_path"`
+	ClientCACert string `json:"client_ca_cert" mapstructure:"client_ca_cert" yaml:"client_ca_cert"`
+	RootCertPath string `json:"root_cert_path" mapstructure:"root_cert_path" yaml:"root_cert_path"`
+	SSLMode      string `json:"ssl_mode"       mapstructure:"ssl_mode"       yaml:"ssl_mode"`
+}
```
mbrandenburger, line 30: Let's also make sure that we document the TLSConfig in https://github.com/hyperledger-labs/fabric-smart-client/blob/main/docs/configuration.md

platform/view/services/storage/driver/sql/postgres/tls.go
```diff
@@ -0,0 +1,122 @@
...
+
+	sslMode := tlsCfg.SSLMode
+	switch sslMode {
+	case "disable":
+		connConfig.TLSConfig = nil
+	case "allow", "prefer":
+		connConfig.TLSConfig = tlsConfig
+	case "require":
+		// 'require' mode enforces TLS handshake; verification follows default behavior
+		connConfig.TLSConfig = tlsConfig
```
mbrandenburger, line 93: If I am not wrong, "require" in postgres world means to encrypt the connection, but do not verify the server certificate. Thus, I believe we need to set `InsecureSkipVerify = true`

platform/view/services/storage/driver/sql/postgres/tls.go
```diff
@@ -0,0 +1,122 @@
...
+				opts.Intermediates = x509.NewCertPool()
+				for _, cert := range cs.PeerCertificates[1:] {
+					opts.Intermediates.AddCert(cert)
+				}
+			}
+			_, err := cs.PeerCertificates[0].Verify(opts)
+
+			return err
+		}
+		connConfig.TLSConfig = tlsConfig
```
mbrandenburger, line 100: I believe in postgres world "verify-ca" means: verify the CA chain but not the hostname. Similar as above, I believe we need to set `tlsConfig.InsecureSkipVerify = true` here too.

platform/view/services/storage/driver/sql/postgres/tls.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		tlsConfig.Certificates = []tls.Certificate{cert}
+	}
+
+	sslMode := tlsCfg.SSLMode
+	switch sslMode {
+	case "disable":
+		connConfig.TLSConfig = nil
+	case "allow", "prefer":
+		connConfig.TLSConfig = tlsConfig
```
mbrandenburger: this seems to be the same as `require`. I suggest that we just skip `allow/prefer` as it's not really clear if payload is encrypted or not ... see https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS

Thus, we should focus just on ...
- disable
- require
- verify-ca
- verify-full

Does this make sense to you?

platform/view/services/storage/driver/sql/postgres/config.go
mbrandenburger: Please check the PR for consistent use of our errors package.

platform/view/services/storage/driver/sql/postgres/tls.go
```diff
@@ -0,0 +1,122 @@
...
+type TLSConfig struct {
+	Enabled      bool   `json:"enabled"        mapstructure:"enabled"        yaml:"enabled"`
+	ServerName   string `json:"server_name"    mapstructure:"server_name"    yaml:"server_name"`
+	CertPath     string `json:"cert_path"      mapstructure:"cert_path"      yaml:"cert_path"`
+	KeyPath      string `json:"key_path"       mapstructure:"key_path"       yaml:"key_path"`
+	ClientCACert string `json:"client_ca_cert" mapstructure:"client_ca_cert" yaml:"client_ca_cert"`
+	RootCertPath string `json:"root_cert_path" mapstructure:"root_cert_path" yaml:"root_cert_path"`
+	SSLMode      string `json:"ssl_mode"       mapstructure:"ssl_mode"       yaml:"ssl_mode"`
+}
+
```
mbrandenburger, line 31: only a single blank line here.

platform/view/services/storage/driver/sql/postgres/tls_test.go
```diff
@@ -0,0 +1,268 @@
...
+				t.Helper()
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), "failed to load client key pair")
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			connConfig, err := createTLSConnConfig(tt.dataSource, tt.tlsCfg)
```
mbrandenburger, line 267: The test name suggest to test `RegisterTLSConnection` ... but instead we test `createTLSConnConfig`. This is misleading. I suggest we rename the test.

platform/view/services/storage/driver/sql/postgres/tls_test.go
```diff
@@ -0,0 +1,268 @@
...
+		dataSource string
+		tlsCfg     TLSConfig
+		verify     func(t *testing.T, connConfig *pgx.ConnConfig, err error)
+	}{
+		{
+			name:       "SSLMode disable",
+			dataSource: "host=localhost port=5432 user=postgres dbname=test",
+			tlsCfg: TLSConfig{
+				SSLMode: "disable",
+			},
```
mbrandenburger, line 121: we should also have a test case where tlsCfg is nil.

platform/view/services/storage/driver/sql/postgres/tls_test.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (m *mockConfig) UnmarshalDriverOpts(name driver2.PersistenceName, v any) error {
+	if baseOpts, ok := m.values["baseOpts"]; ok {
+		return mapstructure.Decode(baseOpts, v)
+	}
+	return nil
+}
+
+func generateSelfSignedCert(t *testing.T, tempDir string) (string, string) {
```
mbrandenburger, line 70: I am wondering if we could reuse code we already have to generate the tls crypto material. For instance, here we have `platform/view/services/grpc/tlsgen/ca.go`. While importing this is code as is is not ideal - we could refactor the package to make it better "shareable". That could also happen in a follow up PR. WDYT?

platform/view/services/storage/driver/sql/postgres/tls_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	privBytes, err := x509.MarshalPKCS8PrivateKey(privateKey)
+	require.NoError(t, err)
+	err = pem.Encode(keyOut, &pem.Block{Type: "PRIVATE KEY", Bytes: privBytes})
+	require.NoError(t, err)
+
+	return certPath, keyPath
+}
+
+func TestRegisterTLSConnection(t *testing.T) {
```
mbrandenburger: In addition to these unit tests, we should also run a test with a "real" postgres instance.

I suggest to add a test, i.e., `TestPostgresWithTLS` following the the scheme of `TestPostgres` in  `platform/view/services/storage/driver/sql/postgres/sql_test.go`. Probably this would require to extend the configuration of that postgres test container for TLS setup. 

Moreover, we should modify one of our integration tests to use postgres with TLS. Maybe the `integration/fsc/stoprestart` test would be a good candidate. WDYT?

mbrandenburger (changes requested):
Thank you @SurbhiAgarwal1 for this great work! Adding TLS support for postgres is great step forward for FSC. I have a few suggestions below on how we could improve this implementation. Please have a look.

SurbhiAgarwal1:
Hi @mbrandenburger, thanks for the suggestion! I'll update this soon.

mbrandenburger (changes requested):
Thank you @SurbhiAgarwal1 for the updates. Please have a look at the code formatting. You can run `make checks` to see if there are any issues with the formatting before pushing the commit.

SurbhiAgarwal1:
Hi @mbrandenburger, thanks for pointing that out! I've fixed the code formatting issues and pushed the updates. The checks should pass now.

mbrandenburger:
> The checks should pass now.
Not yet :(

mbrandenburger:
@SurbhiAgarwal1 did you manage to run `make checks` without any errors on your machine?

SurbhiAgarwal1:
@mbrandenburger I ran into some environment issues running the make checks target directly on my Windows machine, but I manually ran the underlying tools (gofumpt, gci, and go vet) that make checks relies on.

Those tools caught the errors.Wrapf static analysis warning as well as the formatting issues (like import ordering and octal file modes). I've fixed those up and verified that the linters pass cleanly locally now. The latest commit I just pushed should turn the CI green!

mbrandenburger:
> I've fixed those up and verified that the linters pass cleanly locally now. The latest commit I just pushed should turn the CI green!

Great progress! I can see that the number of issues reported by the linter is significantly lower. 

Next, we just need to fix the `paralleltest` issue in `platform/view/services/storage/driver/sql/postgres/tls_test.go`. If possible add `t.Parallel()` to the tests, otherwise use the `func TestName(t *testing.T) { //nolint:paralleltest`.

mbrandenburger:
Yeah! The linter is happy. Next, let's check why one of the postgres unit test fails.

If you run `make unit-tests-postgres` you can run all postgres tests locally.

SurbhiAgarwal1:
Hi @mbrandenburger , I have pushed a new commit to address the failing utest-postgres CI checks.

- **Fixed the script line endings**: The init-ssl.sh script was failing in the Linux container because of Windows \r\n line endings. Added a quick strings.ReplaceAll in Go to force Unix line endings so it runs cleanly everywhere.
- **Fixed the SSL startup crash**: We were passing -c ssl=on via the Docker entrypoint, which crashed the temporary initdb server since the certs hadn't been copied yet. I moved the SSL config directly into postgresql.conf during the init script instead.
- **Linter fix**: Added the missing tt := tt capture in TestCreateTLSConnConfig to make the paralleltest linter happy.

Everything is passing locally now on my end! Let me know if you need any other changes.

Screenshot 2026-07-16 010252

mbrandenburger:
> Linter fix: Added the missing tt := tt capture in TestCreateTLSConnConfig to make the paralleltest linter happy.

I believe this should not be necessary with our "modern" go version. I am wondering if you have the linter version installed which is referenced here in the FSC repo. Note that from time to time we update the tools and you may want to run `make install-tools` to update on your environment.

mbrandenburger:
Great effort @SurbhiAgarwal1 ... all tests are passing! I will next do another review. Thank you again.

SurbhiAgarwal1:
Hi @mbrandenburger , Can you reopen this issue?

SurbhiAgarwal1:
Hi @mbrandenburger, thank you for reopening the PR! 

All CI checks are green. Whenever you have a chance, could you please take a look for your final review and update the review status? Thanks again for your guidance!

mbrandenburger (approved):

PR #1545, opened 2026-06-25
bug: idemix cache  #1544

### LINKED ISSUE #1544: bug: idemix cache
Currently, the idemix cache in `platform/fabric/core/generic/msp/idemix/cache.go` spawn NumCPU go routines even if the required cache level is lower than the number of available cores. This induced an excessive generation of pseudonyms. 

This Issue asks to fix this problem suggesting to spawn just a single go routine.

adecaro:
This PR fixes the bug reported by #1544 

Closes #1544

### DIFF
--- a/platform/fabric/core/generic/msp/idemix/cache.go
+++ b/platform/fabric/core/generic/msp/idemix/cache.go
@@ -7,7 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package idemix
 
 import (
-	"runtime"
+	"context"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


platform/fabric/core/generic/msp/idemix/cache.go
```diff
@@ -49,10 +54,9 @@ func (c *IdentityCache) Identity(opts *driver.IdentityOptions) (view.Identity, [
 
 	c.once.Do(func() {
 		if cap(c.cache) > 0 {
-			// Spin up as many background goroutines as we need to prepare identities in the background.
-			for i := 0; i < runtime.NumCPU(); i++ {
-				go c.provisionIdentities()
-			}
+			var backgroundCtx context.Context
+			backgroundCtx, c.cancel = context.WithCancel(context.Background())
+			go c.provisionIdentities(backgroundCtx)
```
mbrandenburger, line 83: This is a good step forward. However, the currently never call `cancel`. 

Should we introduce something like `Close()` to allow discarding the go routine? 

```go
// Missing — needed to pair with context.WithCancel
func (c *IdentityCache) Close() {
    if c.cancel != nil {
        c.cancel()
    }
}
```

Though I am not yet sure what would be the best strategy to call `Close()` when the identity cache is not needed anymore. WDYT?

platform/fabric/core/generic/msp/idemix/cache.go
```diff
@@ -30,13 +30,18 @@ type IdentityCache struct {
...
 
 func NewIdentityCache(backed IdentityCacheBackendFunc, size int, opts *driver.IdentityOptions) *IdentityCache {
 	ci := &IdentityCache{
-		backed: backed,
-		cache:  make(chan identityCacheEntry, size),
-		opts:   opts,
+		backed:       backed,
+		cache:        make(chan identityCacheEntry, size),
+		opts:         opts,
+		cacheTimeout: 5 * time.Millisecond,
```
mbrandenburger: previously we had a hardcoded timeout of 1 second. 5 millisecond seems to be very aggressive. Probably a good idea to let the user define the timeout via some `IdentityCacheOptions`?

platform/fabric/core/generic/msp/idemix/cache_test.go
```diff
@@ -7,28 +7,44 @@ SPDX-License-Identifier: Apache-2.0
...
 	require.Equal(t, []byte("audit"), audit)
+
+	// fetch from cache
+	id, audit, err = c.Identity(nil)
+	require.NoError(t, err)
+	require.Equal(t, view.Identity([]byte("hello world")), id)
+	require.Equal(t, []byte("audit"), audit)
+
+	time.Sleep(100 * time.Millisecond)
+	require.Equal(t, int32(104), counter.Load())
```
mbrandenburger: this might be flaky on a loaded CI run. Using `Eventually` might be safer option here. WDYT?

mbrandenburger (changes requested):
Thank you @adecaro for creating this bug fix. Please see my comments below.

adecaro:
Hi @mbrandenburger , thanks for the review. I'll address your comments.

platform/fabric/core/generic/msp/idemix/cache.go
adecaro: good idea. I would still keep the value to 5 milliseconds. This timeout says: It an entry is not ready in cache in the next 5 milliseconds, I'll generate one on the spot. It is a way to reduce latency as much as possible.

platform/fabric/core/generic/msp/idemix/cache.go
adecaro: make sense, will do

mbrandenburger (approved):

PR #1547, opened 2026-06-26
feat(fabric/config): make PKCS11 SessionCacheSize configurable

### LINKED ISSUE #1546: feat(fabric/config): PKCS11 session cache size is not configurable (pinned to fabric-lib-go default of 10)
### Description

The Fabric MSP PKCS11 config (`platform/fabric/core/generic/config/ds.go`, type `PKCS11`) exposes the standard PKCS11 options (Library, Label, Pin, KeyIDs, ...) but not the session cache size. The x509 `ToPKCS11OptsOpts` converter (`platform/fabric/core/generic/msp/x509/config.go`) builds a `pkcs11.PKCS11Opts` without setting it, so the PKCS11 session pool is always pinned to fabric-lib-go's hardcoded default of 10 (`defaultSessionCacheSize` in `bccsp/pkcs11`).

On a node that drives a large, concurrent PKCS11/HSM signing workload, 10 cached sessions cap the pool and serialize signing under load — there is no way to raise it from configuration.

The currently pinned fabric-lib-go v1.1.3 keeps `sessionCacheSize` unexported, so even adding a config field would have no API to set it. v1.1.4 exposes `SessionCacheSize` on `bccsp/pkcs11.PKCS11Opts` (#49), making it settable.

### Expected

The PKCS11 session cache size is configurable from the MSP PKCS11 config, falling back to the existing default (10) when unset.

### Proposed fix

Bump fabric-lib-go to v1.1.4 and plumb a new `SessionCacheSize` field from `config.PKCS11` into `pkcs11.PKCS11Opts` via the x509 `ToPKCS11OptsOpts` converters. Add the field to the `!pkcs11` build-tag stub as well so the default build compiles. Omitting it keeps current behavior (zero → fabric-lib-go's default of 10).

### Version

`main`

EvanYan1024:
Fixes #1546.

The Fabric MSP PKCS11 config exposed the standard PKCS11 options but not the session cache size, so the x509 `ToPKCS11OptsOpts` converter always left it at fabric-lib-go's hardcoded default of 10. On nodes that drive a large, concurrent PKCS11/HSM signing workload, that pins the session pool and serializes signing under load, with no config knob to raise it.

fabric-lib-go v1.1.4 exposes `SessionCacheSize` on `bccsp/pkcs11.PKCS11Opts` (#49), so this:

- adds a `SessionCacheSize` field to `config.PKCS11` and to the `!pkcs11` build-tag stub so both build variants compile;
- forwards it into `PKCS11Opts` in the x509 `ToPKCS11OptsOpts` converters (the package-level call site in `config.go` and the `pkcs11`-tag sub-package converter);
- forwards it in the NWO integration test-network converter as well, so configuring the field is not silently dropped when spinning up test networks.

Because FSC is multi-module, fabric-lib-go is bumped v1.1.3 → v1.1.4 in each module that pins it (root, `integration`, and `comm/host/libp2p`) to keep the version consistent across the repo. v1.1.4 also carries its own lib bumps (its #51), which MVS-resolves a few indirect deps forward (miekg/pkcs11, mapstructure, x/mod, x/tools, x/telemetry).

Backward compatible: omitting the field leaves it zero, which fabric-lib-go maps to its default cache size of 10. Verified `go build ./...` (default) for each module, plus `go build -tags pkcs11` on the affected x509 and NWO pkcs11 packages.

Commits are split per module: the root x509 feature, the NWO integration wiring, and a libp2p version-alignment chore (that module does not use the PKCS11 change).

Closes #1546


mbrandenburger (changes requested):
Thank you @EvanYan1024 for working on this. This is a good step forward. I think the only thing we are missing is some documentation, for instance in `docs/configuration.md` as part of the `core.yaml` description. WDYT?

EvanYan1024:
Thanks, @mbrandenburger! Added the documentation in `docs/configuration.md` under the PKCS11 section (8d13fba7).

mbrandenburger (approved):
Thank you!

PR #1551, opened 2026-07-01
feat: tag release script

### LINKED ISSUE #1550: multi module tag automation
### Description

With the new multi-module structure we need a better way to generate git tags as a single tag only refers to the main module.

It would be nice to have a script that generates git tags for all go modules (except tools) for a given tag name.

### Implementation Steps

- write bash script to create git tags for all go modules (except tools)
- add make target

### Additional Information

_No response_

mbrandenburger:
Closes #1550


adecaro (approved):
LGTM

PR #1553, opened 2026-07-02
perf(fabricx/queryservice): add batched GetTransactionStatuses

### LINKED ISSUE #1552: fabricx queryservice only exposes per-tx status queries although the committer API supports batching
## Describe the issue

`RemoteQueryService.GetTransactionStatus` resolves one transaction per committer round-trip, even though the underlying request message already supports batching — the implementation sends `committerpb.TxStatusQuery{TxIds: []string{txID}}`, where `TxIds` is a repeated field.

Any component that tracks finality of many in-flight transactions (e.g. a finality poller) is forced into one gRPC round-trip per transaction. In our deployment, polling tens of thousands of pending transactions this way capped finality settlement at ~92 tx/s while workers sat blocked in `GetTransactionStatus`, far below what the committer itself sustains.

## Proposed fix

Expose the batching the committer API already provides:

```go
// GetTransactionStatuses resolves many transaction statuses in one batched query,
// keyed by transaction ID; transactions unknown to the committer are omitted.
GetTransactionStatuses(txIDs []string) (map[string]int32, error)
```

Keying the result by `TxStatus.Ref.TxId` keeps the caller independent of response ordering. No committer-side change is required.

I have the change ready (method + `QueryService` interface + regenerated mocks + unit tests) and will submit a PR.

## Who consumes this

- **Immediate consumer:** the Fabric Token SDK's fabricx finality layer. In our deployment we replaced its per-transaction status queries with a batch poller built on this method to track finality of tens of thousands of in-flight transactions; we intend to propose that poller upstream to the token SDK (panurus), where it would consume this API through the query service.
- **In-tree:** today `fabricx/core/vault.Status` and `fabricx/core/ledger.GetTransactionByID` resolve statuses one transaction at a time against the query service. The batched variant gives any component that needs the status of many transactions (finality sweeps, recovery) an efficient primitive without changing those existing call sites.

EvanYan1024:
Closes #1552

## What

Adds `GetTransactionStatuses(txIDs []string) (map[string]int32, error)` to the fabricx query service: many transaction statuses resolved in one committer round-trip.

- `RemoteQueryService` implementation — a single `TxStatusQuery{TxIds: txIDs}` call; the committer request message is already `repeated`, so no committer-side change is needed.
- Result keyed by `TxStatus.Ref.TxId`, independent of response ordering. Transactions unknown to the committer are omitted from the map (callers treat a missing entry as "not final yet"). Statuses without a `Ref` are dropped. Empty input short-circuits without a network call.
- Added to the `QueryService` interface; counterfeiter mocks regenerated (`queryservice/mock`, `ledger/mock`); the hand-written `mockQueryService` in vault tests extended.
- Unit tests: out-of-order/partial responses keyed correctly, ref-less statuses dropped, empty-input short-circuit, client error propagation.

## Why

The query service only exposed a per-tx call, forcing finality tracking of many in-flight transactions into one gRPC round-trip each. In our deployment this capped finality settlement at ~92 tx/s with tens of thousands of pending transactions, while pollers sat blocked in `GetTransactionStatus` — far below what the committer sustains.

## Notes for reviewers

- A single `RequestTimeout` covers the whole batch; callers issuing very large batches should chunk on their side.
- `go build ./platform/fabricx/...`, `go vet`, and the full `./platform/fabricx/...` test suite pass.

### DIFF
--- a/platform/fabricx/core/committer/queryservice/query.go
+++ b/platform/fabricx/core/committer/queryservice/query.go
@@ -92,21 +92,34 @@ func (s *RemoteQueryService) GetTransactionStatus(txID string) (int32, error) {
 }
 
 // GetTransactionStatuses resolves the status of many transactions in one query,
-// keyed by transaction ID; transactions unknown to the committer are omitted.
+// keyed by transaction ID. Duplicate input IDs are collapsed before querying.
+// Unlike GetTransactionStatus, which returns an error for a transaction unknown
+// to the committer, unknown transactions are omitted from the result so callers
+// can treat a missing entry as "not final yet".
 func (s *RemoteQueryService) GetTransactionStatuses(txIDs []string) (map[string]int32, error) {
 	if len(txIDs) == 0 {
 		return map[string]int32{}, nil
+	}
+
+	seen := make(map[string]struct{}, len(txIDs))
+	unique := make([]string, 0, len(txIDs))
+	for _, txID := range txIDs {
+		if _, ok := seen[txID]; ok {
+			continue
+		}
+		seen[txID] = struct{}{}
+		unique = append(unique, txID)
 	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), s.config.RequestTimeout)
 	defer cancel()
 
 	now := time.Now()
-	res, err := s.client.GetTransactionStatus(ctx, &committerpb.TxStatusQuery{TxIds: txIDs})
+	res, err := s.client.GetTransactionStatus(ctx, &committerpb.TxStatusQuery{TxIds: unique})
 	if err != nil {
 		return nil, errors.Wrap(err, "query transaction statuses")
 	}
-	logger.Debugf("QS GetTransactionStatuses: %d ids -> %d statuses in %v", len(txIDs), len(res.GetStatuses()), time.Since(now))
+	logger.Debugf("QS GetTransactionStatuses: %d ids -> %d statuses in %v", len(unique), len(res.GetStatuses()), time.Since(now))
 
 	out := make(map[string]int32, len(res.GetStatuses()))
 	for _, st := range res.GetStatuses() {
--- a/platform/fabricx/core/committer/queryservice/query.go
+++ b/platform/fabricx/core/committer/queryservice/query.go
@@ -91,6 +91,32 @@ func (s *RemoteQueryService) GetTransactionStatus(txID string) (int32, error) {
 	return int32(res.Statuses[0].Status), nil
 }
 
+// GetTransactionStatuses resolves the status of many transactions in one query,
+// keyed by transaction ID; transactions unknown to the committer are omitted.
+func (s *RemoteQueryService) GetTransactionStatuses(txIDs []string) (map[string]int32, error) {
+	if len(txIDs) == 0 {
+		return map[string]int32{}, nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), s.config.RequestTimeout)
+	defer cancel()
+
+	now := time.Now()
+	res, err := s.client.GetTransactionStatus(ctx, &committerpb.TxStatusQuery{TxIds: txIDs})
+	if err != nil {
+		return nil, errors.Wrap(err, "query transaction statuses")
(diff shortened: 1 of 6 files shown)


platform/fabricx/core/committer/queryservice/query.go
```diff
@@ -91,6 +91,32 @@ func (s *RemoteQueryService) GetTransactionStatus(txID string) (int32, error) {
 	return int32(res.Statuses[0].Status), nil
 }
 
+// GetTransactionStatuses resolves the status of many transactions in one query,
+// keyed by transaction ID; transactions unknown to the committer are omitted.
+func (s *RemoteQueryService) GetTransactionStatuses(txIDs []string) (map[string]int32, error) {
```
mbrandenburger, line 99: While I think the behavior of this method makes sense, `GetTransactionStatus` returns an error for transactions unknown to the committer whereas here we omit them. So, both methods behave slightly different. I believe if we highlight this in the description better we are fine.

platform/fabricx/core/committer/queryservice/query.go
```diff
@@ -91,6 +91,32 @@ func (s *RemoteQueryService) GetTransactionStatus(txID string) (int32, error) {
...
+	if len(txIDs) == 0 {
+		return map[string]int32{}, nil
+	}
```
mbrandenburger, line 102: I am wondering if we should tackle txID input deduplication here as well. The committer should deal with this correctly, but avoiding sending duplicates seems to be a good idea.  A simple `map[string]struct{}` dedup would be negligible overhead compared to the gRPC call I believe. WDYT?

platform/fabricx/core/committer/queryservice/query.go
```diff
@@ -91,6 +91,32 @@ func (s *RemoteQueryService) GetTransactionStatus(txID string) (int32, error) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), s.config.RequestTimeout)
+	defer cancel()
```
mbrandenburger, line 115: I am bit concerned with the request timeout here. Batch requests that could cover thousands of IDs under load, could run into issues with a request timeout tweaked for single ID calls. Ideally all the queryservice APIs would receive a context provided by the user to deal with this, however, this problem is out-of-scope for this PR.

mbrandenburger (changes requested):
Thank you @EvanYan1024 for this solid PR. I left a few minor comments below.

adecaro:
@EvanYan1024 , wondering if we can use this function also in Panurus to batch queries for status. For finality, we use a mix of notifier and queries.

platform/fabricx/core/committer/queryservice/query.go
EvanYan1024: Good point — expanded the godoc to explicitly contrast with `GetTransactionStatus`'s error-on-unknown behavior and to state that a missing entry means "not final yet".

platform/fabricx/core/committer/queryservice/query.go
EvanYan1024: Agreed — added: duplicate IDs are now collapsed before the query (`map[string]struct{}`), with a test asserting the deduped list on the wire.

platform/fabricx/core/committer/queryservice/query.go
EvanYan1024: Agree with the analysis. Short term, callers issuing very large batches should chunk on their side (noted in the PR description) — that also bounds the blast radius of a timeout to one chunk. Longer term, plumbing a caller-provided context through the queryservice APIs is the right direction; as you said it is out of scope here — happy to open a follow-up issue to track it if useful.

EvanYan1024:
@adecaro yes — that's exactly the intent. In our deployment we replaced the token SDK's per-transaction finality status queries in the fabricx finality layer with a batch poller built on this method (working alongside the existing notifier), which removed the one-round-trip-per-transaction bottleneck. We plan to propose that batch poller to Panurus as a follow-up; it consumes this API through the query service.

mbrandenburger (approved):

PR #1557, opened 2026-07-06
bug(fabricx): skip CertificateID for idemix  #1556

### LINKED ISSUE #1556: bug(fabricx): skip CertificateID for idemix
The function `createSCEnvelope` in `platform/fabricx/core/transaction/marshal.go`, fails when the creator of the transaction is an idemix identity. More precisely, the failure happens here:

```go
	mspIdentity, err := toMSPSignerIdentityWithCertificateID(signerID)
	if err != nil {
		return nil, errors.Wrap(err, "converting creator to cached identity")
	}
```

This code assumes that signerID is an 509 identity. The solution is simple, just stkip idemix identities by looking at the type in `msp.SerializedIdentity`

adecaro:
This PR fixes #1556 in the following way: It enhance the channel membership service to be able to tell if an MSP is and Idemix-based MSP. If that is the case, then CertID generation will be skipped.

### DIFF
--- a/platform/fabricx/core/transaction/marshal.go
+++ b/platform/fabricx/core/transaction/marshal.go
@@ -61,7 +61,7 @@ func (t *Transaction) createSCEnvelope() (*cb.Envelope, error) {
 	if membership == nil {
 		return nil, errors.Errorf("no channel membership for channel [%s]", t.Channel())
 	}
-	creatorBytes, err := toMSPSignerIdentityWithCertificateID(signerID, ch.ChannelMembership().IsIdemixMSP)
+	creatorBytes, err := toMSPSignerIdentityWithCertificateID(signerID, membership.IsIdemixMSP)
 	if err != nil {
 		return nil, errors.Wrap(err, "converting creator to cached identity")
 	}
--- a/platform/fabricx/core/transaction/marshal.go
+++ b/platform/fabricx/core/transaction/marshal.go
@@ -52,13 +52,18 @@ func (t *Transaction) createSCEnvelope() (*cb.Envelope, error) {
 		return nil, errors.Wrapf(err, "signer not found for %s while creating tx envelope for ordering", signerID.UniqueID())
 	}
 
-	mspIdentity, err := toMSPSignerIdentityWithCertificateID(signerID)
+	ch, err := t.fns.Channel(t.Channel())
+	if err != nil {
+		return nil, errors.Wrapf(err, "get channel [%s]", t.Channel())
+	}
+
+	membership := ch.ChannelMembership()
+	if membership == nil {
+		return nil, errors.Errorf("no channel membership for channel [%s]", t.Channel())
+	}
+	creatorBytes, err := toMSPSignerIdentityWithCertificateID(signerID, ch.ChannelMembership().IsIdemixMSP)
 	if err != nil {
 		return nil, errors.Wrap(err, "converting creator to cached identity")
-	}
-	creatorBytes, err := proto.Marshal(mspIdentity)
-	if err != nil {
-		return nil, errors.Wrap(err, "marshalling creator identity")
 	}
 	signatureHeader := &cb.SignatureHeader{Creator: creatorBytes, Nonce: t.Nonce()}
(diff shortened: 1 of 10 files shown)


platform/fabricx/core/transaction/marshal.go
```diff
@@ -52,7 +52,12 @@ func (t *Transaction) createSCEnvelope() (*cb.Envelope, error) {
 		return nil, errors.Wrapf(err, "signer not found for %s while creating tx envelope for ordering", signerID.UniqueID())
 	}
 
-	mspIdentity, err := toMSPSignerIdentityWithCertificateID(signerID)
+	ch, err := t.fns.Channel(t.Channel())
+	if err != nil {
+		return nil, errors.Wrapf(err, "get channel [%s]", t.Channel())
+	}
```
mbrandenburger, line 58: Additionally, we should add a nil guard for `ch.ChannelMembership()` similar as you do in `transaction.go` below.

platform/fabricx/core/transaction/transaction_test.go
```diff
@@ -420,7 +420,9 @@ func TestToMSPSignerIdentityWithCertificateID(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
-			id, err := toMSPSignerIdentityWithCertificateID(tc.identity)
+			id, err := toMSPSignerIdentityWithCertificateID(tc.identity, func(mspID string) bool {
+				return false
+			})
```
mbrandenburger, line 433: we should add a test to cover the case where `idemixMSP=true`.

mbrandenburger (changes requested):
Thank you @adecaro for this fix. Good catch. This PR is in a very good shape. A few small things and we are ready to go.

platform/fabricx/core/transaction/marshal.go
adecaro: okay

platform/fabricx/core/transaction/transaction_test.go
adecaro: yes

platform/fabricx/core/transaction/marshal.go
```diff
@@ -52,13 +52,18 @@ func (t *Transaction) createSCEnvelope() (*cb.Envelope, error) {
...
-		return nil, errors.Wrap(err, "converting creator to cached identity")
+		return nil, errors.Wrapf(err, "get channel [%s]", t.Channel())
+	}
+
+	membership := ch.ChannelMembership()
+	if membership == nil {
+		return nil, errors.Errorf("no channel membership for channel [%s]", t.Channel())
 	}
-	creatorBytes, err := proto.Marshal(mspIdentity)
+	creatorBytes, err := toMSPSignerIdentityWithCertificateID(signerID, ch.ChannelMembership().IsIdemixMSP)
```
mbrandenburger:
```suggestion
	creatorBytes, err := toMSPSignerIdentityWithCertificateID(signerID, membership.IsIdemixMSP)
```

platform/fabricx/core/transaction/transaction.go
```diff
@@ -666,22 +666,62 @@ func (t *Transaction) getProposalResponse(signer SerializableSigner) (*pb.Propos
...
+	if err := proto.Unmarshal(identity, sID); err != nil {
+		return nil, errors.Wrap(err, "unmarshal serialized identity")
+	}
+	if idemixMSP(sID.GetMspid()) {
+		logger.Debugf("identity [%s] is an idemix identity, return it. [%s]", identity, sID.GetMspid())
+		return identity, nil
+	}
+
+	logger.Debugf("identity [%s] is NOT an idemix identity, return it. [%s]", identity, sID.GetMspid())
+
```
mbrandenburger:
```suggestion
```

I believe the debug statement is a bit misleading. I would just remove it.

mbrandenburger (changes requested):
Thank you @adecaro for working on this. Two little things below.

mbrandenburger (approved):

PR #1563, opened 2026-07-07
test(coverage): use -coverpkg so cross-package code is credited

### LINKED ISSUE #1564: Coverage reports 0% for code exercised only by tests in other packages
### Description

- Coverage reports (Coveralls) show 0% for source files that are actually well-covered, because they are only exercised by tests living in a different package.

- Concrete example: platform/view/services/storage/driver/sql/common/test_utils.go contains exported test helpers used by the sqlite and postgres driver test packages. It shows 0.0% in Coveralls despite ~50/54 statements being executed on every run.

- Root cause: Go's go test -cover attributes coverage per-package, a package's statements are only instrumented in its own test binary. When package common is exercised by sqlite's tests, those statements aren't counted, because common has no tests of its own. The coverage numbers reported to Coveralls are therefore inaccurate and understate real coverage across the repo. 

### Implementation Steps

- [x] Add `GO_COVERPKG ?= -coverpkg=./...` to the Makefile (separate from `GO_TEST_PARAMS`, which CI overrides).
- [x] Add `$(GO_COVERPKG)` to the `go test` line in `unit-tests`, `unit-tests-postgres`, and `coverage-local`.
- [x] Verify `.../sql/common/test_utils.go` now shows real coverage instead of 0%.
- [x] Confirm the repo-wide number changes as expected.

### Additional Information

The same pattern affects any shared helper or indirectly exercised package repo wide. Fixing it (via -coverpkg=./... in the unit-test Makefile targets) raises filtered full-repo coverage the higher number being the more accurate one.

SaidAltury-ibm:
Adds a `GO_COVERPKG=-coverpkg=./...` flag to the `go test` unit-test targets in the Makefile (`unit-tests`, `unit-tests-postgres`, `coverage-local`).

## Impact on the Coveralls number
full-repo coverage moves **77% → 81%** (+4 pts). This increase because the repo has a lot of cross package exercised code
closes #1564

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,18 +139,24 @@ GO_PACKAGES_SDK = $$(go list ./... | grep '/sdk/dig$$')
 GO_TEST_PARAMS ?= -race -cover
 TEST_PKGS ?= $(GO_PACKAGES)
 
+# Instrument every package under the root module so that coverage is credited
+# even when a package is exercised only by tests living in another package
+# (e.g. the exported helpers in .../sql/common used by the sqlite/postgres tests).
+# Kept separate from GO_TEST_PARAMS because CI overrides GO_TEST_PARAMS wholesale.
+GO_COVERPKG ?= -coverpkg=./...
+
 .PHONY: unit-tests
 unit-tests: ## Run unit tests
 	@echo "Running unit tests..."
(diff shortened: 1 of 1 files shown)


mbrandenburger (approved):
Thank you @SaidAltury-ibm

PR #1569, opened 2026-07-08
fix(fabric/db): open VaultStore.Store write transaction with BeginTx(ctx)

### LINKED ISSUE #1568: VaultStore.Store opens its write transaction with Begin(), so it is not tied to the request context
`VaultStore.Store` receives a `context.Context` and runs every statement with `tx.ExecContext(ctx, ...)`, but the enclosing write transaction is opened with a bare `Begin()`:

- https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/services/db/driver/sql/postgres/vault.go#L54
- https://github.com/hyperledger-labs/fabric-smart-client/blob/main/platform/fabric/services/db/driver/sql/sqlite/vault.go#L56

Because the transaction lifecycle is not tied to the context, a canceled or timed-out request does not roll the transaction back: `database/sql` only releases the connection when the surrounding code returns, so it can sit in `idle in transaction` state holding a slot in the bounded write pool. The read path already does this correctly — `common/vault.go` opens its read transactions with `BeginTx(ctx, ...)`.

Expected behavior: the write transaction follows the request context — `BeginTx(ctx, nil)` — so a cancellation rolls it back automatically and returns the connection to the pool.

Happy to send a PR: it is a one-line change in each of the two vault stores, plus declaring `BeginTx` on the `WriteDB` interface (both implementations already provide it through the embedded `*sql.DB`).

EvanYan1024:
Closes #1568

`VaultStore.Store` already receives the request context and runs every statement with `ExecContext`, but the enclosing write transaction was opened with a bare `Begin()`. A canceled or timed-out request therefore did not roll the transaction back, and the connection could sit in `idle in transaction` state holding a slot in the bounded write pool. The read path (`common/vault.go`) already ties its transactions to the context with `BeginTx`.

This opens the write transaction with `BeginTx(ctx, nil)` in the postgres and sqlite vault stores, and declares `BeginTx` on the `WriteDB` interface — both implementations (`*sql.DB` and sqlite's `retryWriteDB`, which embeds `*sql.DB`) already provide it, so no implementation changes are needed. Callers passing `context.Background()` see no behavior change.

We run the postgres store in our test environments under sustained load; tying write transactions to the request context is standard `database/sql` hygiene that prevents canceled requests from pinning pool connections.

### Testing

`go build ./...`, `go vet` and `go test -race` on `platform/fabric/services/db/...` and `platform/view/services/storage/driver/sql/...` pass locally (postgres testcontainer suites require Docker and run in CI). `golangci-lint run` with the repo config reports no issues.

### DIFF
--- a/platform/fabric/services/db/driver/sql/postgres/vault.go
+++ b/platform/fabric/services/db/driver/sql/postgres/vault.go
@@ -51,7 +51,7 @@ func (db *VaultStore) Store(ctx context.Context, txIDs []driver.TxID, writes dri
 	db.GlobalLock.RLock()
 	defer db.GlobalLock.RUnlock()
 
-	tx, err := db.writeDB.Begin()
+	tx, err := db.writeDB.BeginTx(ctx, nil)
 	if err != nil {
 		return errors.Wrapf(err, "failed to initiate db transaction")
 	}
--- a/platform/fabric/services/db/driver/sql/sqlite/vault.go
+++ b/platform/fabric/services/db/driver/sql/sqlite/vault.go
@@ -53,7 +53,7 @@ func (db *VaultStore) Store(ctx context.Context, txIDs []driver.TxID, writes dri
 	db.GlobalLock.RLock()
 	defer db.GlobalLock.RUnlock()
 
-	tx, err := db.writeDB.Begin()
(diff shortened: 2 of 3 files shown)


mbrandenburger (approved):
Thank you @EvanYan1024

PR #1571, opened 2026-07-08
feat(comm): expose LocalPKID in SessionInfo for symmetric session identity  #1570

### LINKED ISSUE #1570: feat(comm): expose LocalPKID in SessionInfo for symmetric session identity
# Summary

SessionInfo currently exposes EndpointPKID — the cryptographically verified public key identifier of the remote peer. However, a view has no symmetric way to resolve its own identity from the session without out-of-band lookups or hardcoded configuration.

This issue asks to add LocalPKID to SessionInfo, providing the local node's peer ID as populated by the communication layer. It is the local-side counterpart of EndpointPKID: the value that remote peers observe as their EndpointPKID when talking to this node.

# Motivation

 Protocols that require each party to sign messages and verify counterpart signatures currently have no clean way to obtain the local identity from within a view. The typical workaround is to hardcode an alias or pass the identity as an input parameter. With LocalPKID available on the session, both roles can resolve and use their cryptographic material symmetrically, without any out-of-band configuration.

adecaro:
Fixes #1570

### DIFF
--- a/docs/platform/view/programming-model.md
+++ b/docs/platform/view/programming-model.md
@@ -364,8 +364,39 @@ case <-time.After(1 * time.Minute):
 
 For complete examples, see:
 - [pingpong](../../../integration/fsc/pingpong) - Simple ping/pong protocol with timeout and validation
+- [signedpingpong](../../../integration/fsc/signedpingpong) - Symmetric message signing using `SessionInfo.LocalPKID` and `SessionInfo.RemotePKID`
 - [stoprestart](../../../integration/fabric/stoprestart) - Session handling with stop/restart scenarios
 - [IOU](../../../integration/fabric/iou) - Complex Fabric integration with state management and endorsement collection
+
+## Symmetric Signing with Session Identity
+
+In protocols where each party must sign its message and verify the counterpart's signature, both identities — local and remote — can be resolved directly from the session without additional parameters or out-of-band lookups.
+
+`SessionInfo` exposes two complementary fields for this:
+
+- `RemotePKID`: the public key identifier of the **remote** peer, cryptographically bound by the transport layer.
+- `LocalPKID`: the public key identifier of the **local** node, the same value the remote peer sees as its `RemotePKID`.
+
+A view uses the endpoint service to map either PKID to a full `view.Identity`, then retrieves a signer or verifier from the sig service:
+
+```go
+endpointSvc := endpoint.GetService(viewCtx)
+sigSvc, err := sig.GetService(viewCtx)
+
+// Resolve own identity and sign the outgoing payload.
+localID, err := endpointSvc.GetIdentity("", session.Info().LocalPKID)
+signer, err := sigSvc.GetSigner(localID)
+signature, err := signer.Sign(payload)
+
+// Resolve the remote identity and verify the incoming signature.
+remoteID, err := endpointSvc.GetIdentity("", session.Info().RemotePKID)
+verifier, err := sigSvc.GetVerifier(remoteID)
+err = verifier.Verify(payload, signature)
+```
+
+Both sides of the protocol apply the same pattern — the initiator's `LocalPKID` equals the responder's `RemotePKID` for the same session, and vice versa. This makes the logic symmetric and avoids hardcoding identity references in either view.
+
+For a complete end-to-end example of this pattern with negative-case checks, see [`integration/fsc/signedpingpong`](../../../integration/fsc/signedpingpong).
 
 ## Suggested Development Flow
 
--- a/integration/fsc/signedpingpong/responder.go
+++ b/integration/fsc/signedpingpong/responder.go
@@ -1,0 +1,102 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package signedpingpong
+
+import (
+	"bytes"
+	"time"
+
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/utils"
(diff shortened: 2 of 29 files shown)


docs/platform/view/programming-model.md
```diff
@@ -364,9 +364,40 @@ case <-time.After(1 * time.Minute):
...
 - [IOU](../../../integration/fabric/iou) - Complex Fabric integration with state management and endorsement collection
 
+## Symmetric Signing with Session Identity
+
+In protocols where each party must sign its message and verify the counterpart's signature, both identities — local and remote — can be resolved directly from the session without additional parameters or out-of-band lookups.
+
+`SessionInfo` exposes two complementary fields for this:
+
+- `EndpointPKID`: the public key identifier of the **remote** peer, cryptographically bound by the transport layer.
+- `LocalPKID`: the public key identifier of the **local** node, the same value the remote peer sees as its `EndpointPKID`.
```
mbrandenburger: I am wondering if we should rethink naming here. What about `RemotePKID` and `LocalPKID`. Or `SenderPKID` and `ReceiverPKID`?

I understand that we already have `Endpoint` and `EndpointPKID` as fields and here we just add `LocalPKID`.  

My personal preference would be the following naming:

`Endpoint` -> `RemoteEndpoint`
`EndpointPKID` -> `RemotePKID`
`LocalPKID`

platform/view/services/comm/session.go
```diff
@@ -134,6 +135,7 @@ func (n *NetworkStreamSession) Info() view.SessionInfo {
 		CallerViewID: n.callerViewID,
 		Endpoint:     n.endpointAddress,
 		EndpointPKID: n.endpointID,
+		LocalPKID:    n.localPKID,
```
mbrandenburger: Can you please check that we also populate `LocalPKID` in `platform/view/services/comm/session/bidi.go` - while probably not really an issue better be consistent.

platform/view/services/comm/sig_exchange_test_utils.go
```diff
@@ -0,0 +1,220 @@
...
+	return raw
+}
+
+func unmarshalSignedEnvelope(raw []byte) (*signedEnvelope, error) {
+	env := &signedEnvelope{}
+	if err := json.Unmarshal(raw, env); err != nil {
+		return nil, err
+	}
+	return env, nil
+}
```
mbrandenburger: we have this code for testing also within the signed pingpong test. Maybe we move to shared place? Moreover, I believe we should not call this struct *Envelope as this terms is very well recognised with our Fabric(-x) types, what about just calling it `SignedMessage` or similar WDYT?

platform/view/services/comm/host/libp2p/host.go
```diff
@@ -129,6 +129,14 @@ func (h *host) Wait() {
 	h.finderWg.Wait()
 }
 
+// PeerID returns the local node's libp2p peer identifier as a string. This is
+// the value peers observe as the remote peer ID (and SessionInfo.EndpointPKID)
+// when communicating with this node. It shadows the embedded host3.Host.ID(),
+// which returns a peer.ID, to satisfy the host.P2PHost interface.
+func (h *host) PeerID() host2.PeerID {
+	return h.Host.ID().String()
```
mbrandenburger: The libp2p impl of `h.Host.ID().String()` calls base58 encoding every time. Since the ID won't change during the lifetime, we could think about storing the host ID as string and just return it here instead of recomputing every time and thereby increase unnecessary allocations. Maybe not a big deal.

integration/fsc/signedpingpong/responder.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "bob: failed resolving alice's identity")
+	}
+	verifier, err := sigSvc.GetVerifier(aliceID)
+	if err != nil {
+		return nil, errors.Wrap(err, "bob: failed getting verifier for alice")
+	}
+	if err := verifier.Verify(env.Payload, env.Signature); err != nil {
+		return nil, errors.Wrap(err, "bob: alice's signature verification failed")
+	}
```
mbrandenburger: I think this is a easy to understand flow for developer with the corresponding background. I am wondering if we could additionally provide a high-level abstraction service that maybe wraps around Session or allows the session with certain options.

```go
sesion := viewCtx.Session(WithMesssageVerification(), WithMessageSigning()) 

// or
sesion := SecureSession(viewCtx.Session())
```

The send and receive methods would automatically do the message un/wrapping and sign/verify.

platform/view/services/comm/sig_exchange_test_utils.go
```diff
@@ -0,0 +1,220 @@
...
+
+	require.NoError(t, p.Verify(remoteID, env.Payload, env.Signature),
+		"[%s] signature verification failed for payload %q", p.Name, string(env.Payload))
+
+	// Negative check: a tampered payload must not verify against the same signature.
+	tampered := append([]byte("x"), env.Payload...)
+	require.Error(t, p.Verify(remoteID, tampered, env.Signature),
+		"[%s] tampered payload unexpectedly verified", p.Name)
+
+	return remoteID
```
mbrandenburger: Note that this method is also called from inside `wg.Go` and thus should use `assert` over `require`. Maybe better return the error and let the caller use `require` or `assert` accordingly.

mbrandenburger (changes requested):
Thank you @adecaro for working on this PR. Great work adding additional security to the p2p comm stack. Please see my comment below.

adecaro:
Hi @mbrandenburger , I have addressed most of your comments. Regarding `SecureSession`, I would prefer to handle that in another PR if it is okay for you. Thanks 🙏

mbrandenburger (approved):

PR #1582, opened 2026-07-15
fix(metrics): cache collectors in the prometheus provider

### LINKED ISSUE #1581: Prometheus metrics provider returns unregistered collectors on duplicate metric requests
### Description

`metrics.Provider` builds and registers a fresh collector on every `NewCounter`/`NewGauge`/`NewHistogram` call. When the same metric is requested more than once (by two components, or by one component called repeatedly):

- with `SkipRegisterErr: false`, the second request panics with `duplicate metrics collector registration attempted`;
- with `SkipRegisterErr: true` (how the SDK wires the provider), the registration error is silently swallowed and the second requester receives a collector that is **not registered**: everything it records is invisible on the `/metrics` endpoint.

Every request also pays a full collector construction plus a registry round-trip (global lock), even when the metric already exists.

The same class of issue surfaced in the token-sdk (LFDT-Panurus/panurus#1906, LFDT-Panurus/panurus#1907), where the conclusion was to fix it once in the smart client's provider so that every consumer is protected.

### Proposed fix

Cache collectors in `prometheus.Provider`, keyed by metric kind and fully-qualified name: repeated requests return the collector built first and bypass the registry entirely; reusing a name with different options or a different metric type fails with a descriptive error. PR to follow.

EvanYan1024:
## Motivation

`metrics.Provider` builds and registers a fresh collector on every request. When the same metric is requested more than once, the second registration fails: with `SkipRegisterErr: false` it panics, and with `SkipRegisterErr: true` (how the SDK wired the provider) the error is swallowed and the requester silently receives an **unregistered** collector — everything it records is invisible on the `/metrics` endpoint. Every request also pays a collector construction plus a registry round-trip even when the metric already exists.

The same class of issue surfaced in the token-sdk (LFDT-Panurus/panurus#1906, LFDT-Panurus/panurus#1907), where the discussion concluded the deduplication belongs here in the smart client's provider, so every consumer is protected. cc @mbrandenburger

## Changes

- `prometheus.Provider` caches collectors under `<kind>|<fully-qualified name>`: repeated requests return the collector registered first and bypass both collector construction and the registry; first creation is serialized with a double-checked lock (panic-safe, so a failed registration does not poison the cache).
- `SkipRegisterErr` is removed entirely (the `Provider` field and the `NewMetricsProvider` parameter). With the cache plus the `AlreadyRegisteredError` handling below, duplicate registration is fully covered, so the only failures left are genuine programming errors — the same name requested with different options or a different collector type. Those now always fail fast with a descriptive panic instead of silently handing out a collector that is invisible on `/metrics`.
- Cache hits validate the requested definition: reusing a name with different options or a different metric type panics with a descriptive error.
- `AlreadyRegisteredError` is handled with a checked type assertion: a compatible existing collector is shared; an incompatible one produces a descriptive error instead of an unchecked interface conversion panic.
- `LabelHelp` is doc-only and never reaches the prometheus collector, so it is excluded from the cached definition: requests differing only in `LabelHelp` share the same collector instead of conflicting.
- The cached options snapshot their `LabelNames`/`Buckets`, so callers mutating those slices after creation cannot corrupt the cached definition.
- The caller-package-based namespace/subsystem inference is unaffected: it resolves before the cache key is computed, which is also why the cache lives inside the provider rather than in a decorator around it.

Known limitation: prometheus descriptors do not cover histogram `Buckets`/native-histogram options, so a `HistogramVec` with an identical descriptor registered outside this provider is reused as-is (the standard `ExistingCollector` idiom).

## Tests

- repeated and concurrent requests for the same definition register exactly once (asserted with a counting registerer) and share one collector;
- a collector registered directly with the registerer beforehand is shared: the provider returns a handle backed by the existing collector and both write to the same series;
- requests differing only in `LabelHelp` share one collector;
- conflicting label names, conflicting metric types, and a pre-registered non-vector collector with the same descriptor fail with descriptive errors;
- mutating the opts slices after creation does not affect the cached definition;
- a panicking registration leaves the cache usable.

Fixes #1581


platform/view/services/metrics/prometheus/provider.go
```diff
@@ -135,10 +162,60 @@ func parseFullPkgName(fullPkgName string, replacements map[string]string, params
...
+}
+
+// fromCache returns the cached collector after checking that opts match the
+// ones it was created with. On mismatch it panics, or falls back to a fresh
+// unregistered collector when SkipRegisterErr is set.
+func fromCache[V prom.Collector, O any](p *Provider, key string, e *cacheEntry, opts O, build func() V) V {
+	c, sameType := e.collector.(V)
+	if sameType && reflect.DeepEqual(e.opts, opts) {
+		return c
+	}
```
mbrandenburger, line 191: I am wondering if we have an issue with `o.LabelHelp` here. It will be used the the deep equal comparison but its not part of the `key`. So two otherwise-identical metrics that differ only in LabelHelp now panic if `!p.SkipRegisterErr`?! WDYT?

platform/view/services/metrics/prometheus/provider.go
```diff
@@ -135,10 +162,60 @@ func parseFullPkgName(fullPkgName string, replacements map[string]string, params
...
+	if p.SkipRegisterErr {
+		return build()
```
mbrandenburger: I am not sure about this path here. When `SkipRegisterErr` is set `true`, we will create a collector without registering it. Hence, we still have the problem that the consumer code could get a collector that is not visible via `/metrics`. 
In other words, for the conflict case this code reintroduces the same silent behavior we actually try to fix in this PR.

Is it useful at all to cover this situation? Do we still want `SkipRegisterErr`. When will registering actual fail - if something is already registered?! That particular case we now cover with the cache. Not sure yet what the best strategy is. WDYT?

platform/view/services/metrics/prometheus/provider.go
```diff
@@ -135,10 +162,60 @@ func parseFullPkgName(fullPkgName string, replacements map[string]string, params
...
+// registered, that collector is returned instead so all requesters share it.
+func register[V prom.Collector](p *Provider, fqName string, c V) prom.Collector {
+	err := prom.Register(c)
+	if err == nil {
+		return c
+	}
+	are := prom.AlreadyRegisteredError{}
+	if errors.As(err, &are) {
+		if existing, ok := are.ExistingCollector.(V); ok {
+			return existing
```
mbrandenburger, line 205: I have the feeling we are missing a test for the happy path if a collector already exists.

mbrandenburger (changes requested):
Thank you @EvanYan1024 for creating this PR. I think we are going into the right direction here - great thing! Please see my comments below.

Additionally, once we get this PR in - we should have a follow-up issue/PR to cleanup `operations/metrics.go` where it seems we solved the same issue just for a single gauge.

platform/view/services/metrics/prometheus/provider.go
EvanYan1024: Good catch. `LabelHelp` is doc-only and never reaches the prometheus collector, so it is now excluded from the cached definition entirely — requests differing only in `LabelHelp` share the same collector instead of panicking. Added a test for this.

platform/view/services/metrics/prometheus/provider.go
EvanYan1024: Agreed — I removed `SkipRegisterErr` entirely in `0e4305b4`. With the cache plus the `AlreadyRegisteredError` handling, duplicate registration is fully covered, so the only failures left are genuine programming errors (same name requested with different options or a different collector type). Those should fail fast rather than silently hand out a collector that is invisible on `/metrics`.

Worth noting: the only in-tree caller (`sdk/dig/sdk.go`) hardcoded `skipRegisterErr=true`, so the panic path was effectively never active in practice — every conflict was already being swallowed silently. Dropping the flag also removes the bool parameter from `operations.NewMetricsProvider`.

platform/view/services/metrics/prometheus/provider.go
EvanYan1024: Added in `0e4305b4`: a collector registered directly with the default registerer beforehand — the provider returns a counter backed by the existing collector and both handles write to the same series.

EvanYan1024:
Thanks for the review @mbrandenburger! 
Agreed on the `operations/metrics.go` cleanup (the single-gauge workaround) — happy to open a follow-up issue/PR once this lands.

mbrandenburger (approved):
Thank you @EvanYan1024 for working on this. The changes are solid. Could you please update the PR description with the latest changes related to the removal of `SkipRegisterErr`, then I will merge. Thanks

EvanYan1024:
> 谢谢@EvanYan1024感谢您为此付出的努力。这些改动已经很完善了。能否请您更新 PR 描述，添加与移除相关的最新更改`SkipRegisterErr`，然后我将合并。谢谢。

done

PR #1583, opened 2026-07-15
Fabric-x updates

mbrandenburger:
This PR updates fabric-x-common dependency and the fabric-x-committer test container.

Note that the golang version bump comes from fabric-x-common.


SaidAltury-ibm (approved):
LGTM

PR #1587, opened 2026-07-16
refactor(metrics): drop version gauge caching workaround

### LINKED ISSUE #1586: Remove the version gauge registration workaround in operations/metrics.go
Since #1582 the prometheus provider caches collectors, the package-level singleton in `operations/metrics.go` that guards `fsc_version` against duplicate registration is redundant.

Follow-up suggested by @mbrandenburger in the #1582 review.

EvanYan1024:
## Motivation

`operations/metrics.go` guards `fsc_version` against duplicate registration with a package-level singleton (`gaugeLock`/`promVersionGauge`), special-casing `*prometheus.Provider`. Since #1582 the prometheus provider caches collectors itself, so this workaround is redundant — the follow-up cleanup suggested by @mbrandenburger in the #1582 review.

## Changes

- `versionGauge` reduces to `provider.NewGauge(fscVersion)`; the provider-type switch, the lock, and the cached gauge are removed.
- `versionGauge` stays in place, so the caller-package-based namespace/subsystem inference is unaffected.

## Tests

Covered by the existing suites: `operations` (the system tests run through `versionGauge` on startup) and `prometheus` (the provider caching semantics from #1582). All pass with `-race`.

Fixes #1586

### DIFF
--- a/platform/view/services/metrics/operations/metrics.go
+++ b/platform/view/services/metrics/operations/metrics.go
@@ -7,37 +7,19 @@ SPDX-License-Identifier: Apache-2.0
 package operations
 
 import (
-	"sync"
-
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics/disabled"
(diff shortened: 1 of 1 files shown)


mbrandenburger (approved):
Thank you @EvanYan1024

PR #1590, opened 2026-07-16
fix(fabricx): deterministic vault Marshaller.Marshal

### LINKED ISSUE #1588: fabricx vault Marshaller.Marshal produces non-deterministic output for identical ReadWriteSet
## Summary

`Marshaller.Marshal` in `platform/fabricx/core/vault/marshal.go` can serialize the exact same `vault.ReadWriteSet` to different byte sequences across repeated calls. The decoded content is semantically identical each time, but the raw bytes differ, which breaks anything relying on marshalling being deterministic/reproducible (e.g. byte-for-byte comparison, hashing, or `Vault.Match`, which compares stored vs. provided bytes directly).

## Root cause

In `marshal()` (`platform/fabricx/core/vault/marshal.go`), the function collects reads/writes into intermediate Go maps (`namespaceSet`, and per-namespace `readSet`, `writeSet`, `readWriteSet`), then builds the output proto's repeated fields by ranging directly over those maps:

```go
namespaces := make([]*applicationpb.TxNamespace, 0)
for _, namespace := range namespaceSet {          // random order
    for _, read := range namespace.readSet { ... }        // random order
    for _, write := range namespace.writeSet { ... }      // random order
    for _, readWrite := range namespace.readWriteSet {...} // random order
}
```

Go intentionally randomizes map iteration order on every `range`. Since `applicationpb.TxNamespace.ReadsOnly/ReadWrites/BlindWrites` and `Tx.Namespaces` are proto `repeated` (list) fields, `proto.Marshal` faithfully serializes whatever order the slices were built in — so the random map iteration order leaks straight into the serialized bytes.

Notably, `vault.ReadWriteSet` already carries `OrderedWrites`/`OrderedReads` fields specifically meant to preserve deterministic insertion order, but `Marshal` ignores them and rebuilds fresh unordered maps instead.

## Reproduction

A minimal test (`TestMarshal_Deterministic` in `platform/fabricx/core/vault/marshal_test.go`) builds one `ReadWriteSet` with 8 namespaces × 16 keys and calls `Marshal` 30 times on the same input, asserting each result equals the first byte slice. It reliably fails on the very first repeated call, for all of: mixed read-write, reads-only, and writes-only RWSets.

```go
rws := commonvault.EmptyRWSet()
// populate rws with several namespaces/keys ...

m := vault.NewMarshaller()
m.NsInfo = nsInfo

first, _ := m.Marshal("tx1", &rws)
for i := 0; i < 30; i++ {
    out, _ := m.Marshal("tx1", &rws)
    require.Equal(t, first, out) // fails
}
```

## Impact

Any code path that assumes `Marshal(txID, rws)` is stable across calls for the same logical RWSet can break, e.g.:
- `Vault.Match` (`platform/fabricx/core/vault/vault.go`) re-marshals a stored RWSet and compares bytes against provided bytes — this comparison can spuriously fail even when the RWSets are logically identical.
- Any hashing/signing/consensus-relevant use of the marshalled bytes.

## Suggested fix

Sort namespace IDs and per-namespace keys before appending to the output slices (or reuse the existing `OrderedWrites`/`OrderedReads` ordering) so slice population order — and therefore the serialized bytes — no longer depends on Go's randomized map iteration order.

adecaro:
Fix #1588

### DIFF
--- a/platform/fabricx/core/vault/marshal.go
+++ b/platform/fabricx/core/vault/marshal.go
@@ -9,6 +9,7 @@ package vault
 import (
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


mbrandenburger (approved):
Thank you @adecaro! Good catch and solid fix!

PR #1594, opened 2026-07-17
fix(security): harden responder-view input validation and endorsement checks

### LINKED ISSUE #1593: Harden responder-view input validation and endorsement checks
## Summary

A security review of the responder-view attack surface (code paths reached when this node answers a remote peer over a P2P session) identified several hardening opportunities in transaction/envelope deserialization and endorsement collection. Fixes have been implemented locally and are pending submission as a PR.

## Areas addressed

- **Input validation on untrusted wire bytes**: protobuf deserialization of envelopes/proposals did not fully validate nested message structure before use, which could affect availability when processing malformed input.
- **Error handling in the P2P responder path**: error details returned to a failed responder view were more verbose than necessary when relayed back to the remote caller.
- **Endorsement collection logic**: a control-flow issue in the sequential endorsement-collection path could cause a binding check to be bypassed under specific conditions; the parallel collection path was missing a verification step present in the sequential path.
- **Unbounded wait in a responder view**: one responder view could wait indefinitely for a peer response with no timeout.

## Status

Fixes for all of the above are implemented and covered by new/updated unit and fuzz tests. A PR will follow shortly.

## Scope note

Exploit-level detail is intentionally omitted from this public issue; happy to share the full write-up with maintainers privately if useful for review.

adecaro:
## Summary

Hardens the responder-view attack surface — code reachable when this node
answers a remote, untrusted peer over a P2P session — against malformed
input and endorsement-verification bypass.

- Validate wire-deserialized proposals/envelopes/transactions (nil/length
  checks on protobuf-decoded fields) instead of panicking on malformed
  input from a remote peer.
- Stop leaking internal panic/error detail to the remote caller via
  `Session.SendError`; full detail is still logged locally.
- Enforce endorsement signature and party-binding verification on both the
  sequential and parallel endorsement-collection paths.
- Add a timeout to the responder's message-receive wait so a silent peer
  can no longer block a goroutine indefinitely.
- Guard an unindexed access when no proposal responses are present.

All existing tests that previously asserted the vulnerable/buggy behavior
(`require.Panics`, accept-without-verification) have been updated to assert
the fixed/secure behavior. Fuzz seed corpora were extended with the
previously-crashing inputs, now exercised as ordinary non-panicking cases.

Refs #1593.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./platform/...`
- [x] `go test ./platform/fabric/core/generic/transaction/...`
- [x] `go test ./platform/fabric/services/endorser/...`
- [x] `go test ./platform/view/services/view/p2p/...`
- [x] `go test ./platform/fabric/services/state/...` (downstream consumer)
- [x] 15s live fuzz run of `FuzzUnpackEnvelopeFromBytes` — no new panics

### DIFF
--- a/platform/fabric/services/endorser/endorsement_proposal.go
+++ b/platform/fabric/services/endorser/endorsement_proposal.go
@@ -38,6 +38,14 @@ type answer struct {
 	party view.Identity
 }
 
+// defaultParallelEndorsementTimeout bounds, in aggregate, how long the initiator waits for
+// all contacted parties to answer when no explicit timeout has been configured via
+// WithTimeout. It matters beyond just the final receive step: session establishment and
+// sending the proposal to each party (both done inside collectEndorsement, before its own
+// receive-timeout kicks in) are not otherwise bounded, so an unresponsive or malicious party
+// could park its goroutine indefinitely and, with it, this view's wait on answerChannel.
+const defaultParallelEndorsementTimeout = 30 * time.Second
+
 type parallelCollectEndorsementsOnProposalView struct {
 	tx      EndorsementsOnProposalTransaction
 	parties []view.Identity
@@ -57,9 +65,14 @@ func (c *parallelCollectEndorsementsOnProposalView) Call(viewCtx view.Context) (
 	}
 	answerChannel := make(chan *answer, len(c.parties))
 
+	timeout := c.timeout
+	if timeout <= 0 {
+		timeout = defaultParallelEndorsementTimeout
+	}
+
 	logger.DebugfContext(viewCtx.Context(), "Collect endorsements from %d parties for TX [%s]", len(c.parties), c.tx.ID())
 	for _, party := range c.parties {
-		go c.collectEndorsement(viewCtx, party, stateRaw, answerChannel)
+		go c.collectEndorsement(viewCtx, party, stateRaw, timeout, answerChannel)
 	}
 
 	fns, err := fabric.GetFabricNetworkService(viewCtx, c.tx.Network())
@@ -74,10 +87,17 @@ func (c *parallelCollectEndorsementsOnProposalView) Call(viewCtx view.Context) (
 	}
 	vProviders := []fabric.VerifierProvider{&verifierProviderWrapper{m: ch.MSPManager()}}
 
+	deadline := time.NewTimer(timeout)
+	defer deadline.Stop()
+
 	for i := 0; i < len(c.parties); i++ {
 		logger.DebugfContext(viewCtx.Context(), "Wait for endorsement")
-		// TODO: put a timeout
-		a := <-answerChannel
+		var a *answer
+		select {
+		case a = <-answerChannel:
+		case <-deadline.C:
+			return nil, errors.Errorf("timeout waiting for endorsement from [%d] parties", len(c.parties)-i)
+		}
 		logger.DebugfContext(viewCtx.Context(), "Received endorsement")
 		if a.err != nil {
 			return nil, errors.Wrapf(a.err, "got failure [%s] from [%s]", a.party.String(), a.err)
@@ -127,6 +147,7 @@ func (c *parallelCollectEndorsementsOnProposalView) collectEndorsement(
 	viewCtx view.Context,
 	party view.Identity,
 	raw []byte,
+	timeout time.Duration,
 	answerChan chan *answer,
 ) {
 	defer logger.Debugf("Received answer for endorsement of TX [%s] from [%v]", c.tx.ID(), party)
@@ -145,7 +166,7 @@ func (c *parallelCollectEndorsementsOnProposalView) collectEndorsement(
 		return
 	}
 	r := &Response{}
-	if err := s.ReceiveWithTimeout(r, c.timeout); err != nil {
+	if err := s.ReceiveWithTimeout(r, timeout); err != nil {
 		answerChan <- &answer{err: err, party: party}
 		return
 	}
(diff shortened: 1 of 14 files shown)


platform/view/services/view/p2p/service_test.go
```diff
@@ -121,6 +122,95 @@ func TestService_MasterSessionError(t *testing.T) {
...
+// read), Service.respond now sends a generic, non-identifying error back to the remote,
+// untrusted caller via Session.SendError instead of the raw panic/error text. The
+// detailed error stays local (logged by the recover() in Service.respond), so a remote
+// peer can no longer learn internal error detail (e.g. local storage/db error strings,
+// stack-adjacent context) about the responder's internals by causing a responder view
+// to panic.
+func TestService_PanicDoesNotLeakInternalErrorToRemoteCaller(t *testing.T) {
+	t.Parallel()
+
+	sensitivePanicText := "filed getting rw set: could not open /var/lib/fabric/kvs/vault.db: permission denied"
```
SaidAltury-ibm:
```suggestion
	sensitivePanicText := "failed getting rw set: could not open /var/lib/fabric/kvs/vault.db: permission denied"
```

maybe also the source of the checked msg have a typo

platform/fabric/services/endorser/flow.go
```diff
@@ -39,8 +41,15 @@ func (s receiveView) Call(viewCtx view.Context) (any, error) {
 	// Wait to receive a state
 	ch := session.Receive()
 
-	// TODO: add timeout
-	msg := <-ch
+	timeout := time.NewTimer(time.Second * 10)
```
SaidAltury-ibm: 10 seconds looks like a magic number for me i would like something explaining the choice WYT?

platform/fabric/services/endorser/endorsement_proposal.go
```diff
@@ -64,6 +67,13 @@ func (c *parallelCollectEndorsementsOnProposalView) Call(viewCtx view.Context) (
...
+
+	ch, err := fns.Channel(c.tx.Channel())
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting channel [%s:%s]", c.tx.Network(), c.tx.Channel())
+	}
+	vProviders := []fabric.VerifierProvider{&verifierProviderWrapper{m: ch.MSPManager()}}
+
 	for i := 0; i < len(c.parties); i++ {
 		logger.DebugfContext(viewCtx.Context(), "Wait for endorsement")
 		// TODO: put a timeout
```
SaidAltury-ibm: the parallel proposal path still has an unbounded wait, why is it out of scope?

SaidAltury-ibm (commented):
THANK YOU @adecaro, I have added some comments tell my WYT?

mbrandenburger (approved):
Thank you @adecaro. This PR is solid for my side (modulo the findings from @SaidAltury-ibm). Approving here already and looking forward for the minor improvements @SaidAltury-ibm suggested.

platform/view/services/view/p2p/service_test.go
adecaro: Fixed — corrected the typo in the actual panic message (`platform/fabric/services/endorser/transaction.go`: `filed getting rw set` -> `failed getting rw set`) and updated this test's expectation and comment to match, rather than just patching the test string. See 9efd6665.

platform/fabric/services/endorser/flow.go
adecaro: Named it: added `receiveTimeout = 10 * time.Second` as a package-level const with a comment explaining it matches the existing default receive timeout used for session-level reads elsewhere in the codebase (`session.defaultReceiveTimeout`). See 9efd6665.

platform/fabric/services/endorser/endorsement_proposal.go
adecaro: Good catch, fixed. `c.timeout` was only ever consulted inside each per-party goroutine's own `ReceiveWithTimeout` call - it bounded the final receive step, but not session setup/send (both done earlier in `collectEndorsement`), and more importantly it was never actually set: the only production call site (`state.NewParallelCollectEndorsementsOnProposalView`) never calls `WithTimeout`, so `c.timeout` was zero, which `ReceiveWithTimeout`/`time.NewTimer` treats as "fire immediately"-adjacent rather than "no timeout" - not actually unbounded in the way it looked, but not a sane default either. Added `defaultParallelEndorsementTimeout = 30 * time.Second` used whenever no explicit timeout is configured, and wrapped the `Call` loop's wait on `answerChannel` itself in a `select` against an aggregate deadline, so a silent party can't block the whole view indefinitely regardless of what happens inside its own goroutine. Added `TestParallelCollectEndorsementsOnProposalView_TimesOutOnSilentParty` to cover it. See 9efd6665.

adecaro:
Thanks for the thorough review, @SaidAltury-ibm! Pushed 9efd6665 addressing all three comments:

1. Fixed the actual typo (`filed getting rw set` -> `failed getting rw set`) in `endorser/transaction.go`'s panic message, plus the test string/comment.
2. Named the 10s magic number in `flow.go` as `receiveTimeout`, documented to match the existing session-level default elsewhere in the codebase.
3. The parallel proposal path's outer wait on `answerChannel` is no longer unbounded: added a `defaultParallelEndorsementTimeout` (30s) applied whenever `WithTimeout` isn't explicitly set (which is the case at the only production call site), and wrapped the wait in a `select`/deadline so a silent party can't block the view forever. Added a regression test (`TestParallelCollectEndorsementsOnProposalView_TimesOutOnSilentParty`).

`go build ./...`, `go vet ./platform/...`, and the full `platform/fabric/...` + `platform/view/...` test suites are green.

SaidAltury-ibm (approved):
thank you @adecaro

PR #1600, opened 2026-07-21
revert(security): stop redacting responder-view errors sent to remote callers

adecaro:
## Summary

#1594 changed `Service.respond` to send a hardcoded `"responder failed"` string back to the remote caller instead of the responder view's actual error, to avoid leaking internal error/panic detail across the wire.

That decision belongs one layer up. The P2P runner's job is to decide *that* an error gets returned to the remote caller, not *what* it may contain — deciding what information is safe to disclose to a remote, potentially untrusted party is the responder application's decision to make, not the runtime's. `respond()` calls `SendError` for **every** responder-view error, not just panics — including ordinary business-logic errors (e.g. an auditor view rejecting a transfer for exceeding a payment/holding limit, or a `TokenService` validation failure). Hardcoding the runtime's response erases information legitimate callers rely on to react to (and tests rely on to assert against) the specific rejection reason, and takes away the application's ability to choose what to disclose.

This surfaced downstream in Fabric Token SDK / Panurus: bumping to a commit that included #1594 broke integration tests that check the auditor's specific rejection message returned across a P2P session (see [LFDT-Panurus/panurus#1969](https://github.com/LFDT-Panurus/panurus/pull/1969)).

This PR reverts just the redaction: `respond()` sends back the responder's actual `err.Error()` again, matching pre-#1594 behavior, and leaves it to the application/view layer to wrap or sanitize errors that could carry internal detail before returning them. The rest of #1594 (wire-deserialization validation at proposal/transaction boundaries, endorsement signature/binding verification hardening) is untouched and stays in place — only the generic-error substitution in `platform/view/services/view/p2p/service.go` is reverted.

Documentation updated to describe this responsibility split: [`docs/platform/view/view-api.md`](docs/platform/view/view-api.md) (`SendError` semantics) and [`docs/platform/view/services/view-service.md`](docs/platform/view/services/view-service.md) (P2P Service error-forwarding behavior) now state that the runtime forwards `err.Error()` verbatim and that filtering/redaction is the responder view's responsibility.

## Test plan
- [x] `go build ./...`
- [x] `go vet -all ./platform/view/...`
- [x] `go test ./platform/view/services/view/p2p/...` — updated `TestService_PanicDoesNotLeakInternalErrorToRemoteCaller` → `TestService_PanicIsReturnedToRemoteCaller` to assert the panic text *is* forwarded again, matching the reverted behavior.

### DIFF
--- a/docs/platform/view/services/view-service.md
+++ b/docs/platform/view/services/view-service.md
@@ -407,6 +407,8 @@ The View Service integrates with the P2P Service for handling incoming messages:
 4. **View Execution**: Runs the responder view in the context
 5. **Error Handling**: Sends error messages back to the initiator on failure
 
+The text sent is the raw error returned by the responder view (`err.Error()`), forwarded unmodified via `Session().SendError(...)`. The runtime does not inspect or filter it — deciding what information is safe to disclose to a remote, potentially untrusted caller is the responder view's responsibility, not the P2P service's. See the [View API](../view-api.md#send-and-receive) for guidance on what responder views should and shouldn't return as errors.
+
 ### Message Flow
 
 ```mermaid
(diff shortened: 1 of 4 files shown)


mbrandenburger (approved):
Thank you @adecaro for working on this.

PR #1603, opened 2026-07-22
feat(fabric/config): support adding a new Fabric network at runtime

### LINKED ISSUE #1602: feat(fabric/config): support adding a new Fabric network at runtime
### Description

Today a Fabric Smart Client node reads its `fabric.*` configuration once at startup
(`core.NewConfig`, `platform/fabric/core/config.go`). There is no supported way to register a
brand-new Fabric network after the process is running: `core.Config` is built once from the
config provider and keeps no reference back to it, so nothing can be merged into the live tree
later.

This adds that capability, scoped to the configuration tree only:

1. `core.Config.AddNetwork(raw []byte, validators ...NetworkConfigValidator) error` — parses the
   given raw YAML, validates every network/channel name it introduces against the existing
   Fabric naming rules (`fabric-x-common`'s `configtx.ValidateChannelID`), rejects any network
   name that already exists, runs any supplied `NetworkConfigValidator`s, and only then merges
   the configuration into the live tree (via the existing `viewconfig.Provider.MergeConfig` /
   `ProvideFromRaw` primitives) and refreshes `Config`'s in-memory state.
2. A `NetworkConfigValidator` extension point (interface + `NetworkConfigValidatorFunc` adapter)
   so SDK consumers can plug in custom, per-network validation (e.g. policy checks on driver
   name, MSP config, channel list) that runs before a candidate network is merged.
3. `FSNProvider.AddNetwork(raw []byte) error` — a thin wrapper that forwards to
   `Config.AddNetwork` using whatever `[]NetworkConfigValidator` the provider was constructed
   with.
4. Wiring: `NewFabricNetworkServiceProvider` takes a `[]NetworkConfigValidator` parameter, and
   the dig constructor in `platform/fabric/sdk/dig/fns/providers.go` collects it from a new
   `"fabric-network-config-validators"` dig group (zero-or-more contributions, same pattern as
   the existing `"fabric-platform-drivers"` group), so any SDK can contribute validators without
   touching `core`.

### Explicitly out of scope

`FSNProvider.newFNS` already re-reads live configuration on first access to a network name, so a
network added via `AddNetwork` becomes lazily instantiable without further changes. This issue
does **not** cover:
- Auto-starting the committer/delivery service for a network added at runtime (`FSNProvider.Start`
  remains a one-shot loop over the networks known at construction time).
- Re-running the one-shot startup registration loops in `platform/fabric/sdk/dig/sdk.go`
  (`registerProcessorsForDrivers`, `registerRWSetLoaderHandlerProviders`) for networks added
  afterwards.

A fully-live materialization of a runtime-added network (auto-starting its committer/delivery,
re-running those registration loops) would be reasonable follow-up work, but is a separate,
larger change.

### Implementation Steps

- [x] Add `DynamicConfigProvider` / `DynamicConfigService` interfaces (`ConfigProvider` +
      `MergeConfig` + `ProvideFromRaw`) to `platform/fabric/core/config.go`.
- [x] Extract the network-scanning logic of `NewConfig` into a reusable `loadNetworks` helper.
- [x] Add `Config.AddNetwork(raw []byte, validators ...NetworkConfigValidator) error`, with
      network-name, channel-name, and duplicate-network/duplicate-channel validation, and a
      mutex-guarded refresh of `Config`'s state after a successful merge.
- [x] Add the `NetworkConfigValidator` interface and `NetworkConfigValidatorFunc` adapter.
- [x] Add `FSNProvider.AddNetwork` and thread a `[]NetworkConfigValidator` through
      `NewFabricNetworkServiceProvider`.
- [x] Wire a new `"fabric-network-config-validators"` dig group into
      `platform/fabric/sdk/dig/fns/providers.go`.
- [x] Unit tests covering: valid network addition, rejection of an already-existing network,
      rejection of malformed network/channel names, rejection of duplicate channel names within
      one network, custom validators running (and blocking) the merge, and the same behavior
      exercised through `FSNProvider.AddNetwork`.
- [x] Godoc for every new/changed exported symbol.

### Additional Information

Design modeled on the existing `AddConfiguration(raw []byte)` pattern used elsewhere in the
project for runtime config injection, adapted to Fabric's network/channel naming rules.

adecaro:
## Summary

Adds support for registering a new Fabric network at runtime, scoped to the configuration
tree only (see [Explicitly out of scope](#explicitly-out-of-scope) below).

- `core.Config.AddNetwork(raw []byte, validators ...NetworkConfigValidator) error` — parses the
  given raw yaml, validates every network/channel name it introduces against the existing
  Fabric naming rules, rejects any network name that already exists, runs any supplied
  `NetworkConfigValidator`s, and only then merges it into the live configuration tree (via the
  existing `MergeConfig`/`ProvideFromRaw` primitives on `viewconfig.Provider`) and refreshes
  `Config`'s in-memory state.
- `NetworkConfigValidator` interface + `NetworkConfigValidatorFunc` adapter — a pluggable,
  per-network validation hook that runs before a candidate network's configuration is merged.
- `FSNProvider.AddNetwork(raw []byte) error` — forwards to `Config.AddNetwork` using the
  `[]NetworkConfigValidator` the provider was constructed with.
- `NewFabricNetworkServiceProvider` now takes a `[]NetworkConfigValidator` parameter, wired via a
  new `"fabric-network-config-validators"` dig group in
  `platform/fabric/sdk/dig/fns/providers.go` (mirroring the existing `"fabric-platform-drivers"`
  group), so any SDK can contribute validators without touching `core`.
- Godoc added for every new/changed exported symbol touched by this change.

## Explicitly out of scope

`FSNProvider.newFNS` already re-reads live configuration on first access to a network name, so a
network added via `AddNetwork` becomes lazily instantiable without further changes. This PR does
**not**:
- Auto-start the committer/delivery service for a network added at runtime (`FSNProvider.Start`
  remains a one-shot loop over the networks known at construction time).
- Re-run the one-shot startup registration loops in `platform/fabric/sdk/dig/sdk.go`
  (`registerProcessorsForDrivers`, `registerRWSetLoaderHandlerProviders`) for networks added
  afterwards.

## Test plan

- [x] `make lint-fmt` / `make checks` — clean.
- [x] `go test -race -cover ./platform/fabric/core/... ./platform/fabric/sdk/...` — all pass.
- [x] `go test -race --run '(TestWiring)' ./platform/{common,fabric,fabricx,view}/sdk/dig/...` —
      all pass.
- [x] New unit tests in `platform/fabric/core/config_test.go` cover: valid network addition,
      rejecting an already-existing network, rejecting malformed network/channel names,
      rejecting duplicate channel names within one network, allowing the same channel name
      across different networks, custom validators blocking/allowing a merge, and the same
      behavior exercised through `FSNProvider.AddNetwork`.

Fixes #1602

### DIFF
--- a/platform/fabric/core/config.go
+++ b/platform/fabric/core/config.go
@@ -7,87 +7,323 @@ SPDX-License-Identifier: Apache-2.0
 package core
 
 import (
+	"regexp"
+	"sort"
 	"strings"
+	"sync"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	commondriver "github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
+	genericconfig "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/config"
+	viewconfig "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/config"
 )
 
+// ConfigProvider is the minimal read surface Config needs to discover and describe Fabric
+// networks: the ability to unmarshal a configuration key into an arbitrary Go value.
 type ConfigProvider interface {
 	UnmarshalKey(key string, rawVal any) error
 }
 
-type FSNConfig struct {
-	Default bool   `yaml:"default,omitempty"`
-	Driver  string `yaml:"driver,omitempty"`
+// DynamicConfigProvider is a ConfigProvider that also supports injecting new
+// configuration into the live configuration tree at runtime.
+type DynamicConfigProvider interface {
+	ConfigProvider
+	// MergeConfig merges the given raw yaml configuration into the current configuration.
+	MergeConfig(raw []byte) error
+	// ProvideFromRaw returns a new, independent provider whose configuration is loaded
+	// from the given raw yaml representation.
+	ProvideFromRaw(raw []byte) (*viewconfig.Provider, error)
 }
 
+// DynamicConfigService is a commondriver.ConfigService that also supports injecting new
+// configuration into the live configuration tree at runtime. It satisfies DynamicConfigProvider.
+type DynamicConfigService interface {
+	commondriver.ConfigService
+	// MergeConfig merges the given raw yaml configuration into the current configuration.
+	MergeConfig(raw []byte) error
+	// ProvideFromRaw returns a new, independent provider whose configuration is loaded
+	// from the given raw yaml representation.
+	ProvideFromRaw(raw []byte) (*viewconfig.Provider, error)
+}
+
+// FSNConfig is the per-network configuration read from the `fabric.<name>` configuration key.
+type FSNConfig struct {
+	// Default marks this network as the default one. At most one network may set this to true.
+	Default bool `yaml:"default,omitempty"`
+	// Driver is the name of the driver.Driver that should be used to instantiate this network.
+	// If empty, every registered driver is tried in turn.
+	Driver string `yaml:"driver,omitempty"`
+}
+
+// Config exposes the set of Fabric networks configured for this node: their names, which one
+// is the default, and their per-network FSNConfig. It is safe for concurrent use; AddNetwork
+// may add new networks to a live Config while other goroutines read from it.
 type Config struct {
+	mu sync.RWMutex
+
+	provider       DynamicConfigProvider
 	configurations map[string]*FSNConfig
 	names          []string
 	defaultName    string
 }
 
-func NewConfig(configProvider ConfigProvider) (*Config, error) {
-	var value any
-	if err := configProvider.UnmarshalKey("fabric", &value); err != nil {
-		return nil, errors.Wrap(err, "failed unmarshalling `fabric` key")
-	}
-	m := value.(map[string]any)
-	var names []string
-	var defaultName string
-	configurations := map[string]*FSNConfig{}
-	for k := range m {
-		name := k
-		if strings.ToLower(name) != "enabled" {
-			names = append(names, name)
-
-			var fsnConfig FSNConfig
-			if err := configProvider.UnmarshalKey("fabric."+name, &fsnConfig); err != nil {
-				return nil, errors.Wrapf(err, "failed unmarshalling `fabric.%s` key", name)
-			}
-			configurations[name] = &fsnConfig
-			logger.Debugf("found fabric network [%s], driver [%s]", name, fsnConfig.Driver)
-
-			// is this default?
-			if fsnConfig.Default || name == "default" {
-				if len(defaultName) != 0 {
-					logger.Warnf("only one network should be set as default, ignoring [%s], default is set to [%s]", name, defaultName)
-					continue
-				}
-				defaultName = name
-			}
-		}
-	}
-	if len(names) == 0 {
-		return nil, errors.New("no fabric network configured")
-	}
-
-	if len(defaultName) == 0 {
-		defaultName = names[0]
-		if len(names) > 1 {
-			logger.Warnf("no default network configured, set it to [%s]", defaultName)
-		}
+// NewConfig builds a Config by scanning the `fabric` key of the given configProvider. It
+// returns an error if no Fabric network is configured.
+func NewConfig(configProvider DynamicConfigProvider) (*Config, error) {
+	configurations, names, defaultName, err := loadNetworks(configProvider)
+	if err != nil {
+		return nil, err
 	}
 
 	return &Config{
+		provider:       configProvider,
 		names:          names,
 		defaultName:    defaultName,
 		configurations: configurations,
 	}, nil
 }
 
+// loadNetworks scans the `fabric` key of the given provider and returns the discovered
+// per-network configurations, the ordered (sorted) list of network names, and the resolved
+// default network name. Every network name and channel name it discovers is validated against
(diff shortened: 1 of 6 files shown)


platform/fabric/core/config_test.go
```diff
@@ -0,0 +1,250 @@
...
+	require.NoError(t, err)
+	assert.ElementsMatch(t, []string{"default"}, cfg.Names())
+	assert.Equal(t, "default", cfg.DefaultName())
+
+	fsnConfig, err := cfg.Config("default")
+	require.NoError(t, err)
+	assert.Equal(t, "generic", fsnConfig.Driver)
+}
+
+func TestAddNetwork_Valid(t *testing.T) {
```
mbrandenburger, line 112: I think we should add some missing test scenarios where both networks are set as `default: true` or default: true`. The validator "should" return an error if not exactly one network is set as default? WDYT?

platform/fabric/core/config.go
mbrandenburger, line 142: as we populate `names` when iterating over the map above, we may have a non-deterministic behavior here. When we call `loadNetworks()` twice, we may get different `defaultName` returned. Would it harm to sort names before setting a default? Seems this issue was here before. WDYT?

platform/fabric/core/config_test.go
```diff
@@ -0,0 +1,250 @@
...
+	// the live tree must be left untouched
+	assert.ElementsMatch(t, []string{"default"}, cfg.Names())
+}
+
+func TestAddNetwork_RejectsBadNetworkName(t *testing.T) {
+	t.Parallel()
+	cases := []string{
+		"Bad_Net",  // uppercase and underscore
+		"1network", // must start with a letter
+		"",         // empty (would be skipped by the scanner, covered separately)
```
mbrandenburger: should we just remove this test case here as it not fired below anyway due to ...
```go
if len(name) == 0 {
```

And instead add a test like `TestAddNetwork_EmptyNetworkName`?

platform/fabric/core/config.go
```diff
@@ -8,31 +8,93 @@ package core
 
 import (
 	"strings"
+	"sync"
+
+	"github.com/hyperledger/fabric-x-common/common/configtx"
```
mbrandenburger: I believe that `platform/fabric/core/config.go` should not rely on `fabric-x-common`. Can we just have this little channel name validator function as a internal function of our fabric platform?

mbrandenburger: However, we also use this code now to validate fabric-x configs - where it would be totally fine to use fabric-x-common. Tricky one.

mbrandenburger: Moreover, do we also validate network/channel names when parsing the config core.yaml during startup? We should check that there is no inconsistency.

mbrandenburger (changes requested):
Thank you @adecaro for working on this. This PR is in a solid shape already. Please see my additional comments below.

adecaro:
Hi @mbrandenburger, thanks for the review. I agree with your points, I'll address them. 🙏

platform/fabric/core/config.go
adecaro: Fixed in 0cb15d4: `loadNetworks` now collects the map keys into a slice and sorts them (`sort.Strings`) before iterating, so `names` and the resolved `defaultName` no longer depend on Go's non-deterministic map iteration order.

platform/fabric/core/config_test.go
adecaro: Fixed in 0cb15d4: added `explicitDefaults()` which collects every network that is explicitly `default: true` or named `default`. `loadNetworks` now errors with "multiple fabric networks are set as default" if more than one is found, both at startup (`NewConfig`) and in `AddNetwork` (checking existing + incoming networks together, since each side only sees its own set). Added `TestNewConfig_RejectsMultipleExplicitDefaults`, `TestNewConfig_RejectsExplicitDefaultAlongsideNamedDefault`, and `TestAddNetwork_RejectsSecondDefault`.

platform/fabric/core/config_test.go
adecaro: Fixed in 0cb15d4: removed the `""` case from the table test (confirmed via a scratch test that it does NOT actually get skipped — it flows through to `validateNetworkName` and produces a real, meaningful error) and added `TestAddNetwork_EmptyNetworkName` exercising that path directly, asserting the error message and that `cfg.Names()` is left untouched.

platform/fabric/core/config.go
adecaro: Fixed in 0cb15d4: removed the `fabric-x-common` import entirely and added a small local `validateIdentifier` (regex `^[a-z][a-z0-9.-]*$`, max 249 chars) reimplementing the same Fabric channel-ID rules, used by `validateNetworkName`/`validateChannels`. On the fabric-x concern: since the rule is just the standard Fabric identifier convention (not fabric-x-specific), duplicating it locally is fine and avoids the extra module dependency in core fabric platform code — if fabric-x ever needs different rules we can special-case it then.

Re: startup validation consistency — good catch, they were inconsistent. `loadNetworks` (which backs both `NewConfig` at startup and `AddNetwork` at runtime) now calls `validateNetworkName`/`validateChannels` inline while scanning, so the exact same checks run on both paths. Verified against all existing YAML fixtures in the repo — none violate the new rules.

mbrandenburger (approved):
Thank you @adecaro for updating the PR. Solid!

PR #1605, opened 2026-07-22
feature(nwo): allow minimal FSC fabric config for runtime injection

### LINKED ISSUE #1604: feature(nwo): allow minimal FSC fabric config for runtime injection
### Problem Description

When NWO generates a network topology, every FSC node linked to a `fabric` (or `fabricx`) network gets a full `fabric:` configuration block rendered into its `core.yaml` — MSPs, peers, TLS settings, ordering, channels/chaincodes, endpoint resolvers, etc. This block is generated from `DefaultFSCFabricExtensionTemplate` and looks like:

```yaml
fabric:
  enabled: true
  <networkName>:
    default: true
    driver: generic
    msps:
      ...
    peers:
      ...
    channels:
      ...
    endpoint:
      ...
```

There's no way to tell NWO to skip generating this block and instead only assert that fabric support is enabled, so the rest of the fabric configuration can be supplied to the FSC node at runtime (e.g. injected by an external process/orchestrator instead of being baked in at generation time).

### Proposed Solution

Add an opt-in flag on the fabric `Topology` that, when enabled, makes NWO emit only the minimal stanza for every FSC node in that topology, instead of the full generated block:

```yaml
fabric:
  enabled: true
```

Proposed API:

```go
topology := fabric.NewDefaultTopology()
topology.EnableMinimalFSCFabricConfig()
```

- The flag applies to the whole topology (all FSC nodes in the network), not per-node.
- Since `fabricx.Topology` embeds `*fabric.Topology`, the flag and setter are inherited automatically — no fabricx-specific change needed.
- fabricx's scv2 sidecar extensions (`queryService`, `notificationService`), which independently contribute fragments under the same `fabric:` key, are unaffected and continue to be merged in as today.
- Default behavior (flag unset) is unchanged — this is purely opt-in.

### Alternatives Considered

- A per-FSC-node flag instead of a topology-wide one: rejected as unnecessary complexity: the use case (deferring fabric config to runtime injection) applies uniformly to every FSC node in a network.

### Implementation Notes

- `integration/nwo/fabric/topology/topology.go`: add `MinimalFSCFabricConfig bool` field + `EnableMinimalFSCFabricConfig()` setter on `Topology`.
- `integration/nwo/fabric/topology/core_template.go`: add a `MinimalFSCFabricExtensionTemplate` constant containing just `fabric:\n  enabled: true`.
- `integration/nwo/fabric/network/network_support.go`: in `GenerateCoreConfig`'s FSCPeer branch, select the minimal template instead of `FSCFabricExtensionTemplate()` when the flag is set. Only the FSCPeer branch is affected — real peer `core.yaml` generation is untouched.

adecaro:
Fixes #1604

### Problem

When NWO generates a network topology, every FSC node linked to a `fabric` (or `fabricx`) network gets a full `fabric:` block rendered into its `core.yaml` — MSPs, peers, TLS settings, ordering, channels/chaincodes, endpoint resolvers, etc. There's no way to have NWO generate everything else normally but skip this block, so the fabric configuration can be supplied to the FSC node at runtime instead of being baked in at generation time.

### Change

**NWO: opt-in minimal FSC config**

Add an opt-in `Topology.MinimalFSCFabricConfig` flag (with an `EnableMinimalFSCFabricConfig()` setter) on the fabric `Topology`. When enabled, NWO emits only:

```yaml
fabric:
  enabled: true
```

into every FSC node's `core.yaml`, instead of the full `DefaultFSCFabricExtensionTemplate` block.

- The flag is topology-wide (applies to every FSC node in the network), not per-node.
- `fabricx.Topology` embeds `*fabric.Topology`, so the flag and setter are inherited automatically — no fabricx-specific change was needed.
- fabricx's scv2 sidecar extensions (`queryService`, `notificationService`) independently contribute `fabric:` fragments and are unaffected — they continue to be deep-merged in via `common.MergeYAML` as before.
- Only the FSC-node (`FSCPeer`) branch of `GenerateCoreConfig` is touched; real-peer `core.yaml` generation (`DefaultCoreTemplate`) is unchanged.
- Default behavior (flag unset) is byte-for-byte unchanged.

`integration/nwo/fabric/network/network_support.go` also gains `Network.RenderFSCFabricExtension(p *topology.Peer)`, extracted from `GenerateCoreConfig`'s FSCPeer branch. It always renders the *full* extension YAML for a given FSC peer, regardless of `MinimalFSCFabricConfig`, so the full configuration remains obtainable even for a node that booted minimal — needed to actually inject it at runtime (see below).

**Platform: allow a Fabric node to boot with zero configured networks**

`platform/fabric.enabled: true` with no `fabric.<name>` networks previously failed node startup outright (`"no fabric network configured"` from `platform/fabric/core/config.go`, and `"no fabric network names found"` from `platform/fabric/sdk/dig/sdk.go`). That made `MinimalFSCFabricConfig` nodes unable to even boot, since minimal mode's `core.yaml` has no networks at all.

- `platform/fabric/core/config.go`: `NewConfig`/`loadNetworks` now accept zero networks as a valid starting state — `Names()` is empty and `DefaultName()` is `""` until a network is added.
- `platform/fabric/sdk/dig/sdk.go`: `registerProcessorsForDrivers` now no-ops (with a log line) instead of erroring when zero networks are configured at startup.
- `platform/fabric/core.FSNProvider.AddNetwork` (already existing from #1603) is how networks get added after boot — this PR is what makes booting with zero networks (and thus later calling `AddNetwork`) actually work end to end.

**New integration test: `integration/fabric/iou_runtime_config/`**

A copy of `integration/fabric/iou/`, modified to demonstrate the full runtime-injection flow:

1. FSC nodes boot with `Topology.MinimalFSCFabricConfig` enabled — only `fabric:\n  enabled: true` in `core.yaml`.
2. `commands.InjectAll` renders each FSC peer's full extension via `Network.RenderFSCFabricExtension` and calls a new `"inject"` view (`views/inject.go`) on every running node.
3. `InjectNetworkView` calls `FSNProvider.AddNetwork(raw)` and then performs the additional wiring the Fabric SDK normally does at boot for networks known at startup: installing the default RWSet processor, registering the per-channel endorser-transaction RWSet handler, and starting each channel's committer and delivery service.
4. The standard IOU create/update/check flow then runs exactly as in `integration/fabric/iou/`.

Covers the `NoReplication` variants: LibP2P+TLS, WebSocket+TLS, WebSocket without TLS (mirroring `iou`'s T1–T3). The replica variant is deferred — `InjectAll` currently targets FSC nodes by their base name, not per-replica client keys.

Wired into `Makefile` (`fabric-iou_runtime_config` integration target) and `.github/workflows/tests.yml` (added to the monitoring-image-setup condition, since the topology enables Prometheus/Grafana/OTLP like `fabric-iou` does). It's automatically picked up by the CI matrix via `make list-integration-tests`.

### Testing

- `go build ./...` and `go vet ./...` — clean, both in the root module and `integration/`.
- Unit tests:
  - `integration/nwo/fabric/network/minimal_fsc_config_test.go`: `TestGenerateCoreConfig_MinimalFSCFabricConfig` (minimal mode emits exactly `fabric:\n  enabled: true`) and `TestGenerateCoreConfig_DefaultFSCFabricConfig` (default behavior unchanged).
  - `platform/fabric/core/config_test.go`: `TestNewConfig_EmptyIsAllowed`, `TestAddNetwork_IntoEmptyConfig`.
  - `platform/fabric/sdk/dig/sdk_test.go`: `TestRegisterProcessorsForDrivers_NoOpOnEmptyConfig`.
  - `go test -race -cover ./...` (relevant packages) — all pass.
- Integration: `integration/fabric/iou_runtime_config` run live against Fabric 3.1.4 + Docker for all three `NoReplication` variants (LibP2P, WebSocket+TLS, WebSocket no-TLS) — all pass, confirming the runtime-injected network can actually process the IOU flow (endorsement, commit, delivery, state queries) with no configuration baked in at node startup.


mbrandenburger (approved):
Thank you @adecaro. The suggested changes seem to look fine - it would be nice to have this feature used in one of our integration tests; for instance, to test the live re-configuration of a node - but this will be another work stream.

adecaro:
@mbrandenburger , if we merge https://github.com/hyperledger-labs/fabric-smart-client/pull/1603 first, then I can add the test in this PR as well. What do you think?

integration/fabric/iou_runtime_config/iou_test.go
```diff
@@ -0,0 +1,69 @@
...
+		AfterEach(s.TearDown)
+		It("succeeded", s.TestSucceeded)
+	})
+
+	Describe("IOU Life Cycle With Websockets and no TLS", Label("T3"), func() {
+		s := NewTestSuite(fsc.WebSocket, integration.NoReplication, false)
+		BeforeEach(s.Setup)
+		AfterEach(s.TearDown)
+		It("succeeded", s.TestSucceeded)
+	})
```
mbrandenburger: I think for the config test we can run this with just the p2p websocket stack - because we are injecting fabric platform re-configuration.

mbrandenburger:
```suggestion
	Describe("IOU Reconfiguration at Runtime With Websockets", Label("T1"), func() {
		s := NewTestSuite(fsc.WebSocket, integration.NoReplication, true)
		BeforeEach(s.Setup)
		AfterEach(s.TearDown)
		It("succeeded", s.TestSucceeded)
	})
```

integration/fabric/runtimeconfig/topology.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/tracing"
+)
+
+type Opts struct {
+	CommType        fsc.P2PCommunicationType
+	ReplicationOpts *integration.ReplicationOptions
+	TLSEnabled      bool
+}
+
+func Topology(opts *Opts) []api.Topology {
```
mbrandenburger, line 29:
```suggestion
// Topology create the "classic" IOU topology and additionally registers the `InjectNetworkViewFactory` for runtime reconfiguration.
func Topology(opts *Opts) []api.Topology {
```

mbrandenburger (changes requested):
Thank you @adecaro - adding the integration tests are very good. Please see my comments below.

integration/fabric/iou_runtime_config/topology.go
```diff
@@ -0,0 +1,107 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package iou_runtime_config
```
mbrandenburger: a future linter might complain about this package name. Looking at our other integration tests - we do not use underscores. We could name the package and itest just `runtimeconfig` (just dropping IOU).

integration/fabric/iou_runtime_config/iou_test.go
adecaro: Good call — collapsed the suite to just the WebSocket case in aedd200. The other two P2P variants weren't adding coverage for the injection path itself.

adecaro: Applied verbatim in aedd200.

integration/fabric/runtimeconfig/topology.go
adecaro: Added in aedd200.

integration/fabric/iou_runtime_config/topology.go
adecaro: Renamed the package/dir to `runtimeconfig` (dropped the underscore and the `iou` prefix) in aedd200 — also updated the Makefile target, CI matrix entry, and the port constant accordingly.

mbrandenburger (approved):
Thank you @adecaro

adecaro:
Hi @mbrandenburger , can I merge this?

PR #1608, opened 2026-07-24
feat(logging): extract context values into context-aware log lines as zap fields

### LINKED ISSUE #1607: feat(logging): extract context values into context-aware log lines as zap fields
## Problem Description

`platform/common/services/logging` exposes context-aware log methods
(`InfowContext`, `ErrorfContext`, `DebugfContext`, … — 10 in total via the
`otelLogger` interface) backed by `otelzap.SugaredLogger`. Today the
`context.Context` passed to these methods is only used to emit an
OpenTelemetry span event — nothing carried in the context (request ID,
tenant, channel, tx ID, etc.) ever makes it into the zap log line itself.
Callers who want that correlation data on every log line currently have to
thread it through manually via `.With(...)` at every call site, which is
easy to forget and inconsistent across the codebase.

## Proposed Solution

Add a declarative `(context key → zap field name)` mechanism so that any
value present in the context is automatically extracted and attached to the
log line (`zap.Any(name, value)`) whenever a `*Context` method is called,
with no per-call-site plumbing required.

Configuration surface, mirroring the existing `RegisterReplacer`/`Replacers()`
pattern in `config.go`:

```go
// Register once, e.g. in an init() or bootstrap path.
logging.RegisterContextLogField("request.id", requestIDKey)

// Or via bulk config:
logging.Init(logging.Config{
    ContextLogFields: []logging.ContextLogField{
        {Key: requestIDKey, Name: "request.id"},
        {Key: tenantKey, Name: "tenant"},
    },
})
```

From then on:

```go
ctx := context.WithValue(context.Background(), requestIDKey, "abc-123")
logger.InfowContext(ctx, "handling request") // log line now includes request.id=abc-123
```

Values are logged as-is via `zap.Any` (no implicit sanitization — that stays
scoped to `OtelSanitize`'s existing span-event-body use). Only a declarative
list is supported (no custom extractor-function escape hatch), keeping the
surface small and the cost of a disabled log call predictable.

Because context-field extraction sits on a very hot path (every context-aware
log call in the process, including ones that are disabled by level), the
decorator must level-gate before touching the registry or context at all, and
avoid allocating an extra logger clone on the enabled `*w*Context` path by
appending fields directly into the `keysAndValues` slice instead of calling
`.With()`.

## Alternatives Considered

- A custom extractor-function escape hatch (`func(ctx) []zap.Field`) instead
  of a declarative list — rejected to keep the API surface small and the
  performance characteristics easy to reason about.
- Sanitizing extracted values by default (like `OtelSanitize`) — rejected;
  values are logged as-is, sanitization stays opt-in/out of scope.

## Implementation Notes

- `platform/common/services/logging/config.go` — `ContextLogField` type,
  `ctxLogFields` registry (`sync.RWMutex`-guarded, mirrors `replacers`),
  `RegisterContextLogField`/`ContextLogFields()`, `Config.ContextLogFields`
  merged idempotently in `Init`.
- `platform/common/services/logging/otel.go` — `ctxFieldLogger` decorator
  around `*otelzap.SugaredLogger` implementing all 10 `*Context` methods;
  level-gates via `SugaredLogger.Level()` before extracting; `*w*Context`
  methods append fields into `keysAndValues` instead of calling `.With()`.
- `platform/common/services/logging/logger.go` — caller-skip bump to keep
  logged caller/file/line accurate through the new decorator layer.

adecaro:
## Summary

- Adds a declarative `(context key -> zap field name)` mechanism so values
  carried in a `context.Context` (request ID, tenant, tx ID, etc.) are
  automatically attached as `zap.Any` fields to every context-aware log call
  (`InfowContext`, `ErrorfContext`, ... — 10 methods total), instead of only
  being used to emit an OTel span event as today.
- Dual configuration surface mirroring the existing `Replacers` pattern:
  `logging.Init(Config{ContextLogFields: [...]})` for bulk config, and
  `logging.RegisterContextLogField(name, key)` for direct registration
  (panics on duplicate name, same as `RegisterReplacer`).
- Values are logged as-is via `zap.Any` — no implicit sanitization; that
  stays scoped to `OtelSanitize`'s existing span-event-body use.
- Performance: the decorator (`ctxFieldLogger` in `otel.go`) level-gates via
  `SugaredLogger.Level()` before touching the registry or the context at
  all, so disabled log calls stay near-zero-cost regardless of how many
  fields are registered anywhere in the process (measured: ~577ns/12-alloc
  regression without the gate vs. ~50ns/1-alloc with it, matching the
  no-fields-registered baseline). The `*w*Context` methods append extracted
  fields directly into the `keysAndValues` slice instead of calling
  `.With()` first, avoiding an extra `SugaredLogger` clone allocation on the
  enabled path.
- Caller-skip bumped by one in `logger.go` to compensate for the new
  decorator frame, so logged caller/file/line still point at the true call
  site (covered by `TestContextLogFields_CallerAccuracy`).

Closes #1607.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./platform/common/services/logging/...`
- [x] `go test ./platform/common/services/logging/...` (new
      `context_fields_test.go`: extraction, absent key, nil context,
      multi-key, survives `With`/`Named`, duplicate-registration panic,
      `Init` idempotency, caller accuracy — all pass)
- [x] `go test ./platform/common/core/generic/committer/... ./platform/common/core/generic/vault/...`
      (spot-check downstream consumers of context-aware logging — no
      behavior change)
- [x] Isolated benchmarks (`bench_test.go`) confirm the disabled-log-call
      cost is independent of registry size after the level-gate fix

### DIFF
--- a/platform/common/services/logging/config.go
+++ b/platform/common/services/logging/config.go
@@ -36,6 +36,20 @@ type Config struct {
 	// Sanitization means that any non-printable character is removed.
 	// This protects the traces from undesired behaviours like missing tracing.
 	OtelSanitize bool
+
+	// ContextLogFields lists context keys to extract as zap fields on every context-aware
+	// log call (e.g. InfowContext, ErrorfContext). Merged into the global registry; see
+	// RegisterContextLogField.
+	ContextLogFields []ContextLogField
+}
+
+// ContextLogField maps a context key to the zap field name used when the value is found
+// in the context and added to a log line.
+type ContextLogField struct {
+	// Key is looked up via ctx.Value(Key).
+	Key any
+	// Name is the zap field name the value is logged under.
+	Name string
 }
 
 var (
@@ -52,8 +66,12 @@ func Init(c Config) {
 
 	// set local configurations
 	configMutex.Lock()
-	defer configMutex.Unlock()
 	config.OtelSanitize = c.OtelSanitize
+	configMutex.Unlock()
+
+	for _, f := range c.ContextLogFields {
+		registerContextLogField(f.Name, f.Key, true)
+	}
 }
 
 func OtelSanitize() bool {
@@ -87,4 +105,46 @@ func Replacers() map[string]string {
 	replacersMutex.RLock()
 	defer replacersMutex.RUnlock()
 	return replacers
+}
+
+var (
+	ctxFieldsMutex sync.RWMutex
+	ctxFieldNames  []string
+	ctxLogFields   = map[string]any{}
+)
+
+// RegisterContextLogField registers a context key to extract as a zap field, under the
+// given field name, on every context-aware log call. Panics if name is already
+// registered.
+func RegisterContextLogField(name string, key any) {
+	registerContextLogField(name, key, false)
+}
+
+func registerContextLogField(name string, key any, overwrite bool) {
+	ctxFieldsMutex.Lock()
+	defer ctxFieldsMutex.Unlock()
+
+	if _, ok := ctxLogFields[name]; ok {
+		if overwrite {
+			ctxLogFields[name] = key
+			return
+		}
+		panic("context log field already exists")
+	}
+
+	ctxLogFields[name] = key
+	ctxFieldNames = append(ctxFieldNames, name)
+}
+
+// ContextLogFields returns the current context log field registrations, in
+// registration order.
+func ContextLogFields() []ContextLogField {
+	ctxFieldsMutex.RLock()
+	defer ctxFieldsMutex.RUnlock()
+
+	fields := make([]ContextLogField, len(ctxFieldNames))
+	for i, name := range ctxFieldNames {
+		fields[i] = ContextLogField{Key: ctxLogFields[name], Name: name}
+	}
+	return fields
 }
--- a/platform/common/services/logging/config.go
+++ b/platform/common/services/logging/config.go
@@ -168,6 +168,8 @@ func ContextLogFields() []ContextLogField {
 // each test start from an empty registry (e.g. via t.Cleanup) instead of relying on
 // globally-unique field names to avoid colliding with registrations left behind by
 // other tests.
+//
+//lint:ignore U1000 used for testing
 func resetContextLogFields() {
 	ctxFieldsMutex.Lock()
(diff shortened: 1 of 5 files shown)


platform/common/services/logging/config.go
```diff
@@ -88,3 +107,52 @@ func Replacers() map[string]string {
...
+)
+
+// RegisterContextLogField registers a context key to extract as a zap field, under the
+// given field name, on every context-aware log call. Panics if name is already
+// registered.
+func RegisterContextLogField(name string, key any) {
+	registerContextLogField(name, key, false)
+}
+
+func registerContextLogField(name string, key any, overwrite bool) {
```
mbrandenburger, line 136: for better testing we could add an unexported `resetContextLogFields()` that allows to call `t.Cleanup` and thus clean the global state. WDYT?

platform/common/services/logging/config.go
```diff
@@ -88,3 +107,52 @@ func Replacers() map[string]string {
...
+	ctxFieldsMutex.Lock()
+	defer ctxFieldsMutex.Unlock()
+
+	if _, ok := ctxLogFields[name]; ok {
+		if !overwrite {
+			panic("context log field already exists")
+		}
```
mbrandenburger, line 143: I am wondering if the `overwrite` + `panic` are the right way for the API. In general a panic here should show a developer error, which should be fixed. What is the case when we want explicit overwrite - just testing?. Note that the behavior of `RegisterContextLogField` with with `overwrite` is not documented. If just for testing - I suggest to go with the "reset" impl as suggested above. WDYT?

mbrandenburger (commented):
Thank you @adecaro. I have a minor suggestion below regarding testing - please see my comments below.

platform/common/services/logging/config.go
adecaro: So, I have introduced `resetContextLogFields` for the tests. Regarding the panic, it is the same behavior that 
`database/sql` has. I think we should be good there then.

platform/common/services/logging/config.go
adecaro: done

mbrandenburger (approved):

PR #1611, opened 2026-07-26
fix(security): harden fabric/core against forged gossip discovery, unrecovered panics, and silent failure paths

### LINKED ISSUE #1610: Harden platform/fabric/core against forged gossip discovery and unrecovered panics
## Summary

A security review of `platform/fabric/core/` (code paths reached when this node consumes untrusted input from a peer or orderer) identified several hardening opportunities. Fixes have been implemented locally and are pending submission as a PR.

## Areas addressed

- **Gossip discovery envelope trust**: Alive/StateInfo envelopes returned by the Discovery service were accepted with no signature verification, so a malicious/compromised discovery-responding peer could forge another org's peer membership info (endpoint, MSPID) and have it treated as authentic.
- **Unrecovered panics on malformed peer input**: several call sites indexed directly into block/block-metadata fields populated by a remote peer, with no nil/bounds checks and no panic recovery anywhere up the call chain — a malformed block could crash the entire process rather than just failing the affected request.
- **Panic on a recoverable failure path**: a membership-update failure during config-block commit caused a `panic` instead of returning an error, even though the block had already been durably committed.
- **Silent failure reporting**: one broadcaster path could report success back to the caller even when the orderer explicitly rejected the transaction, due to an error-wrapping call that returns `nil` when given a `nil` underlying error.
- **Undocumented trust boundary**: a lightweight identity-deserialization fallback path accepts any well-formed self-signed identity by design (intentional degraded-mode behavior), but this trust boundary was not documented, risking a future caller mistakenly treating "signature verifies" as "identity is authorized."

## Status

Fixes for all of the above are implemented and covered by new/updated unit tests, with a clean `go build`/`go test` across the affected package trees. A PR will follow shortly.

## Scope note

Exploit-level detail (multi-step attack chains) is intentionally omitted from this public issue; happy to share the full write-up with maintainers privately if useful for review.

adecaro:
## Summary

Hardens `platform/fabric/core/` against several issues reachable from untrusted peer/orderer input:

- **Gossip discovery envelope trust**: `peersForChannel`/`endorser` now verify each Alive/StateInfo envelope's signature against the public key embedded in that peer's own claimed identity before trusting its content. Previously `EnvelopeToGossipMessage` unmarshaled `Payload` without ever checking `Signature`, so a malicious/compromised discovery-responding peer could forge another org's peer membership (endpoint, MSPID) and have it accepted — a self-consistency check now closes the "arbitrary/garbage signature accepted" gap.
- **Unrecovered panics on malformed peer input**: four independent call sites indexed directly into `*common.Block`/`common.BlockMetadata` fields from a remote `Deliver`/`DeliverFiltered` stream with no nil/bounds checks and no panic recovery anywhere up the call chain. A malformed block from a peer could crash the entire process. All four now guard and return errors instead.
- **Panic on a recoverable failure**: `CommitConfig` `panic`ked on a `MembershipService.Update` failure even after the block was already durably committed. Now returns a wrapped error.
- **Silent success on orderer rejection**: `CFTBroadcaster.Broadcast` called `errors.Wrapf(err, ...)` on an orderer-rejected status where the transport `err` was `nil` — per `github.com/pkg/errors` semantics, wrapping a `nil` error returns `nil`, so a rejected transaction was reported back to the caller as a success. Switched to `errors.Errorf` so rejection is always a non-nil error.
- **Undocumented trust boundary**: the lightweight x509 identity deserializers (`msp/x509.Provider.DeserializeVerifier`, `view/services/id/x509.Deserializer.DeserializeVerifier`) accept any well-formed self-signed identity with no CA/MSP binding — an intentional degraded-mode fallback, not a bug to remove (it backs a tested "verify only" contract). Replaced a stale `// TODO: check the validity of the identity against the msp` with an explicit doc comment on both implementations spelling out the boundary, so callers know to check results against an allow-list rather than treating "signature verifies" as "identity is authorized." Documentation-only, no behavior change.

Fixes #1610.

## Test plan

- [x] `go build ./platform/fabric/... ./platform/view/...`
- [x] `go test ./platform/fabric/... ./platform/view/...` (full tree, zero failures)
- [x] `go test ./platform/fabric/core/generic/discovery/...` — new `TestForgedGossipEnvelopeRejected` (garbage signature, wrong-key signature, and genuine-signature-accepted subtests) plus existing `TestClient`/`TestBadResponses` continue to pass against real signed fixtures
- [x] `go test ./platform/fabric/core/generic/delivery/...` — new `deliverclient_test.go`/`service_test.go`, extended `delivery_test.go`
- [x] `go test ./platform/fabric/core/generic/committer/...` — extended `config_test.go`/`endorsertx_test.go`
- [x] `go test ./platform/fabric/core/generic/ledger/...` — extended `ledger_test.go`
- [x] `go test ./platform/fabric/core/generic/ordering/...` — new `TestCFTBroadcaster_Broadcast_OrdererRejection`
- [x] `go test ./platform/fabric/core/generic/msp/x509/... ./platform/view/services/id/x509/...` — confirmed doc-only change, existing verify-only fallback tests unchanged

### DIFF
--- a/platform/fabric/core/generic/committer/config.go
+++ b/platform/fabric/core/generic/committer/config.go
@@ -146,8 +146,7 @@ func (c *Committer) CommitConfig(ctx context.Context, blockNumber driver.BlockNu
 
 	// once committed, we can update the membership service
 	if err := c.MembershipService.Update(env); err != nil {
-		// this should not have happened
-		panic(err)
+		return errors.Wrapf(err, "failed updating membership service for configtx [%s]", txID)
 	}
 
 	// and apply other updates
--- a/platform/fabric/core/generic/discovery/client.go
+++ b/platform/fabric/core/generic/discovery/client.go
@@ -385,12 +385,18 @@ func peersForChannel(membersRes *discovery.PeerMembershipResult, qt QueryType) (
 	var peers []*Peer
 	for org, peersOfCurrentOrg := range membersRes.PeersByOrg {
 		for _, pp := range peersOfCurrentOrg.Peers {
+			if err := verifyEnvelopeSignature(pp.MembershipInfo, pp.Identity); err != nil {
+				return nil, errors.Wrap(err, "failed verifying alive message signature")
+			}
 			aliveMsg, err := EnvelopeToGossipMessage(pp.MembershipInfo)
 			if err != nil {
 				return nil, errors.Wrap(err, "failed unmarshalling alive message")
 			}
 			var stateInfoMsg *SignedGossipMessage
 			if isStateInfoExpected(qt) {
+				if err := verifyEnvelopeSignature(pp.StateInfo, pp.Identity); err != nil {
+					return nil, errors.Wrap(err, "failed verifying stateInfo message signature")
+				}
 				stateInfoMsg, err = EnvelopeToGossipMessage(pp.StateInfo)
 				if err != nil {
 					return nil, errors.Wrap(err, "failed unmarshalling stateInfo message")
@@ -503,6 +509,12 @@ func (resp response) createEndorsementDescriptor(desc *discovery.EndorsementDesc
 func endorser(peer *discovery.Peer, chaincode, channel string) (*Peer, error) {
 	if peer.MembershipInfo == nil || peer.StateInfo == nil {
 		return nil, errors.Errorf("received empty envelope(s) for endorsers for chaincode %s, channel %s", chaincode, channel)
+	}
+	if err := verifyEnvelopeSignature(peer.MembershipInfo, peer.Identity); err != nil {
+		return nil, errors.Wrap(err, "failed verifying alive message signature")
+	}
+	if err := verifyEnvelopeSignature(peer.StateInfo, peer.Identity); err != nil {
+		return nil, errors.Wrap(err, "failed verifying stateInfo message signature")
 	}
 	aliveMsg, err := EnvelopeToGossipMessage(peer.MembershipInfo)
 	if err != nil {
--- a/platform/fabric/core/generic/discovery/client.go
+++ b/platform/fabric/core/generic/discovery/client.go
@@ -385,7 +385,7 @@ func peersForChannel(membersRes *discovery.PeerMembershipResult, qt QueryType) (
 	var peers []*Peer
 	for org, peersOfCurrentOrg := range membersRes.PeersByOrg {
 		for _, pp := range peersOfCurrentOrg.Peers {
-			if err := verifyEnvelopeSignature(pp.MembershipInfo, pp.Identity); err != nil {
+			if err := verifyEnvelopeSignature(pp.MembershipInfo, pp.Identity, true); err != nil {
 				return nil, errors.Wrap(err, "failed verifying alive message signature")
 			}
 			aliveMsg, err := EnvelopeToGossipMessage(pp.MembershipInfo)
@@ -394,7 +394,7 @@ func peersForChannel(membersRes *discovery.PeerMembershipResult, qt QueryType) (
 			}
 			var stateInfoMsg *SignedGossipMessage
 			if isStateInfoExpected(qt) {
-				if err := verifyEnvelopeSignature(pp.StateInfo, pp.Identity); err != nil {
+				if err := verifyEnvelopeSignature(pp.StateInfo, pp.Identity, false); err != nil {
 					return nil, errors.Wrap(err, "failed verifying stateInfo message signature")
 				}
 				stateInfoMsg, err = EnvelopeToGossipMessage(pp.StateInfo)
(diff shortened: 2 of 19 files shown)


platform/fabric/core/generic/discovery/gossip.go
```diff
@@ -68,3 +69,43 @@ func EnvelopeToGossipMessage(e *gossip.Envelope) (*SignedGossipMessage, error) {
...
+// upstream Fabric's own discovery client performs no verification at all on
+// these envelopes - so rejecting a nil signature here would fail every
+// Discovery query in which the responding peer is itself among the reported
+// peers/endorsers, which is the common case in small networks. StateInfo
+// self-entries do not share this exemption: real peers always produce a
+// genuine signature for them, even for themselves (see
+// gossipChannel.setupSignedStateInfoMessage, which calls a real signer, not
+// NoopSign), so callers must pass false for StateInfo envelopes to keep that
+// verification strict.
+func verifyEnvelopeSignature(e *gossip.Envelope, identity []byte, allowEmptySignature bool) error {
```
mbrandenburger: I am wondering if it would make sense to move this function closer to the code where it is used - that is - into client.go. No strict request to move it - just sharing my initial thought.

platform/fabric/core/generic/ledger/ledger.go
```diff
@@ -122,13 +122,24 @@ func (b *Block) DataAt(i int) []byte {
 
 // ProcessedTransaction returns the ProcessedTransaction at passed index
 func (b *Block) ProcessedTransaction(i int) (driver.ProcessedTransaction, error) {
+	if i < 0 || i >= len(b.Data.Data) {
```
mbrandenburger: We should add additional checks for `b.Data == nil` and `b.Metadata == nil` as you also added in `delivery/service.go`

```go
if b.Data == nil || b.Metadata == nil { return nil, errors.Errorf(...) }
```

platform/fabric/core/generic/discovery/client.go
```diff
@@ -385,12 +385,18 @@ func peersForChannel(membersRes *discovery.PeerMembershipResult, qt QueryType) (
 	var peers []*Peer
 	for org, peersOfCurrentOrg := range membersRes.PeersByOrg {
 		for _, pp := range peersOfCurrentOrg.Peers {
+			if err := verifyEnvelopeSignature(pp.MembershipInfo, pp.Identity, true); err != nil {
```
mbrandenburger, line 444: While this is a great step forward - I am wondering if we also should/must validate `pp.Identity` against the MSP somewhere - or this is check implicitly somewhere happening already? If we don't want this MSP-validation here, we better add another comment in the doc. WDYT?

mbrandenburger (changes requested):
Thank you @adecaro for working on this! Solid PR. Please see my comments below.

platform/fabric/core/generic/discovery/gossip.go
adecaro: Done — moved `verifyEnvelopeSignature` from `gossip.go` into `client.go`, right before `peersForChannel` (its first caller). See f500f391.

platform/fabric/core/generic/ledger/ledger.go
adecaro: Done — added `b.Data == nil` / `b.Metadata == nil` guards to `ProcessedTransaction`, and also fixed `DataAt` in the same file, which had the identical unguarded `b.Data.Data[i]` index (no nil check, no bounds check at all) — matching the fix already applied to the fabricx driver's `Block.DataAt`. See 906eb02e.

platform/fabric/core/generic/discovery/client.go
adecaro: Good catch, and no, this isn't happening implicitly elsewhere — `verifyEnvelopeSignature` only proves the envelope was signed by the key in `pp.Identity` itself (self-consistency), not that `pp.Identity` is a CA-issued identity trusted by the channel/local MSP. Under the hood it uses the same lightweight, no-CA-binding provider (`mspx509.NewIdentityFromBytes`) as `msp/x509.Provider.DeserializeVerifier`, which this PR already documents as an intentional "verify only" fallback elsewhere.

Went with your doc-only fallback for this PR — added an explicit doc comment on `verifyEnvelopeSignature` spelling out that boundary (same treatment as the x509 deserializers) rather than trying to bolt on MSP/CA validation here, since the right trust anchor for that (local MSP config vs. the same discovery response's `ConfigResult`, which isn't independently trusted either) is an architectural call bigger than this hardening patch. Filed #1623 to track it as a follow-up. See f500f391.

platform/fabric/core/generic/committer/config.go
```diff
@@ -146,8 +146,7 @@ func (c *Committer) CommitConfig(ctx context.Context, blockNumber driver.BlockNu
 
 	// once committed, we can update the membership service
 	if err := c.MembershipService.Update(env); err != nil {
-		// this should not have happened
-		panic(err)
+		return errors.Wrapf(err, "failed updating membership service for configtx [%s]", txID)
```
mbrandenburger, line 149: what is our strategy here to update the membership service eventually? With the panic - we killed the FSC node - not good at all - but very obvious that something serious happened. If we not return an error, we need to handle this situation accordingly - retry? partially apply? do something? otherwise the FSC node continues to run with a stale membership service - which may also prevent any "progress".

I have the feeling that this strategy may require a follow up PR. WDYT?

mbrandenburger (approved):

platform/fabric/core/generic/committer/config.go
adecaro: indeed, we need to distinguish what was the reason for the failure. Some might require the panic of the node, some other just a retry. I'll open an Issue.

PR #1613, opened 2026-07-26
fix(security): harden fabricx driver against malformed/malicious remote input

### LINKED ISSUE #1612: Harden platform/fabricx/ against malformed/malicious remote input
## Summary

A security review of `platform/fabricx/` (the Hyperledger Fabric-X driver) identified several hardening opportunities in endorsement verification, notification-stream handling, ledger block parsing, channel-config propagation, and transaction-commit handling. Fixes have been implemented locally and are pending submission as a PR.

## Areas addressed

- **Endorsement verification (`core/transaction`)**: signature verification did not fully account for multi-entry endorsement sets or cross-check the identity claimed by an endorsement against the actual signer.
- **Notification stream handling (`core/finality`)**: a lock could remain held indefinitely across a blocking channel operation once the underlying notification stream failed, affecting availability of finality tracking.
- **Ledger block parsing (`core/ledger`)**: block and transaction-metadata fields returned by a remote committer were dereferenced/indexed without validating they were present or within bounds first.
- **Channel-config propagation (`core/channel/config`)**: a server-reported version counter was trusted in isolation to decide whether a new channel configuration should be applied, without cross-validating it against the configuration content itself.
- **Transaction-commit handling (`core/committer`)**: a transaction-status lookup indexed a metadata slice without fully validating its bounds against the transaction index first.

## Status

Fixes for all of the above are implemented and covered by new/updated unit tests (all passing, `go vet` clean, `-race` clean). A PR will follow shortly.

## Scope note

Exploit-level detail is intentionally omitted from this public issue; happy to share the full write-up with maintainers privately if useful for review.

adecaro:
## Summary

Hardens `platform/fabricx/` — the Hyperledger Fabric-X driver — against
malformed or malicious input from remote endpoints (endorsing peers,
committers, notifiers). Several code paths trusted remote-controlled data
without sufficient validation, allowing a compromised or misbehaving remote
endpoint to bypass endorsement verification, deadlock finality tracking,
crash the client process, or suppress legitimate channel-config propagation.

- **`core/transaction`**: `VerifyEndorsement` now verifies every
  `EndorsementWithIdentity` entry in a namespace's endorsement set (not just
  the first), and cross-checks each entry's claimed identity against the
  actual endorser rather than accepting it at face value.
- **`core/finality`**: `notificationListenerManager` no longer holds
  `handlersMu` across a blocking send on `requestQueue`. Once the underlying
  notification stream dies, `AddFinalityListener` now fails fast with an
  error instead of deadlocking every subsequent listener registration for
  any transaction.
- **`core/ledger`**: block header, block data, and the
  `TRANSACTIONS_FILTER` metadata entry returned by the remote committer are
  now nil/bounds-checked before being dereferenced or indexed.
- **`core/channel/config`**: `ChannelConfigMonitor` no longer trusts the
  remote-reported config `Version` counter in isolation. It now also
  cross-validates the sequence number embedded in the config envelope's own
  content, so a compromised committer can't suppress a genuinely new config
  merely by lying about `Version`.
- **`core/committer`**: `HandleFabricxTransaction` now bounds-checks the
  `TRANSACTIONS_FILTER` entry against the transaction index (including an
  off-by-one in the outer length check that only guaranteed indices before
  the filter entry existed, not the entry itself) before indexing into it.

All previously-panicking/bypassing behavior is now covered by regression
tests asserting the fixed behavior in place of the old vulnerable/buggy
assertions.

Fixes #1612.

## Test plan

- [x] `go build ./platform/fabricx/...`
- [x] `go vet ./platform/fabricx/...`
- [x] `go test ./platform/fabricx/... -race` — all packages pass
- [x] Isolated `-run` of each new/updated regression test individually

### DIFF
--- a/platform/fabricx/core/channel/config/monitor.go
+++ b/platform/fabricx/core/channel/config/monitor.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/protoutil"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabricx/core/committer/queryservice"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc"
@@ -60,12 +61,13 @@ type ChannelConfigMonitor struct {
 	channel           string
 
 	// State management
-	mu          sync.RWMutex
-	running     bool
-	ctx         context.Context
-	cancel      context.CancelFunc
-	wg          sync.WaitGroup
-	lastVersion int64
+	mu           sync.RWMutex
+	running      bool
+	ctx          context.Context
+	cancel       context.CancelFunc
+	wg           sync.WaitGroup
+	lastVersion  int64
+	lastSequence int64
 }
 
 // NewChannelConfigMonitor creates a new ChannelConfigMonitor service instance
@@ -107,6 +109,7 @@ func NewChannelConfigMonitor(
 		channel:           channel,
 		running:           false,
 		lastVersion:       -1,
+		lastSequence:      -1,
 	}, nil
 }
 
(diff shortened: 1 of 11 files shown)


mbrandenburger (approved):
Thank you @adecaro for working on this! Well done!

PR #1618, opened 2026-07-27
fix(fabricx): stop retaining RWSets and tx statuses in the vault

### LINKED ISSUE #1599: Investigate: RWSet  retained and grow unbounded on endorsement-only nodes
### What happened?

Under sustained high-throughput load (hundreds of tx/s over several days), an
**endorsement/approval** node's memory grew to ~12 GB and did **not** free after the load
stopped. A heap profile attributes ~10.75 GB almost entirely to read/write-set construction,
and the memory is **in-use** (`inuse_space`), i.e. held live during processing — not garbage
awaiting GC.

### How can we reproduce it?

TBD

### Environment

FSC

### Additional Information

_No response_

SaidAltury-ibm:
closes #1599
The FabricX vault held two in-memory maps (rwsets, txStatuses) that grew
unboundedly: nothing on the FabricX path ever pruned or read them. FabricX has
no local commit pipeline — commit/validation happens in the external
fabric-x-committer, and finality is delivered by the gRPC notifier — so the
vault methods that populated or read those maps (CommitTX/DiscardTx/
SetDiscarded/Match/RWSExists) are never invoked in the FabricX wiring. Endorse-
only nodes were pinning ~1.5 GB of RWSets at rest as a result (PR #1206).

Changes:
- Remove the rwsets and txStatuses maps, the txStatusInfo struct, and the mutex.
  NewRWSet/NewRWSetFromBytes no longer retain the RWSet; the wrapper owns it.
- The five commit-pipeline interface methods now panic instead of returning a
  value. They exist only to satisfy fdriver.Vault; reaching one means the vault
  was wired into a generic committer by mistake, which is a programming error,
  not a runtime condition. Panicking fails loudly and cannot be silently
  swallowed — RWSExists in particular has no error channel, so its old `false`
  return was an actively misleading answer.
- Status now delegates to the batched Statuses so both report a tx unknown to
  the committer as Unknown ("not final yet") rather than as an error — the
  condition a caller polling an in-flight tx will hit.
- Bytes() passes namespace-version info to the marshaller per call instead of
  through a shared mutable field, fixing a data race between concurrent Bytes()
  calls. The marshaller is now stateless; the racy field and its exported
  wrapper are removed. Covered by a new -race test.

### DIFF
--- a/platform/fabricx/core/vault/vault.go
+++ b/platform/fabricx/core/vault/vault.go
@@ -8,7 +8,6 @@ package vault
 
 import (
 	"context"
-	"sync"
 
 	"github.com/hyperledger/fabric-x-common/api/committerpb"
 
@@ -19,31 +18,13 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabricx/core/committer/queryservice"
 )
 
-// txStatusInfo holds local transaction status information including validation code,
-// error message, and block/transaction index for committed transactions.
-type txStatusInfo struct {
-	code    fdriver.ValidationCode // Validation code (Valid, Invalid, Unknown, etc.)
-	message string                 // Error or status message
-	block   cdriver.BlockNum       // Block number where transaction was committed
-	index   cdriver.TxNum          // Transaction index within the block
-}
-
-// Vault is a vault implementation that uses QueryService for remote state queries
-// and manages read-write sets (RWSets) and transaction statuses locally.
-//
-// It provides a hybrid architecture where:
-//   - Read operations are delegated to a remote QueryService
-//   - RWSets are managed in-memory for performance
-//   - Transaction statuses are cached locally with remote fallback
-//   - All operations are thread-safe using RWMutex
-//
-// Vault implements the fdriver.Vault interface for FabricX.
+// Vault implements the fdriver.Vault interface for FabricX. Several interface methods
+// (CommitTX/DiscardTx/SetDiscarded/Match/RWSExists) exist only to satisfy that contract; FabricX
+// has no local commit pipeline, so they are never invoked in the FabricX wiring and panic if
+// reached (which would indicate the vault was wired into a generic committer by mistake).
 type Vault struct {
-	queryService queryservice.QueryService            // Remote query service for state queries
-	marshaller   *Marshaller                          // Marshaller for RWSet serialization
-	rwsets       map[cdriver.TxID]*vault.ReadWriteSet // Local RWSet storage
-	txStatuses   map[cdriver.TxID]*txStatusInfo       // Local transaction status cache
-	mu           sync.RWMutex                         // Mutex for thread-safe access
+	queryService queryservice.QueryService // Remote query service for state and status queries
+	marshaller   *Marshaller               // Marshaller for RWSet serialization
 }
 
 // NewVault creates a new Vault instance with the given QueryService.
@@ -58,8 +39,6 @@ func NewVault(qs queryservice.QueryService) *Vault {
 	return &Vault{
 		queryService: qs,
 		marshaller:   NewMarshaller(),
-		rwsets:       make(map[cdriver.TxID]*vault.ReadWriteSet),
-		txStatuses:   make(map[cdriver.TxID]*txStatusInfo),
 	}
 }
 
@@ -349,15 +328,13 @@ func (r *rwSetWrapper) Bytes() ([]byte, error) {
 		}
 	}
 
-	// Set nsInfo in marshaller before marshalling
-	r.v.marshaller.NsInfo = nsInfo
-
-	return r.v.marshaller.Marshal(string(r.txID), r.rws)
+	// Pass nsInfo per call rather than through shared marshaller state, so concurrent
+	// Bytes() calls on different RWSets from the same vault do not race.
+	return r.v.marshaller.Marshal(string(r.txID), r.rws, nsInfo)
 }
 
-// Done performs cleanup for this RWSet. Currently a no-op as no cleanup is needed.
+// Done is a no-op. The vault does not retain the ReadWriteSet.
 func (r *rwSetWrapper) Done() {
-	// No cleanup needed
 }
 
 // Equals compares this RWSet with another RWSet for equality.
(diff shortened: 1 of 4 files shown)


platform/fabricx/core/vault/vault.go
```diff
@@ -429,56 +397,25 @@ func (v *Vault) NewRWSetFromBytes(ctx context.Context, txID cdriver.TxID, rwset
...
-
-	return validationCode, "", nil
+	return v.mapStatusToValidationCode(statusCode), "", nil
 }
 
-// Statuses returns the validation statuses for multiple transactions.
-// Each transaction status is retrieved using the Status method.
+// Statuses returns the validation statuses for multiple transactions by delegating to Status for
+// each one (which queries the remote QueryService).
 func (v *Vault) Statuses(ctx context.Context, txIDs ...cdriver.TxID) ([]cdriver.TxValidationStatus[fdriver.ValidationCode], error) {
```
mbrandenburger, line 423: We should change the impl of this method to leverage the batched query call (`v.queryService.GetTransactionStatuses`) instead of doing the loop here.

See:
https://github.com/hyperledger-labs/fabric-smart-client/blob/6bf5010c0a7f374d004e1dd84e9bb7b8795cf49f/platform/fabricx/core/committer/queryservice/query.go#L100

platform/fabricx/core/vault/vault.go
mbrandenburger: I think we this is/was even wrong. Note that `v.queryService.GetTransactionStatus(string(txID))` returns an error if the committer does not know the transaction. Then `Status(ctx context.Context, txID cdriver.TxID)` catches that error and returns the correct status `fdriver.Unknown` + error. 

Here in this code, we drop the `fdriver.Unknown` code and just return an error - even though this is not an "error" situation. I believe a "clean" implementation using `v.queryService.GetTransactionStatuses` will simplify this code and make more clear.

platform/fabricx/core/vault/vault.go
```diff
@@ -429,56 +397,25 @@ func (v *Vault) NewRWSetFromBytes(ctx context.Context, txID cdriver.TxID, rwset
 	}, nil
 }
 
-// SetDiscarded marks a transaction as discarded (invalid) with the given error message.
-// The status is stored locally.
+// SetDiscarded exists only to satisfy the fdriver.Vault interface and is not implemented.
 func (v *Vault) SetDiscarded(ctx context.Context, txID cdriver.TxID, message string) error {
```
mbrandenburger:
```suggestion
func (v *Vault) SetDiscarded(_ context.Context, txID cdriver.TxID, message string) error {
```

platform/fabricx/core/vault/vault.go
```diff
@@ -355,9 +333,8 @@ func (r *rwSetWrapper) Bytes() ([]byte, error) {
 	return r.v.marshaller.Marshal(string(r.txID), r.rws)
```
mbrandenburger: I believe we should take the chance and fix this race error here as well. 

The issue is if two goroutines call `rwSetWrapper.Bytes()`, there will be a race condition on `r.v.marshaller.NsInfo`. Luckily, we can just use the prive `mashal` function and pass `nsInfo` directly.

```suggestion
return r.v.marshaller.marshal(r.txID, r.rws, nsInfo)
```

platform/fabricx/core/vault/vault.go
```diff
@@ -537,55 +463,18 @@ func (v *Vault) InspectRWSet(ctx context.Context, rwset []byte, namespaces ...cd
...
 
-// RWSExists checks whether an RWSet exists locally for the given transaction ID.
+// RWSExists exists only to satisfy the fdriver.Vault interface and always returns false.
 func (v *Vault) RWSExists(ctx context.Context, id cdriver.TxID) bool {
-	v.mu.RLock()
-	defer v.mu.RUnlock()
-	_, exists := v.rwsets[id]
-	return exists
+	return false
 }
```
mbrandenburger, line 490: I think we can be more aggressive here (and the other unsupported) methods. WDYT?

```go
// RWSExists is not supported: the fabricx wiring has no local commit pipeline
// (see platform/fabricx/core/channel/provider.go). Reaching this method means the
// vault was wired into a generic committer, which is a programming error.
func (v *Vault) RWSExists(context.Context, cdriver.TxID) bool {
      panic("fabricx vault: RWSExists called; fabricx has no local commit pipeline")
}
```

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm - this is a solid PR. Please see my comments below.

platform/fabricx/core/vault/vault_test.go
```diff
@@ -462,3 +456,40 @@ func TestRWSet_GetWriteAt(t *testing.T) {
...
+
+	const goroutines = 16
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
+	for g := range goroutines {
+		go func(g int) {
+			defer wg.Done()
+			// Each goroutine gets its own RWSet, but they all share v.marshaller.
+			rws, err := v.NewRWSet(ctx, driver.TxID("tx"))
+			require.NoError(t, err)
```
mbrandenburger: we should not call `require.*` within the goroutine. Instead we could use `assert` or maybe even better to use `errgroup`. WDYT?

platform/fabricx/core/vault/vault.go
```diff
@@ -429,89 +397,66 @@ func (v *Vault) NewRWSetFromBytes(ctx context.Context, txID cdriver.TxID, rwset
...
-		return status.code, status.message, nil
-	}
-	v.mu.RUnlock()
-
-	// Query remote service
-	statusCode, err := v.queryService.GetTransactionStatus(string(txID))
+	statuses, err := v.Statuses(ctx, txID)
 	if err != nil {
-		return fdriver.Unknown, "", errors.Wrapf(err, "failed to get transaction status for txID=%s", txID)
+		return fdriver.Unknown, "", err
```
mbrandenburger: I believe we should still keep the error wrapping to report the txID that has failed/is unknown. Note that the error returned by `v.Statuses` does not include the queried txIDs (which is ok - to prevent spamming the logs).

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm - greate updates on this PR. I think this PR is almost ready to be merged. Please see my minor comments below. Thank you

mbrandenburger (approved):

PR #1631, opened 2026-07-31
fix(comm): require TLS 1.3 minimum for websocket p2p client connections

### LINKED ISSUE #1630: websocket p2p client TLS config allows negotiating down to TLS 1.2
## Description

In `platform/view/services/comm/host/websocket/config.go`, the mTLS configuration for the websocket P2P transport is inconsistent between client and server:

- `newServerTLSConfig` pins `MinVersion: tls.VersionTLS13`
- `newClientTLSConfig` only set `MinVersion: tls.VersionTLS12`

Since both `MinVersion` and `MaxVersion` are set to `tls.VersionTLS13` on the server, the mismatch had no practical effect against this server, but the client-side minimum still permits weaker downstream negotiation (e.g. connecting to third-party or misconfigured peers, or if `MaxVersion` is ever relaxed), and it's inconsistent with the "TLS 1.3 only" guarantee implied by the server config.

## Expected behavior

Both the websocket client and server TLS configs should require TLS 1.3 as a hard minimum, matching the documented mTLS hardening for the P2P websocket transport.

## Additional context

Related test fixtures in `platform/view/services/comm/host/websocket/ws/multiplexed_provider_test.go` also allowed TLS 1.2 as a minimum for their client/server pair and should be updated for consistency.

adecaro:
## Summary
- The websocket P2P server TLS config already pinned `MinVersion: tls.VersionTLS13`, but the client config still allowed negotiating down to `tls.VersionTLS12`. This aligns the client with the server so both sides of the mTLS handshake reject anything below TLS 1.3.
- Updates the `ws` package test fixtures (`multiplexed_provider_test.go`) that also allowed TLS 1.2 as a minimum, for consistency with the enforced policy.
- Documents the TLS 1.3-only guarantee in `docs/platform/view/services/comm/websocket.md`.

Fixes #1630

## Test plan
- [x] `go build ./platform/view/services/comm/host/websocket/...`
- [x] `go vet ./platform/view/services/comm/host/websocket/...`
- [x] `go test ./platform/view/services/comm/host/websocket/...` (both `websocket` and `websocket/ws` packages pass)

### DIFF
--- a/platform/view/services/comm/host/websocket/ws/multiplexed_provider_test.go
+++ b/platform/view/services/comm/host/websocket/ws/multiplexed_provider_test.go
@@ -316,6 +316,63 @@ func TestRejectsPeerIDMismatch(t *testing.T) { //nolint:paralleltest,tparallel
 	}
 }
 
+// TestRejectsTLS12ClientHandshake asserts that the server-side TLS config's
+// MinVersion: tls.VersionTLS13 (config.go) is actually enforced: a client
+// that only offers TLS 1.2 must fail the handshake rather than being
+// negotiated down.
+func TestRejectsTLS12ClientHandshake(t *testing.T) { //nolint:paralleltest
+	testSetup(t)
+	t.Cleanup(func() {
+		goleak.VerifyNone(t, goleak.IgnoreCurrent())
+	})
+
+	p := NewMultiplexedProvider(noop.NewTracerProvider(), &disabled.Provider{}, 0)
+	t.Cleanup(func() {
+		_ = p.KillAll()
+	})
+	serverTLSConfig, clientTLSConfig, srcID := testMutualTLSConfigs(t, false)
+
+	received := make(chan struct{}, 1)
+	srv := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		err := p.NewServerStream(w, r, func(_ host.P2PStream) {
+			received <- struct{}{}
+		})
+		assert.NoError(t, err)
+	}))
+	srv.TLS = serverTLSConfig
+	srv.StartTLS()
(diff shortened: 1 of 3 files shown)


platform/view/services/comm/host/websocket/ws/multiplexed_provider_test.go
```diff
@@ -437,14 +437,14 @@ func testMutualTLSConfigs(t *testing.T, insecureSkipVerify bool) (*tls.Config, *
...
-		MinVersion:   tls.VersionTLS12,
+		MinVersion:   tls.VersionTLS13,
 		MaxVersion:   tls.VersionTLS13,
 		Certificates: []tls.Certificate{serverCert},
 		ClientAuth:   tls.RequireAndVerifyClientCert,
 		ClientCAs:    caCertPool,
 	}
 	clientTLSConfig := &tls.Config{
-		MinVersion:   tls.VersionTLS12,
+		MinVersion:   tls.VersionTLS13,
```
mbrandenburger, line 504: We probably should add a negative handshake test where the client indeed tries to connect with TLS12 and gets rejected. WDYT?

mbrandenburger (changes requested):
Thank you @adecaro for this PR.

platform/view/services/comm/host/websocket/ws/multiplexed_provider_test.go
adecaro: Good call — added `TestRejectsTLS12ClientHandshake` in 88a879c3, which forces the client's `MinVersion`/`MaxVersion` down to TLS 1.2 and asserts the handshake against the server fails with `protocol version not supported`, the server-side callback is never invoked, and no client connection is registered.

mbrandenburger (approved):

PR #1639, opened 2026-08-05
fix(test): fix flaky platform/fabric/services/endorser

mbrandenburger:
`TestParallelCollectEndorsementsOnProposalView_TimesOutOnSilentParty` was flaky because `parallelCollectEndorsementsOnProposalView.Call` armed its aggregate deadline with the same duration that `collectEndorsement` passes to `ReceiveWithTimeout`, so both fired at once and `select` picked between them at random — one path reports `"time out reached on session"`, the other `"timeout waiting for endorsement"`, and only the first matched the assertion. 

The aggregate deadline now runs a fixed grace period longer than the per-party timeout, so a silent party is consistently reported by the error that names it, leaving the aggregate deadline as the backstop for the session-setup and send steps that nothing else bounds. 

Along the way this fixes a real bug in the same loop: the `AppendProposalResponse` error path wrapped `a.err`, which is always `nil` there, so `Wrapf` returned `nil` and `Call` handed the caller `(nil, nil)` instead of the append failure. Adds tests for the aggregate deadline and for the swallowed append error, and extracts the fake wiring the parallel-endorsement tests share into a fixture.

### DIFF
--- a/platform/fabric/services/endorser/endorsement_proposal.go
+++ b/platform/fabric/services/endorser/endorsement_proposal.go
@@ -38,13 +38,27 @@ type answer struct {
 	party view.Identity
 }
 
-// defaultParallelEndorsementTimeout bounds, in aggregate, how long the initiator waits for
-// all contacted parties to answer when no explicit timeout has been configured via
+// defaultParallelEndorsementTimeout is the timeout applied when none has been configured via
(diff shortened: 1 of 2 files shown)


SaidAltury-ibm (approved):
LGTM thank you @mbrandenburger

PR #1640, opened 2026-08-05
fix finality listener leak

### LINKED ISSUE #1626: fabricx: finality listener map grows without bound; entries are only removed when the committer notifies
## Summary

`notificationListenerManager.handlers` accumulates one entry per registered transaction and has no
local expiry. The only thing that removes an entry is an inbound notification from the remote
committer. The public `AddFinalityListener` contract explicitly tells callers they do *not* need to
unregister, so when a notification never arrives, nothing — neither the framework nor the
application — will ever clean the entry up. Each entry pins the caller's listener closure, whose
retained size is unbounded.

Additionally, `AddFinalityListener` does not reject an empty `txID`, which both violates the
documented contract and creates an entry that can never be removed by any notification.

### The contract makes registration fire-and-forget

`platform/fabric/committer.go:42-45`:

```go
// AddFinalityListener registers a listener for transaction status for the passed transaction id.
// If the status is already valid or invalid, the listener is called immediately.
// When the listener is invoked, then it is also removed.
// The transaction id must not be empty.
```

"When the listener is invoked, then it is also removed" — the framework owns cleanup. Callers
follow this correctly and do not unregister; see
`integration/fabricx/simple/views/create.go:94`, which registers a listener and never removes it.

### But in the FabricX implementation, cleanup depends entirely on the remote

In `platform/fabricx/core/finality/nlm.go`, the only paths that shrink `handlers` are:

| Line | Path | Trigger |
|------|------|---------|
| `:134` | dispatcher | a notification arrived for that txID |
| `:259` | `AddFinalityListener` | rollback when the request send fails |
| `:296` | `RemoveFinalityListener` | explicit removal by the caller |
| `:171` | `listen()` teardown | `clear(n.handlers)` when the stream dies |

Since the contract says callers won't use `RemoveFinalityListener`, and `:259` only covers a failed
send, the steady-state removal path is `:134` — a notification from the committer.

The 10 s timeout at `nlm.go:239` is a field *inside the outbound request*, asking the committer to
give up and reply with `TimeoutTxIds`. There is no local timer:

```go
req := &committerpb.NotificationRequest{
	TxStatusRequest: &committerpb.TxIDsBatch{
		TxIds: txIDs,
	},
	// TODO: set a proper timeout
	Timeout: durationpb.New(10 * time.Second),
}
```

If that reply never arrives, the entry is permanent.

### Concrete ways the reply never arrives

1. The committer accepts the subscription but never emits for that txID — server-side bug, overload,
   or a dropped subscription.
2. `notifyStream.Send(req)` succeeds (`nlm.go:99`) but the committer discards the request. The map
   entry is written at `nlm.go:225`, *before* the send at `nlm.go:256`, so it is already there.
3. **Empty txID.** `AddFinalityListener` guards only `listener == nil`:

   ```go
   // nlm.go:211-214
   func (n *notificationListenerManager) AddFinalityListener(txID driver.TxID, listener fabric.FinalityListener) error {
       if listener == nil {
           return errors.New("listener nil")
       }
   ```

   An empty `txID` therefore creates `handlers[""]` and sends a subscription for `TxIds: [""]`. No
   committer will ever return a status for `""`, so the entry can never be removed.

   The generic implementation *does* guard this —
   `platform/common/core/generic/committer/listenermgr.go:54-57`:

   ```go
   func (c *finalityListenerManager[V]) AddListener(txID driver.TxID, toAdd driver.FinalityListener[V]) error {
       if len(txID) == 0 {
           return errors.Errorf("tx id must be not empty")
       }
   ```

   and `integration/fabric/iou/views/approver.go:95` asserts that this returns an error. The FabricX
   implementation accepts it silently, so the two drivers behave differently for the same API call.

### Impact

The map entry itself is small — a slice header plus the 64-character txID key, on the order of
150 bytes. The significant cost is the retained **listener closure**. In the `IsFinal` path the
closure captures a buffered channel (`platform/fabricx/core/channel/wrappers.go:68-71`); an
application-supplied listener can capture arbitrarily more state. So the bytes pinned per leaked
entry are not visible from the size of the map.

Partial mitigation that already exists: `clear(n.handlers)` at `nlm.go:170-172` when `listen()`
exits. A node whose notification stream periodically drops and reconnects self-heals; a node with a
stable stream accumulates indefinitely.

Note also that the FabricX implementation does not implement the "called immediately if already
final" half of the contract locally — it delegates that entirely to the remote committer.

## Proposed fix

**1. Add a local expiry to `handlers`, so cleanup never depends on remote liveness.**

Store a deadline alongside each handler list:

```go
type handlerEntry struct {
	listeners []fabric.FinalityListener
	expiresAt time.Time
}

handlers map[driver.TxID]*handlerEntry
```

Set `expiresAt` in `AddFinalityListener` from a new `listenerTTL` field on
`notificationListenerManager` (defaulted next to `DefaultHandlerTimeout` at `nlm.go:33`, and
comfortably longer than the 10 s request timeout at `nlm.go:239` — e.g. 30 s — so the remote is
always given its chance first).

Sweep expired entries from the dispatcher goroutine, which is already awake on every notification
(`nlm.go:106-164`), plus a `time.Ticker` in the same `select` so that a completely idle stream still
sweeps. On expiry, invoke the listeners with `fdriver.Unknown` before dropping them — that is exactly
what the remote's `TimeoutTxIds` path would have delivered (`nlm.go:181-184`), so callers such as
`finalityServiceAdapter.IsFinal` (`wrappers.go:66-122`) see no behavioural change and stop blocking.

**2. Reject an empty txID**, matching the documented contract and the generic implementation:

```go
if len(txID) == 0 {
	return errors.New("tx id must be not empty")
}
```

**3. Optional, follow-up:** implement the "if already final, invoke immediately" half of the
contract by consulting `QueryService.GetTransactionStatuses` on registration, so a listener
registered after the transaction already committed does not wait for a notification that will
never come.

## Testing

`platform/fabricx/core/finality` already has a test suite with mocks for the notifier client
(`core/finality/mock/`), so this is unit-testable without a network:

- register a listener, never deliver a notification, advance past the TTL, assert the entry is gone
  and the listener was invoked with `Unknown`;
- assert `AddFinalityListener("", l)` returns an error and creates no entry;
- assert an entry removed by a real notification is *not* double-invoked by the sweeper.

## Code references

- `platform/fabricx/core/finality/nlm.go:41-42` — the `handlers` map and its mutex
- `platform/fabricx/core/finality/nlm.go:106-164` — dispatcher; removal on notification at `:134`
- `platform/fabricx/core/finality/nlm.go:170-172` — `clear(n.handlers)` on stream teardown
- `platform/fabricx/core/finality/nlm.go:211-262` — `AddFinalityListener`; missing txID guard at `:212`, insert at `:225`, request timeout at `:239`, send at `:256`
- `platform/fabricx/core/finality/nlm.go:178-208` — `parseResponse`, including the `TimeoutTxIds` handling to mirror on local expiry
- `platform/fabricx/core/finality/provider.go:149-153` — manager construction, where a `listenerTTL` default would be wired
- `platform/fabric/committer.go:42-48` — the documented public contract
- `platform/common/core/generic/committer/listenermgr.go:54-57` — the generic empty-txID guard to match
- `platform/fabricx/core/channel/wrappers.go:66-122` — `IsFinal`, the in-tree consumer
- `integration/fabricx/simple/views/create.go:94` — fire-and-forget registration in an integration test

SaidAltury-ibm:
Fixes #1626.

`notificationListenerManager.handlers` only shrank when a notification arrived
from the committer. The contract says the framework removes a listener once it
fires, so callers correctly never unregister — if the notification never comes,
nothing cleans up and the entry keeps the caller's listener closure alive. The
10s `Timeout` on the outbound request isn't a safety net: it asks the
*committer* to give up and reply, over the same stream whose silence is the
problem.

- **Reject an empty txID.** It can never appear in a notification, so its entry
  was unremovable. The generic driver already rejects this; the message now
  matches.
- **Expire stale listeners locally.** Entries carry a deadline and are settled
  with `Unknown` once it passes — the same outcome the committer's own
  `TimeoutTxIds` path produces. Defaults: TTL 2m, sweep 30s.
- **Assert the empty-txID guard against a live committer** in the simple
  integration test.

### DIFF
--- a/integration/fabricx/simple/views/create.go
+++ b/integration/fabricx/simple/views/create.go
@@ -96,6 +96,17 @@ func (i *CreateView) Call(viewCtx view.Context) (any, error) {
 		return nil, err
 	}
 
+	// An empty txID must be rejected: no committer notification could ever match
+	// it, so the registration would never be removed. The generic driver enforces
+	// this too (see integration/fabric/iou/views/approver.go), so this asserts the
+	// two drivers agree against a live committer. A throwaway WaitGroup is used so
+	// that an unexpected success cannot corrupt the real listener's count above.
+	var unusedWG sync.WaitGroup
+	unusedWG.Add(1)
+	if err := lm.AddFinalityListener("", utils.NewFinalityListener(tx.ID(), fdriver.Valid, &unusedWG)); err == nil {
+		return nil, errors.New("AddFinalityListener with an empty txID must fail")
+	}
+
 	// now we have a committer listener registered, we send the approved transaction to the orderer
 	logger.Infof("Submit tx (txID=%v) to ordering service", tx.ID())
 	if _, err = viewCtx.RunView(state.NewOrderingAndFinalityWithTimeoutView(tx, FinalityTimeout)); err != nil {
--- a/integration/fabricx/simple/views/create.go
+++ b/integration/fabricx/simple/views/create.go
@@ -96,17 +96,6 @@ func (i *CreateView) Call(viewCtx view.Context) (any, error) {
 		return nil, err
 	}
 
-	// An empty txID must be rejected: no committer notification could ever match
-	// it, so the registration would never be removed. The generic driver enforces
-	// this too (see integration/fabric/iou/views/approver.go), so this asserts the
-	// two drivers agree against a live committer. A throwaway WaitGroup is used so
-	// that an unexpected success cannot corrupt the real listener's count above.
-	var unusedWG sync.WaitGroup
-	unusedWG.Add(1)
-	if err := lm.AddFinalityListener("", utils.NewFinalityListener(tx.ID(), fdriver.Valid, &unusedWG)); err == nil {
-		return nil, errors.New("AddFinalityListener with an empty txID must fail")
-	}
-
 	// now we have a committer listener registered, we send the approved transaction to the orderer
 	logger.Infof("Submit tx (txID=%v) to ordering service", tx.ID())
 	if _, err = viewCtx.RunView(state.NewOrderingAndFinalityWithTimeoutView(tx, FinalityTimeout)); err != nil {
--- a/platform/fabricx/core/finality/nlm.go
+++ b/platform/fabricx/core/finality/nlm.go
@@ -212,6 +212,13 @@ func (n *notificationListenerManager) AddFinalityListener(txID driver.TxID, list
 	if listener == nil {
 		return errors.New("listener nil")
 	}
+	// An empty txID can never appear in a committer notification, so the entry it
+	// would create is unremovable. The generic driver already rejects this (see
+	// platform/common/core/generic/committer/listenermgr.go); keep the message
+	// identical so both drivers behave the same for the same API call.
+	if len(txID) == 0 {
+		return errors.New("tx id must be not empty")
+	}
 
 	n.handlersMu.Lock()
 	defer n.handlersMu.Unlock()
--- a/platform/fabricx/core/finality/nlm.go
+++ b/platform/fabricx/core/finality/nlm.go
@@ -157,13 +157,50 @@ func (n *notificationListenerManager) listen(ctx context.Context) error {
 	err = g.Wait()
 	logger.Debugf("Notification listener stream stopped.")
 
-	// Cleanup handlers map when listen() exits
-	n.handlersMu.Lock()
-	clear(n.handlers)
-	n.handlersMu.Unlock()
-	logger.Debugf("Cleared handlers map on listen() exit")
+	// The stream is gone, so nothing will ever notify these listeners. Settle them
+	// with Unknown instead of dropping them silently, so anyone blocked in IsFinal
+	// is released now rather than waiting out their own context.
+	//
+	// ctx, not gCtx: the errgroup context is already cancelled by the time g.Wait()
+	// returns, and invokeHandler derives its handler timeout from what we pass, so
+	// gCtx would hand every listener a dead context and deliver nothing. Strip
+	// cancellation from the parent too -- listen() is often returning *because*
+	// ctx was cancelled, and these callbacks still need to run.
+	n.settleAllAndClear(context.WithoutCancel(ctx), fdriver.Unknown)
 
 	return err
+}
+
+// settleAllAndClear empties the handlers map, invoking every listener still in it
+// with the given status. Used on stream teardown, where no notification can
+// arrive any more.
+func (n *notificationListenerManager) settleAllAndClear(ctx context.Context, status int) {
+	type pending struct {
+		txID      string
+		listeners []fabric.FinalityListener
+	}
+
+	var batch []pending
+
+	n.handlersMu.Lock()
+	for txID, entry := range n.handlers {
+		batch = append(batch, pending{txID: txID, listeners: entry.listeners})
+	}
+	clear(n.handlers)
+	n.handlersMu.Unlock()
+
+	if len(batch) == 0 {
+		logger.Debugf("Cleared handlers map on listen() exit")
+		return
+	}
+
+	logger.Debugf("Settling %d pending finality listener(s) with status %d on stream teardown", len(batch), status)
+
+	for _, p := range batch {
+		for _, h := range p.listeners {
+			n.invokeHandler(ctx, h, p.txID, status)
+		}
+	}
 }
 
 func parseResponse(resp *committerpb.NotificationResponse) map[string]int {
--- a/platform/fabricx/core/finality/nlm.go
+++ b/platform/fabricx/core/finality/nlm.go
@@ -32,13 +32,41 @@ var logger = logging.MustGetLogger()
 // will leak goroutines, but this is preferable to blocking the dispatcher.
 const DefaultHandlerTimeout = 5 * time.Second
 
+// DefaultListenerTTL bounds how long a listener may wait for a notification that
+// may never arrive. It is deliberately much longer than the timeout we ask the
+// committer for in AddFinalityListener: that timeout is documented non-strict
+// ("it is possible to receive notifications after the timeout has passed", see
+// notify.proto), so the remote must be given ample room to answer before we give
+// up locally. Expiry is a backstop against silence, not a competitor to the
+// remote deadline.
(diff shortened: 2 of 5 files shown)


integration/fabricx/simple/views/create.go
```diff
@@ -96,6 +96,17 @@ func (i *CreateView) Call(viewCtx view.Context) (any, error) {
...
+	// it, so the registration would never be removed. The generic driver enforces
+	// this too (see integration/fabric/iou/views/approver.go), so this asserts the
+	// two drivers agree against a live committer. A throwaway WaitGroup is used so
+	// that an unexpected success cannot corrupt the real listener's count above.
+	var unusedWG sync.WaitGroup
+	unusedWG.Add(1)
+	if err := lm.AddFinalityListener("", utils.NewFinalityListener(tx.ID(), fdriver.Valid, &unusedWG)); err == nil {
+		return nil, errors.New("AddFinalityListener with an empty txID must fail")
+	}
+
```
mbrandenburger: This does not right to me. Here we are testing the fabric-x `lm.AddFinalityListener` that it actually returns an error if the txID is empty. This is covered by unit tests already. If this would be a view that is used in a fabric and fabric-x integration test, I would see some value to test it here, but this only tests the fabric-x impl.

I think we can remove this.

platform/fabricx/core/finality/nlm.go
```diff
@@ -295,9 +331,103 @@ func (n *notificationListenerManager) RemoveFinalityListener(txID string, listen
...
+		return
+	}
+
+	logger.Debugf("Settling %d expired finality listener(s) with Unknown", len(batch))
+
+	for _, e := range batch {
+		for _, h := range e.listeners {
+			n.invokeHandler(ctx, h, e.txID, fdriver.Unknown)
+		}
+	}
```
mbrandenburger, line 469: This is great! Here we make sure that we actually fire the listeners - which is - super helpful if they are implemented in the way we use in `integration/fabricx/simple/views/create.go`.  

I was wondering if we should do the same above when we cleanup the handler in case of a "disconnect"

```go
	// Cleanup handlers map when listen() exits
	n.handlersMu.Lock()
	// --> fire handler with status unknown here <--
	clear(n.handlers)
	n.handlersMu.Unlock()
	logger.Debugf("Cleared handlers map on listen() exit")
```

WDYT?

platform/fabricx/core/finality/provider.go
```diff
@@ -148,8 +148,10 @@ func newNotifiWithGRPC(network string, grpcClientProvider GRPCClientProvider) (*
 		notifyClient:   notifyClient,
 		requestQueue:   make(chan *committerpb.NotificationRequest),  // Queue for outgoing requests to the committer
 		responseQueue:  make(chan *committerpb.NotificationResponse), // Queue for incoming responses/notifications
-		handlers:       make(map[string][]fabric.FinalityListener),   // Map: txID -> list of listeners
+		handlers:       make(map[driver.TxID]*handlerEntry),          // Map: txID -> listeners + local expiry deadline
 		handlerTimeout: DefaultHandlerTimeout,
+		listenerTTL:    DefaultListenerTTL,
+		sweepInterval:  DefaultSweepInterval,
```
mbrandenburger, line 159: I suggest that we make these settings tunable via our config system. This can be another PR.

platform/fabricx/core/finality/nlm.go
```diff
@@ -212,21 +236,33 @@ func (n *notificationListenerManager) AddFinalityListener(txID driver.TxID, list
...
-		logger.Warnf("The exact same listener is already registered for txID=%v. Skipping.", txID)
-		// Do not register the same instance twice
+	if entry, existed := n.handlers[txID]; existed {
+		if slices.Contains(entry.listeners, listener) {
+			logger.Warnf("The exact same listener is already registered for txID=%v. Skipping.", txID)
+			// Do not register the same instance twice
+			return nil
+		}
+		// A joining listener inherits the existing deadline rather than extending
+		// it, so a busy txID cannot keep its entry alive indefinitely.
```
mbrandenburger, line 261: do we actually cover this behavior via unit tests?

platform/fabricx/core/finality/nlm.go
```diff
@@ -295,9 +331,103 @@ func (n *notificationListenerManager) RemoveFinalityListener(txID string, listen
 		logger.Debugf("Last finality listener removed for txID=%s.", txID)
 		delete(n.handlers, txID)
 	} else {
-		n.handlers[txID] = newHandlers
+		// Mutating the entry in place keeps the existing deadline: removing one
+		// listener must not extend the lifetime of the ones still waiting.
```
mbrandenburger, line 339: do we actually cover this behavior via unit tests?

platform/fabricx/core/finality/nlm.go
```diff
@@ -110,12 +138,27 @@ func (n *notificationListenerManager) listen(ctx context.Context) error {
...
 		for {
 			select {
 			case <-gCtx.Done():
 				return gCtx.Err()
 			case resp = <-n.responseQueue:
+			case <-ticker.C:
+				n.sweepExpired(gCtx)
+				// A tick carries no response; resp still holds the previous one, so
+				// falling through here would dispatch it a second time.
+				continue
```
mbrandenburger: while this is correct -  I suggest that we make this part a bit cleaner.

It would be easier to read if can move the response handling into a separate function and just keep a simple for loop here. For example:

```go
for {
      select {
      case <-gCtx.Done():
              return gCtx.Err()
      case <-ticker.C:
              n.sweepExpired(gCtx)
      case resp := <-n.responseQueue:
              n.dispatch(gCtx, resp)
      }
}
```

WDYT?

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm. This is a solid PR already, please see my comments below.

platform/fabricx/core/finality/provider.go
SaidAltury-ibm: agreed, and filed as #1641 with a `TODO` in `newNotifiWithGRPC` pointing at it

mbrandenburger (approved):
Thank you @SaidAltury-ibm

PR #1644, opened 2026-08-07
fix!: Bind background goroutines to a context

mbrandenburger:
Binds background goroutines to a context so they stop when their owner does.

Constructors that used to spawn an unstoppable goroutine now take a `context.Context`:

- `cache.NewTimeoutCache` / `NewTimeoutEviction` — the eviction loop used `time.Tick`, which can never be stopped
- `runner.NewBatchRunner` / `NewBatchExecutor` — `start()` looped forever
- `events.NewListenerManager` — scanned blocks on `context.Background()`
- `chaincode.NewChaincode` / `NewManager` — the discovery cache is now created once per chaincode instead of once per `NewDiscovery` call

To give those contexts an owner, `Channel` holds a cancel func created in the provider and cancelled in `Close()`. The p2p service drains in-flight `handleMessage` goroutines before its read loop returns, and the SDK's server goroutines are tracked in a `WaitGroup` that shutdown joins before stopping the KVS.

Cancellation is reported instead of swallowed: the batcher returns `ctx.Err()` rather than a zero value with a nil error, and `Notifier.Subscribe` fails once the notifier is closed.

Also fixed along the way:

- `UnsubscribeAll` used `clear()` on the subscriber slice, leaving nil callbacks for `dispatch` to invoke — a panic
- `NewDiscovery`'s nil check used `||` where it meant `&&`, so it dereferenced whatever it was checking
- `NewTimeoutEviction` panicked on a non-positive timeout (`time.NewTicker` does); it now disables eviction
- `SendAndRecv` could block forever if `Recv()` stalled and `Close()` failed to unblock it; there is now a hard timeout, armed only when the caller's context has no deadline of its own

## Breaking changes

A leading `context.Context` was added to `NewBatchExecutor`, `NewBatchRunner`, `NewTimeoutCache`, `NewTimeoutEviction`, `NewListenerManager`, `NewSequentialListenerManager`, `NewDeliveryFLM`, `NewChaincode`, `NewManager` and `RunView`. `view.WithGoContext` is new — unlike `view.WithContext` it sets only the go context, without also forcing `SameContext`.

## Testing

`make checks` and `make unit-tests` pass. Leak coverage uses `goleak`; adopting it surfaced pre-existing leaks in `batch_test.go`, `timeout_test.go` and `chaincode_test.go`, whose fixtures were rooted at `context.Background()` and are now bound to `t.Context()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ integration/**/out/
 
 integration/nwo/cmd/fsccli/cmd
 /.bob/notes/pending-notes.txt
+
+# compiled cc/query chaincode binary (go build artifact)
+platform/fabric/services/state/cc/query/query
--- a/integration/fsc/pingpong/fake/runner.go
+++ b/integration/fsc/pingpong/fake/runner.go
@@ -105,7 +105,7 @@ func (c *DelegatedContext) GetSession(caller view.View, party view.Identity, bou
 			}
 			left := biChannel.LeftSession()
 			right := biChannel.RightSession()
-			view2.RunView(c, responder.ResponderView, view.AsResponder(right))
+			view2.RunView(c.Context(), c, responder.ResponderView, view.AsResponder(right))
 
 			responder.Channel = biChannel
 			return left, nil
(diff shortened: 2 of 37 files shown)


SaidAltury-ibm (approved):

PR #1646, opened 2026-08-08
fix(comm): decouple websocket client stream lifetime from per-call context

### LINKED ISSUE #1645: websocket P2P transport: client stream lifetime incorrectly tied to per-call context, breaking cached stream reuse across rounds
## Summary

`make integration-tests-fsc-pingpong` fails when the FSC nodes are configured to use the **websocket** P2P transport (`fsc.WebSocket`). The same test passes fine with the **libp2p** transport, which pointed at a websocket-transport-specific bug rather than an issue in the pingpong views themselves.

## Root cause

Both `websocket.SimpleProvider.NewClientStream` and `websocket.MultiplexedProvider`'s client-side sub-connection setup (`newClientSubConn`) build the returned `P2PStream`'s internal lifetime directly from the `context.Context` passed in by the caller:

```go
return NewWSStream(conn, ctx, info), nil          // simple_provider.go
return NewWSStream(sc, ctx, info), nil            // multiplexed_provider.go
```

`NewWSStream` derives a cancellable context from that `ctx` and spawns a goroutine that closes the stream as soon as it's cancelled (`<-ctx.Done() -> Close()`).

The problem: client streams/sub-connections are **cached and reused** across multiple protocol rounds (e.g. the pingpong initiator sends 3 ping/pong rounds over the same session/stream). Each round is wrapped in its own short-lived `context.WithTimeout(...)`, cancelled via `defer cancel()` as soon as that round's view returns — this is a completely ordinary, idiomatic Go pattern for bounding a single call. But because the stream's own lifetime was wired to that same per-call context, cancelling it at the end of round 1 immediately tears down the underlying stream/sub-connection out from under round 2, which then fails.

The server side (`NewServerStream` / `newServerSubConn`) already did this correctly — it builds the stream's context from `context.Background()` (only propagating the tracing span from the caller's context), decoupling the stream's lifetime from any particular request. The client side did not follow the same pattern, which is the actual bug.

## Reproduction

1. Run `make integration-tests-fsc-pingpong` (websocket transport, 3 ping/pong rounds over one session).
2. Round 1 succeeds. As soon as its per-round context is cancelled (`defer cancel()`), the underlying websocket client stream/sub-connection is closed.
3. Round 2 fails because the stream it expected to reuse is gone.

Confirmed via the libp2p variant of the same test (`Network-based Ping pong With LibP2P`), which doesn't share this transport-specific context-wiring, passing without issue.

## Expected behavior

A per-round/per-call `context.Context` passed to `NewClientStream` should only be used for that call's own bookkeeping (e.g. tracing span propagation), not for the lifetime of the cached stream/connection it returns — mirroring how the server side already handles this.

adecaro:
## Problem

Fixes #1645.

`make integration-tests-fsc-pingpong` fails on the websocket P2P transport: round 1 of a 3-round ping/pong exchange succeeds, but round 2 fails because the underlying client stream/sub-connection has been torn down.

## Root cause

`SimpleProvider.NewClientStream` and `MultiplexedProvider`'s client-side `newClientSubConn` built the returned `P2PStream`'s internal context directly from the caller-supplied `context.Context`. Client streams/sub-connections are cached and reused across multiple protocol rounds, but each round is wrapped in its own short-lived `context.WithTimeout(...)` + `defer cancel()`. Cancelling that per-call context at the end of round 1 immediately closed the stream, breaking round 2.

The server side (`NewServerStream` / `newServerSubConn`) already avoided this by deriving the stream's context from `context.Background()` and only propagating the tracing span from the caller's context.

## Fix

- `platform/view/services/comm/host/websocket/ws/simple_provider.go`: build the client stream's context from `context.Background()` + `trace.ContextWithRemoteSpanContext(...)`, same as the server side.
- `platform/view/services/comm/host/websocket/ws/multiplexed_provider.go`: same fix in `newClientSubConn`.
- `integration/fsc/pingpong/responder.go`: restructured to mirror `initiator.go`'s `ping()`/`PingView` pattern — a `pong()`/`PongView` pair that wraps each round in its own cancellable context, so the responder exercises the same code path as the initiator (both can hit `NewClientStream`/`newClientSubConn` on a cache miss when sending).
- `platform/view/services/comm/host/websocket/ws/context_detachment_test.go`: new regression tests covering both `SimpleProvider` and `MultiplexedProvider` client streams — verifying the stream survives cancellation of the caller's context, remains usable for a subsequent round, and that closing it through a legitimate path (explicit `Close()` or `KillAll()`) fully releases all resources (channels closed, no leaked goroutines, physical connection closed only when appropriate).
- `platform/view/services/comm/host/websocket/ws/stream.go`, `platform/view/services/view/manager.go`: additional debug logging that was useful in diagnosing this issue, kept for future troubleshooting.

## Test plan

- [x] `go build ./...` / `go vet ./...`
- [x] `go test ./platform/view/services/comm/host/websocket/... -race`
- [x] `make integration-tests-fsc-pingpong`

### DIFF
--- a/integration/fsc/pingpong/initiator.go
+++ b/integration/fsc/pingpong/initiator.go
@@ -12,13 +12,32 @@ import (
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/id"
 	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/view"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 )
 
 var logger = logging.MustGetLogger()
+
+const (
+	pingMessage     = "ping"
+	pongMessage     = "pong"
+	finishedMessage = "finished"
+
+	// DefaultRounds is the number of ping/pong rounds run when Params.Rounds is not set.
+	DefaultRounds = 3
+
+	// roundTimeout bounds a single round end to end: session lookup (which may have to open a
+	// new stream), send, and wait for the peer's answer. Initiator and responder use the same
+	// value, so that neither gives up on the other prematurely.
+	roundTimeout = 1 * time.Minute
+)
+
+// Params is the input of the Initiator view.
+type Params struct {
+	// Rounds is the number of ping/pong rounds to run; values <= 0 select DefaultRounds.
+	Rounds int `json:"rounds,omitempty"`
+}
 
 // PingView sends a single ping and waits for the pong over the context's session.
 type PingView struct {
@@ -27,60 +46,104 @@ type PingView struct {
 
 func (p *PingView) Call(viewCtx view.Context) (any, error) {
 	session, err := viewCtx.GetSession(viewCtx.Initiator(), p.responder)
-	assert.NoError(err)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting session to [%s]", p.responder)
+	}
 
-	logger.Infof("send ping")
-
-	if err := session.SendWithContext(viewCtx.Context(), []byte("ping")); err != nil {
-		return nil, errors.Wrap(err, "failed to send ping")
+	logger.DebugfContext(viewCtx.Context(), "send %s", pingMessage)
+	if err := session.SendWithContext(viewCtx.Context(), []byte(pingMessage)); err != nil {
+		return nil, errors.Wrapf(err, "failed to send %s", pingMessage)
 	}
 
 	ch := session.Receive()
 	select {
-	case msg := <-ch:
+	case msg, ok := <-ch:
+		if !ok {
+			return nil, errors.Errorf("session [%s] closed while waiting for %s", session.Info().ID, pongMessage)
+		}
 		if msg.Status == view.ERROR {
 			return nil, errors.New(string(msg.Payload))
 		}
-		if m := string(msg.Payload); m != "pong" {
-			return nil, errors.Errorf("expected pong, got %s", m)
+		if m := string(msg.Payload); m != pongMessage {
+			return nil, errors.Errorf("expected %s, got %s", pongMessage, m)
 		}
-		logger.Infof("pong received")
+		logger.DebugfContext(viewCtx.Context(), "%s received", pongMessage)
 	case <-viewCtx.Context().Done():
-		return nil, errors.Errorf("expected ping, got %s", viewCtx.Context().Err())
-	case <-time.After(1 * time.Minute):
-		return nil, errors.New("responder didn't pong in time")
+		return nil, errors.Wrapf(viewCtx.Context().Err(), "no %s received in time", pongMessage)
 	}
 
 	return nil, nil
 }
 
-func ping(viewCtx view.Context, responder view.Identity) (any, error) {
-	ctx, cancel := context.WithTimeout(viewCtx.Context(), 10*time.Minute)
-	runCtx := view2.WrapContext(viewCtx, ctx)
-	defer func() {
-		logger.Infof("call cancel on view context [%s:%s]", runCtx.ID(), viewCtx.ID())
-		cancel()
-	}()
-
-	return runCtx.RunView(&PingView{
-		responder: responder,
-	})
+// FinishedView tells the responder that the initiator is done. Closing a session is a local
+// operation only and puts nothing on the wire, so without this message the responder would
+// keep waiting for the next ping.
+type FinishedView struct {
+	responder view.Identity
 }
 
-type Initiator struct{}
-
-func (p *Initiator) Call(viewCtx view.Context) (any, error) {
-	// Retrieve responder identity
-	identityProvider, err := id.GetProvider(viewCtx)
-	assert.NoError(err, "failed getting identity provider")
-	responder := identityProvider.Identity("responder")
-
-	// Open a session to the responder
-	for range 3 {
-		_, err = ping(viewCtx, responder)
-		assert.NoError(err, "ping round failed")
+func (f *FinishedView) Call(viewCtx view.Context) (any, error) {
+	session, err := viewCtx.GetSession(viewCtx.Initiator(), f.responder)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting session to [%s]", f.responder)
 	}
 
-	// Return
+	logger.DebugfContext(viewCtx.Context(), "send %s", finishedMessage)
+	if err := session.SendWithContext(viewCtx.Context(), []byte(finishedMessage)); err != nil {
+		return nil, errors.Wrapf(err, "failed to send %s", finishedMessage)
+	}
+	return nil, nil
+}
+
+// runRound runs v in a context of its own, bounded by roundTimeout and cancelled as soon as the
+// round is over. Every round gets a fresh, short-lived context so that a stale round's
+// cancellation can never affect a later one that happens to reuse the same cached stream; see
+// the regression tests in platform/view/services/comm/host/websocket/ws/context_detachment_test.go.
+func runRound(viewCtx view.Context, v view.View) (any, error) {
+	ctx, cancel := context.WithTimeout(viewCtx.Context(), roundTimeout)
+	defer cancel()
+
+	return view2.WrapContext(viewCtx, ctx).RunView(v)
+}
+
+func ping(viewCtx view.Context, responder view.Identity) (any, error) {
+	return runRound(viewCtx, &PingView{responder: responder})
+}
+
+func finish(viewCtx view.Context, responder view.Identity) (any, error) {
+	return runRound(viewCtx, &FinishedView{responder: responder})
+}
+
+// Initiator runs Rounds (or DefaultRounds, if Rounds is unset) ping/pong rounds against the
+// responder identity, then tells the responder the protocol is over.
+type Initiator struct {
+	Params
+}
+
+func (p *Initiator) Call(viewCtx view.Context) (any, error) {
+	identityProvider, err := id.GetProvider(viewCtx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed getting identity provider")
+	}
+	responder := identityProvider.Identity("responder")
+
+	rounds := p.rounds()
+	logger.DebugfContext(viewCtx.Context(), "ping pong with [%s] over [%d] rounds", responder, rounds)
+	for round := range rounds {
+		if _, err := ping(viewCtx, responder); err != nil {
+			return nil, errors.Wrapf(err, "ping round [%d/%d] failed", round+1, rounds)
+		}
+	}
+	if _, err := finish(viewCtx, responder); err != nil {
+		return nil, errors.Wrapf(err, "failed signalling the end of the protocol after [%d] rounds", rounds)
+	}
+
 	return "OK", nil
+}
+
+func (p *Initiator) rounds() int {
+	if p.Rounds <= 0 {
+		return DefaultRounds
+	}
+	return p.Rounds
 }
--- a/integration/fsc/pingpong/initiator.go
+++ b/integration/fsc/pingpong/initiator.go
@@ -23,13 +23,11 @@ var logger = logging.MustGetLogger()
 // PingView sends a single ping and waits for the pong over the context's session.
 type PingView struct {
 	responder view.Identity
-	session   view.Session
 }
 
 func (p *PingView) Call(viewCtx view.Context) (any, error) {
-	// session, err := viewCtx.GetSession(viewCtx.Initiator(), p.responder)
-	// assert.NoError(err)
-	session := p.session
+	session, err := viewCtx.GetSession(viewCtx.Initiator(), p.responder)
+	assert.NoError(err)
 
 	logger.Infof("send ping")
 
@@ -57,9 +55,6 @@ func (p *PingView) Call(viewCtx view.Context) (any, error) {
 }
 
 func ping(viewCtx view.Context, responder view.Identity) (any, error) {
-	session, err := viewCtx.GetSession(viewCtx.Initiator(), responder)
-	assert.NoError(err)
-
 	ctx, cancel := context.WithTimeout(viewCtx.Context(), 10*time.Minute)
 	runCtx := view2.WrapContext(viewCtx, ctx)
 	defer func() {
@@ -69,7 +64,6 @@ func ping(viewCtx view.Context, responder view.Identity) (any, error) {
 
 	return runCtx.RunView(&PingView{
 		responder: responder,
-		session:   session,
 	})
 }
 
--- a/integration/fsc/pingpong/initiator.go
+++ b/integration/fsc/pingpong/initiator.go
@@ -7,13 +7,71 @@ SPDX-License-Identifier: Apache-2.0
 package pingpong
 
 import (
+	"context"
 	"time"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/id"
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/view"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 )
+
+var logger = logging.MustGetLogger()
+
+// PingView sends a single ping and waits for the pong over the context's session.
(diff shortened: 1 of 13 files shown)


integration/fsc/pingpong/initiator.go
```diff
@@ -7,42 +7,80 @@ SPDX-License-Identifier: Apache-2.0
...
+type PingView struct {
+	responder view.Identity
+}
 
-	// Open a session to the responder
-	session, err := viewCtx.GetSession(viewCtx.Initiator(), responder)
-	assert.NoError(err) // Send a ping
+func (p *PingView) Call(viewCtx view.Context) (any, error) {
+	session, err := viewCtx.GetSession(viewCtx.Initiator(), p.responder)
+	assert.NoError(err)
```
mbrandenburger: I suggest to return the error here. I believe we should transition away from using the assertions within the test views.

integration/fsc/pingpong/initiator.go
```diff
@@ -7,42 +7,80 @@ SPDX-License-Identifier: Apache-2.0
...
+		return nil, errors.Errorf("expected ping, got %s", viewCtx.Context().Err())
 	case <-time.After(1 * time.Minute):
 		return nil, errors.New("responder didn't pong in time")
 	}
 
+	return nil, nil
+}
+
+func ping(viewCtx view.Context, responder view.Identity) (any, error) {
+	ctx, cancel := context.WithTimeout(viewCtx.Context(), 10*time.Minute)
```
mbrandenburger: above the the ping view uses another timeout for one minute - having 10min here seems very long? Is this a reasonable timeout?

integration/fsc/pingpong/initiator.go
```diff
@@ -7,42 +7,80 @@ SPDX-License-Identifier: Apache-2.0
...
+	// Retrieve responder identity
+	identityProvider, err := id.GetProvider(viewCtx)
+	assert.NoError(err, "failed getting identity provider")
+	responder := identityProvider.Identity("responder")
+
+	// Open a session to the responder
+	for range 3 {
+		_, err = ping(viewCtx, responder)
+		assert.NoError(err, "ping round failed")
+	}
```
mbrandenburger: as commented above ... I suggest to return the error instead of using the noError assertion, this is our chance to clean this test up. WDYT?

platform/view/services/comm/host/websocket/ws/stream.go
```diff
@@ -257,7 +258,7 @@ func (s *stream) Write(p []byte) (int, error) {
 func (s *stream) Close() error {
 	var err error
 	s.closeOnce.Do(func() {
-		logger.Debugf("Close connection for context [%s]", s.ContextID())
+		logger.Debugf("close connection for context [%s][%s]", s.ContextID(), string(debug.Stack()))
```
mbrandenburger: the Stack trace might be useful while working on this PR; however - it creates a lot of noise and I believe we might have a perf issue. - I would remove it here. If we really need to keep it better put it behind
` if logger.IsEnabledFor(zapcore.DebugLevel)` ...

platform/view/services/view/manager.go
```diff
@@ -319,8 +319,10 @@ func (cm *Manager) DeleteContext(contextID string) {
 	cm.contextsMu.Lock()
 	defer cm.contextsMu.Unlock()
 
+	logger.Debugf("deleting context [%s]\n", contextID)
```
mbrandenburger:
```suggestion
	logger.Debugf("deleting context [%s]", contextID)
```

platform/view/services/view/manager.go
```diff
@@ -319,8 +319,10 @@ func (cm *Manager) DeleteContext(contextID string) {
...
 	// dispose context
 	if viewCtx, ok := cm.contexts[contextID]; ok {
+		logger.Debugf("found, dispose context [%s]\n", contextID)
```
mbrandenburger:
```suggestion
		logger.Debugf("found, dispose context [%s]", contextID)
```

integration/fsc/pingpong/initiator.go
```diff
@@ -7,42 +7,80 @@ SPDX-License-Identifier: Apache-2.0
...
+type Initiator struct{}
+
+func (p *Initiator) Call(viewCtx view.Context) (any, error) {
+	// Retrieve responder identity
+	identityProvider, err := id.GetProvider(viewCtx)
+	assert.NoError(err, "failed getting identity provider")
+	responder := identityProvider.Identity("responder")
+
+	// Open a session to the responder
+	for range 3 {
```
mbrandenburger: what about making the number of rounds a parameter of the initiator view and let the test harness decide how many pings we want to send? That way we could also reuse the view for "longer" tests/experiments later.

Moreover, rather than hardcoding the number of rounds at the responder, we could send a "finished" message rather than tying both views together like that. Or should it be enough that the initiator closes the session once it is done? WDYT?

platform/view/services/comm/host/websocket/ws/multiplexed_provider.go
```diff
@@ -249,7 +249,13 @@ func (c *multiplexedClientConn) newClientSubConn(ctx context.Context, src host2.
...
 	}
 	logger.Debugf("Stream opened to [%s@%s]", info.RemotePeerID, info.RemotePeerAddress)
-	return NewWSStream(sc, ctx, info), nil
+	// The sub-connection is cached and reused across multiple calls (e.g. multiple rounds of
+	// a protocol), so its lifetime must not be tied to the context of the call that happened
+	// to create it: cancelling that per-call context (e.g. on successful completion) would
+	// otherwise tear down the sub-connection out from under subsequent, unrelated calls that
+	// reuse the same stream. We only propagate the span for tracing purposes, same as
+	// newServerSubConn does for the server side.
+	return NewWSStream(sc, trace.ContextWithRemoteSpanContext(context.Background(), spanContext), info), nil
```
mbrandenburger: I believe we should use `trace.ContextWithSpanContext(....)` here as we dealing with the spanContext based on the "clients" process context. Only the server should use `trace.ContextWithRemoteSpanContext(....)` to wrap the received context (via the unmarshalled metadata). Same comment for `simple_provider.go`

platform/view/services/comm/host/websocket/ws/stream.go
mbrandenburger, line 74: I think we are changing the use of `ctx` here. Previously, we were using the context to signal a stream to finish / close. Now, we are just passing a context for pure tracing to the NewWSStream. I think this is fine but we should document that the we are now relying on calling `Close` on the stream rather than using the passed context to "auto-close".

platform/view/services/comm/host/websocket/ws/context_detachment_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+func subConnCount(c *multiplexedClientConn) int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return len(c.subConns)
+}
+
+// isClosedChan reports whether ch has been closed (a closed channel yields the zero value with
+// ok=false immediately; a non-empty or open channel does not).
+func isClosedChan[T any](ch <-chan T) bool {
```
mbrandenburger, line 64: oh note that this function may not work correctly for buffered channels. If the the channel is already closed buf contains items in the buffer, this method will return a false negative. Maybe drain the channel?

platform/view/services/comm/host/websocket/ws/context_detachment_test.go
```diff
@@ -0,0 +1,292 @@
...
+	// goroutines (asserted in t.Cleanup), the subConn entry is removed, and the subConn's own
+	// resources (its channels) are released.
+	require.NoError(t, client.Close())
+	require.Equal(t, 0, subConnCount(conn))
+	requireSubConnResourcesReleased(t, sc)
+
+	// The physical websocket/TCP connection is shared by all sub-connections, so closing just
+	// this one must NOT close it - only KillAll (or the peer/connection erroring out) may.
+	require.NoError(t, conn.write(MultiplexedMessage{ID: "still-alive-probe", Msg: []byte("x")}),
+		"physical connection must stay open after a single sub-connection closes")
```
mbrandenburger: Can we check after this call that the server is actually still alive? By sending `[]byte("x")` we may let the server kill the connection already as it might not parse the message correctly (it does not contain metadata).  `p.KillAll` probably does not return an error if the the server is already down.

integration/fsc/pingpong/initiator.go
```diff
@@ -7,42 +7,80 @@ SPDX-License-Identifier: Apache-2.0
...
 			return nil, errors.New(string(msg.Payload))
 		}
-		m := string(msg.Payload)
-		if m != "pong" {
+		if m := string(msg.Payload); m != "pong" {
 			return nil, errors.Errorf("expected pong, got %s", m)
 		}
+		logger.Infof("pong received")
+	case <-viewCtx.Context().Done():
+		return nil, errors.Errorf("expected ping, got %s", viewCtx.Context().Err())
```
mbrandenburger:
```suggestion
		return nil, errors.Wrap(viewCtx.Context().Err(), "context done while waiting for pong")
```

integration/fsc/pingpong/responder.go
```diff
@@ -7,41 +7,76 @@ SPDX-License-Identifier: Apache-2.0
...
+	session := p.session
 
 	// Read the message from the initiator
 	ch := session.Receive()
 	var payload []byte
 	select {
 	case msg := <-ch:
 		payload = msg.Payload
+	case <-viewCtx.Context().Done():
+		return nil, viewCtx.Context().Err()
```
mbrandenburger:
```suggestion
		return nil,  errors.Wrap(viewCtx.Context().Err(), "context done while waiting for message")
```

mbrandenburger (changes requested):
Thank you @adecaro for creating this PR. This is a good fix. Please see my comments below for additional improvements.

platform/view/services/comm/host/websocket/ws/context_detachment_test.go
adecaro: Good catch. Fixed `isClosedChan` to drain the channel and only report closed once it observes `ok == false` — but since draining is destructive, it's now only used where that's safe: the post-close assertions in `requireSubConnResourcesReleased`, where consuming any leftovers is harmless. For the "must NOT be released" assertions on a still-live `subConn` (where draining could steal a real message), I added a non-destructive `subConnIsClosed` helper that just reads `subConn.isClosed` under its own mutex instead of touching the channels at all.

platform/view/services/comm/host/websocket/ws/context_detachment_test.go
adecaro: You're right, and it's worse than a false negative — that malformed probe (`[]byte("x")`, no metadata) hits the server's `readIncoming` on an unknown sub-conn ID, takes the `newServerSubConn` path, fails to unmarshal into `StreamMeta`, and trips the peer-identity check, so the server calls `c.Kill()` and tears down the *whole* physical connection. The later `KillAll` then can't distinguish "we closed it" from "it was already dead," which is exactly the ambiguity you're pointing at.

Replaced it with a real round-trip: open a second client stream through the public `NewClientStream` API (same URL as the original, so it reuses the cached physical connection), `sendMsg`/`readMsg` a real ping/pong against the existing `echoPingPong` server callback, then close it. That proves both the physical connection *and* the server survived the first sub-connection's `Close()`, using only the public API.

integration/fsc/pingpong/initiator.go
adecaro: Applied, adapted to what's actually being waited for here: `errors.Wrapf(viewCtx.Context().Err(), "no %s received in time", pongMessage)`. Thanks for catching that the old message said "expected ping" on the *initiator*, which was backwards on top of being generic.

integration/fsc/pingpong/responder.go
adecaro: Applied verbatim (as `errors.Wrap(viewCtx.Context().Err(), "no message received in time")`), thanks.

integration/fsc/pingpong/initiator.go
adecaro: Done — replaced with a returned, wrapped error (`errors.Wrapf(err, "failed getting session to [%s]", p.responder)`) in bcd46a67. Applied the same treatment throughout `initiator.go` and `responder.go`, and dropped the `assert` import entirely.

integration/fsc/pingpong/initiator.go
adecaro: Good catch — the two timeouts were bounding different things inconsistently: the 10-minute outer context covered `GetSession` + `NewClientStream` + `SendWithContext`, while the inner `time.After` only bounded the receive. Since the inner timer always fired first, the 10-minute value was dead code on both sides, and the responder's old 5s inner timeout was actually *tighter* than the initiator's, which is a latent flake given this PR exists precisely because a fresh stream may need to be opened mid-protocol.

Collapsed into a single `roundTimeout = 1 * time.Minute`, applied via a shared `runRound` helper (`context.WithTimeout` + `defer cancel()`) used by both the initiator and responder, and removed both inner `time.After` arms — `viewCtx.Context().Done()` now covers that case directly. 1 minute matches `comm.DefaultEnqueueTimeout` (`platform/view/services/comm/session.go`), the point at which the comm layer itself gives up on a session.

integration/fsc/pingpong/initiator.go
adecaro: Same fix as the comment above on line 30 — this now returns a wrapped error instead of asserting.

platform/view/services/comm/host/websocket/ws/stream.go
adecaro: Removed rather than gated — it was diagnostic residue from developing this fix and not worth keeping around even behind a level check. `Close()`'s debug log is back to `logger.Debugf("close connection for context [%s]", s.ContextID())`, and the now-unused `runtime/debug` import is gone.

platform/view/services/view/manager.go
adecaro: Applied verbatim, thanks.

platform/view/services/view/manager.go
adecaro: Applied verbatim, thanks.

integration/fsc/pingpong/initiator.go
adecaro: Both changes made:

- `Initiator` now embeds `Params{Rounds int}`; `<= 0` selects `DefaultRounds = 3` via a single `rounds()` accessor, and `InitiatorViewFactory.NewView` unmarshals it from the view input (only when non-empty, so the existing `nil`-input call sites keep getting the default). The REST test now exercises this explicitly with `Params{Rounds: 5}`.
- On "should closing the session be enough": it isn't, unfortunately. `NetworkStreamSession.Close()` (`platform/view/services/comm/session.go:220`) is a purely local operation and puts nothing on the wire, and `view.Message.Status` only has `OK`/`ERROR` — there's no EOF-equivalent the responder could observe. So I added an explicit terminal message: after the last round, the initiator runs a small `FinishedView` that sends a `finished` payload, and the responder loops (unbounded, no hardcoded round count) until it sees that message, a closed session, or its own round deadline. This also drops the hardcoded round count on the responder side that you flagged.

One thing worth calling out since it's not visible from the diff: the terminal message is fire-and-forget. If it were ever lost, the responder would just end on a `roundTimeout` error rather than hanging forever — no assertion currently catches that case. I considered making it self-verifying (an ack from the responder, awaited by the initiator) but that re-couples the two views in exactly the way this comment is trying to get away from, so I kept it one-way. Happy to add the ack if you'd rather have that guarantee.

platform/view/services/comm/host/websocket/ws/multiplexed_provider.go
adecaro: You're right — fixed both client sites (here and the equivalent one in `simple_provider.go`) to use `trace.ContextWithSpanContext`, since the span there comes from the local process context (`trace.SpanContextFromContext(ctx)`), not an unmarshalled remote one. Left the two server-side sites on `ContextWithRemoteSpanContext`, since those do wrap a span context unmarshalled off the wire.

Side effect worth flagging since it's not visible from the diff: `tracing.MarshalContext` (`platform/view/services/tracing/utils.go`) serializes `Remote: c.IsRemote()`, so outbound metadata will now correctly carry `remote: false` from the client, with the server still promoting it to `remote: true` on receipt.

platform/view/services/comm/host/websocket/ws/stream.go
adecaro: Added a doc comment on `NewWSStream` recording exactly that invariant: `ctx` now only seeds the tracing parent (all four call sites deliberately pass a `context.Background()`-derived context), so a stream's lifetime is governed by `Close()` (or the underlying connection erroring out), not by cancellation of the passed context. The existing `<-ctx.Done() → s.Close()` goroutine stays as a safety net but is now effectively only reachable via `Close()`'s own `s.cancel()` — the comment calls that out too, so nobody re-wires a cancellable context into it later.

mbrandenburger (changes requested):
Thank you @adecaro for the updates. Great progress. I believe we are almost ready with this PR. Can you please update the pingpong README and the `fake/` package to be consistent with the new code.

mbrandenburger (approved):

PR #1649, opened 2026-08-11
refactor(view): merge SendWithContext into Send, SendErrorWithContext into SendError

### LINKED ISSUE #1648: view.Session: collapse SendWithContext/SendErrorWithContext into Send/SendError
## Problem

`view.Session` (`platform/view/view/session.go`) carries a duplicated context/non-context method pair:

```go
Send(payload []byte) error
SendWithContext(ctx context.Context, payload []byte) error
SendError(payload []byte) error
SendErrorWithContext(ctx context.Context, payload []byte) error
```

In both real implementations the non-context variants are nothing but `context.TODO()` wrappers around the context-taking ones (`comm/session.go`, `comm/session/bidi.go`). This pair is a migration artifact:

- It doubles the interface surface.
- It forces every mock and fake to implement four methods where two suffice.
- It lets call sites silently opt out of cancellation and trace propagation by picking the shorter name.

## Proposal

Collapse each pair into the single ctx-first name. Delete the `*WithContext` variants and have `Send`/`SendError` adopt their signatures:

```go
Send(ctx context.Context, payload []byte) error
SendError(ctx context.Context, payload []byte) error
```

This is an intentional breaking API change — no deprecation shims, no compat wrappers. Downstream consumers (Panurus) pick it up on the next tag.

Scope also includes:
- `jsonSession` (`comm/session/json.go`), which has the same duplicated pair one layer up.
- `commSCCMsgConn`/`commSCCConn` (`comm/io/{msgconn,conn}.go`), which implement `io.Writer` and need a stored ctx threaded through their constructors since `Write([]byte)` has no ctx to forward.
- One semantic-risk call site: `platform/view/services/view/p2p/service.go`, which reports a responder-view failure back to the remote caller. Its context can be cancelled/reused across rounds, so it must use `context.WithoutCancel(viewCtx.Context())` rather than the live `viewCtx.Context()`, preserving today's un-cancellable `context.TODO()` behavior while still gaining trace/log value propagation.

## Outcome

One obvious way to send, cancellation and trace context propagated by construction, and two fewer methods for every implementer.

adecaro:
Closes #1648

## Summary

`view.Session` carried a duplicated context/non-context method pair (`Send`/`SendWithContext`, `SendError`/`SendErrorWithContext`). In both real implementations the non-context variants were nothing but `context.TODO()` wrappers around the context-taking ones — a migration artifact that doubled the interface surface, forced every mock/fake to implement four methods where two suffice, and let call sites silently opt out of cancellation and trace propagation.

This collapses each pair into the single ctx-first name:

```go
Send(ctx context.Context, payload []byte) error
SendError(ctx context.Context, payload []byte) error
```

This is an intentional breaking API change — no deprecation shims, no compat wrappers. Downstream consumers (Panurus) will pick it up on the next tag.

- `jsonSession` gets the same treatment one layer up (its `SendRaw(ctx, raw)` was already ctx-first and is the convention this now matches).
- `commSCCMsgConn`/`commSCCConn` thread and store a ctx through their constructors, since they implement `io.Writer` and `Write([]byte)` has no ctx to forward.
- The one semantic-risk site, `platform/view/services/view/p2p/service.go` (reporting a responder-view failure back to the remote caller), uses `context.WithoutCancel(viewCtx.Context())` instead of the live `viewCtx.Context()` — that context can be cancelled or reused across rounds via `Manager.NewResponderContext`, so passing it live could make error reporting fail with `ctx.Err()` on an otherwise healthy connection, silently dropping the error the remote caller is waiting for. `WithoutCancel` preserves today's un-cancellable behavior while still propagating trace/log values.
- Mocks regenerated via counterfeiter (`platform/view/services/view/mock/session.go`, `platform/fabric/services/endorser/mock/session.go`); hand-written test mocks and call sites updated to match.
- Docs updated (`docs/platform/view/view-api.md`, `docs/platform/view/programming-model.md`, `docs/platform/view/services/comm/readme.md`, `docs/platform/view/services/view-service.md`).

## Test plan

- [x] `go build ./... && go vet ./...` (root module)
- [x] `go build ./... && go vet ./...` (`integration` module)
- [x] `go build ./... && go vet ./...` (`platform/view/services/comm/host/libp2p` module)
- [x] `go test -count=1 ./platform/view/services/comm/... ./platform/view/services/view/... ./platform/fabric/services/endorser/... ./platform/fabric/services/state/... ./platform/view/services/id/...`
- [x] `make checks` and `make lint` clean
- [x] `make tidy` is a no-op (only new stdlib `context` imports)
- [x] `./scripts/find-mocks.sh` → all `OK`
- [x] `grep -rn 'SendWithContext\|SendErrorWithContext' --include='*.go' .` → zero hits outside test function names

### DIFF
--- a/docs/platform/view/view-api.md
+++ b/docs/platform/view/view-api.md
@@ -202,10 +202,8 @@ The View API models node-to-node communication through `view.Session`:
 ```go
 type Session interface {
     Info() SessionInfo
-    Send(payload []byte) error
-    SendWithContext(ctx context.Context, payload []byte) error
-    SendError(payload []byte) error
-    SendErrorWithContext(ctx context.Context, payload []byte) error
+    Send(ctx context.Context, payload []byte) error
+    SendError(ctx context.Context, payload []byte) error
     Receive() <-chan *Message
     Close()
 }
@@ -213,9 +211,9 @@ type Session interface {
 
 ### Send and Receive
 
-`Send` and `SendWithContext` transmit an application payload to the remote party.
+`Send` transmits an application payload to the remote party using the provided context.
 
-`SendError` and `SendErrorWithContext` transmit an error response. These methods set the message status to `view.ERROR`, allowing the recipient to distinguish protocol failures from normal payloads.
+`SendError` transmits an error response using the provided context. This method sets the message status to `view.ERROR`, allowing the recipient to distinguish protocol failures from normal payloads.
 
 The payload passed to `SendError` is transmitted as-is. When a responder view returns an error from `Call`, the P2P runtime forwards `err.Error()` verbatim to the remote caller via `SendError` — it does not inspect, wrap, or redact it. The runtime's job is only to decide *that* an error gets returned; deciding *what* information that error is allowed to contain is the application's responsibility. Since the remote party may be untrusted, responder views should return business-logic errors verbatim (callers legitimately need that detail to react, and tests rely on it), but should wrap or replace errors that could carry internal detail (stack-adjacent context, file paths, credentials) with a sanitized message before returning them.
 
--- a/docs/platform/view/view-api.md
+++ b/docs/platform/view/view-api.md
@@ -371,7 +371,7 @@ func (p *Initiator) Call(viewCtx view.Context) (any, error) {
         return nil, err
     }
 
-    if err := session.Send([]byte("ping")); err != nil {
+    if err := session.Send(viewCtx.Context(), []byte("ping")); err != nil {
         return nil, err
     }
 
@@ -401,12 +401,12 @@ func (p *Responder) Call(viewCtx view.Context) (any, error) {
     select {
     case msg := <-session.Receive():
         if string(msg.Payload) != "ping" {
-            if err := session.SendError([]byte("expected ping")); err != nil {
+            if err := session.SendError(viewCtx.Context(), []byte("expected ping")); err != nil {
                 return nil, err
             }
             return nil, errors.New("expected ping")
         }
-        if err := session.Send([]byte("pong")); err != nil {
+        if err := session.Send(viewCtx.Context(), []byte("pong")); err != nil {
             return nil, err
         }
         return "OK", nil
--- a/platform/view/services/comm/io/msgconn.go
+++ b/platform/view/services/comm/io/msgconn.go
@@ -33,6 +33,9 @@ type commSCCMsgConn struct {
 }
 
 func NewMsgConn(ctx context.Context, index int, session view.Session) (MsgConn, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
 	conn := &commSCCMsgConn{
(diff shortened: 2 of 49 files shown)


platform/view/services/comm/io/msgconn.go
```diff
@@ -30,8 +32,9 @@ type commSCCMsgConn struct {
 	readCounter  uint64
 }
 
-func NewMsgConn(index int, session view.Session) (MsgConn, error) {
+func NewMsgConn(ctx context.Context, index int, session view.Session) (MsgConn, error) {
 	conn := &commSCCMsgConn{
+		ctx:          ctx,
```
mbrandenburger, line 40: Even though `NewMsgConn` is mainly used as a testing harness, we should add a nil check to the ctx to prevent `c.session.Send` to panic during a `Write(...)`.

docs/platform/view/view-api.md
mbrandenburger: This needs to be updated as well. Can you run another grep for the "old API" and make sure we have updated this everywhere. Thanks

mbrandenburger: Same comment as above (Old API).

platform/view/services/comm/session.go
```diff
@@ -142,22 +142,13 @@ func (n *NetworkStreamSession) Info() view.SessionInfo {
 }
 
 // Send sends the payload to the endpoint.
-func (n *NetworkStreamSession) Send(payload []byte) error {
-	return n.SendWithContext(context.TODO(), payload)
-}
-
-// SendWithContext sends the payload to the endpoint with the passed context.Ctx.
-func (n *NetworkStreamSession) SendWithContext(ctx context.Context, payload []byte) error {
+// Send sends the payload to the endpoint with the passed context.Ctx.
```
mbrandenburger:
```suggestion
// Send sends the payload to the endpoint with the passed context.Context.
```

platform/view/services/comm/session.go
```diff
@@ -142,22 +142,13 @@ func (n *NetworkStreamSession) Info() view.SessionInfo {
...
 }
 
-// SendError sends an error to the endpoint with the passed payload.
-func (n *NetworkStreamSession) SendError(payload []byte) error {
-	return n.SendErrorWithContext(context.TODO(), payload)
-}
-
-// SendErrorWithContext sends an error to the endpoint with the passed context.Ctx and payload.
-func (n *NetworkStreamSession) SendErrorWithContext(ctx context.Context, payload []byte) error {
+// SendError sends an error to the endpoint with the passed context.Ctx and payload.
```
mbrandenburger:
```suggestion
// SendError sends an error to the endpoint with the passed context.Context and payload.
```

mbrandenburger (changes requested):
Thank you @adecaro for creating this PR! Please see my comments below.

mbrandenburger (approved):

PR #1651, opened 2026-08-11
feat(fabricx): make finality listener timeouts configurable

### LINKED ISSUE #1641: fabricx: make finality listener timeouts configurable via core.yaml
## Description

The Fabric-x finality listener manager hardcodes three durations that operators
will plausibly want to tune per deployment. They were introduced (or, for
`handlerTimeout`, left as-is) by #1640, which added a local expiry backstop so
cleanup of `handlers` no longer depends on the remote committer notifying.

In `platform/fabricx/core/finality/nlm.go`:

| Constant | Default | What it controls |
|---|---|---|
| `DefaultHandlerTimeout` | 5s | how long a single `OnStatus` callback may run before it is abandoned |
| `DefaultListenerTTL` | 2m | how long a listener may wait for a notification before being settled locally with `Unknown` |
| `DefaultSweepInterval` | 30s | how often expired entries are collected |

They are applied in `newNotifiWithGRPC` (`platform/fabricx/core/finality/provider.go`),
which currently ignores the `ServiceConfigProvider` that `Provider` already holds.

Why this is worth doing:

- **`listenerTTL` is a real deployment tradeoff.** It must sit comfortably above
  the timeout we request from the committer, because `notify.proto` documents that
  timeout as non-strict ("it is possible to receive notifications for this request
  after the timeout has passed"). Set it too low and a transaction that commits
  late is reported `Unknown`; too high and stale entries linger longer. The right
  value depends on committer latency, which varies by deployment.
- **`sweepInterval` trades cleanup latency against wakeups.** Worst-case entry
  lifetime is `listenerTTL + sweepInterval`.
- Reviewers on #1640 asked for these to be tunable
  ([comment](https://github.com/hyperledger-labs/fabric-smart-client/pull/1640#discussion_r3721032025)).

A related item, arguably the same change: `AddFinalityListener` hardcodes a 10s
`Timeout` on the outbound `NotificationRequest` and still carries a `// TODO: set
a proper timeout`. Per `notify.proto`, leaving that field unset makes the
committer apply its own configured default, so hardcoding 10s overrides a
server-side setting with a guess. Worth folding in or splitting out, whichever
the maintainers prefer.

## Implementation Steps

- [ ] Add the fields to `platform/fabricx/core/committer/config/config.go` — it
      already holds `RequestTimeout`, so `handlerTimeout`, `listenerTTL` and
      `sweepInterval` fit naturally alongside it.
- [ ] Keep the existing constants as the fallback defaults when the fields are
      unset, so behaviour is unchanged for deployments that do not configure them.
- [ ] Thread the resolved config into `newNotifiWithGRPC`: it takes
      `(network, GRPCClientProvider)` today and does not consult
      `Provider.configProvider`, which is already available.
- [ ] Preserve the "zero disables expiry" contract: `listenerTTL == 0` currently
      turns local expiry off entirely, and the unit tests rely on it. Decide
      deliberately whether an unset config value means "use the default" or
      "disabled" — these are not the same, and getting it wrong silently disables
      the leak fix.
- [ ] Document the new keys in `docs/configuration.md`.
- [ ] Remove the `TODO` in `newNotifiWithGRPC` that points at this issue.
- [ ] Optionally: make the outbound request `Timeout` configurable too, or leave
      it unset so the committer's own default applies.

## Additional Information

- Follow-up to #1640, which introduced the expiry backstop and the two new
  constants.
- Original leak report: #1626.
- The "zero disables expiry" behaviour is covered by
  `TestSweepExpired/Zero_TTL_Disables_Expiry`, which is the test to watch when
  changing how defaults are resolved.

SaidAltury-ibm:
Add HandlerTimeout, ListenerTTL, and SweepInterval to the notification service config so operators can tune the finality listener's local expiry backstop via core.yaml instead of relying on hardcoded defaults. Unset values fall back to the existing defaults; an explicit listenerTTL of zero continues to disable local expiry.

Also drop the hardcoded 10s outbound request timeout in AddFinalityListener, leaving it unset so the committer applies its own configured default per notify.proto.

Fixes #1641

### DIFF
--- a/docs/platform/fabric-x/configuration.md
+++ b/docs/platform/fabric-x/configuration.md
@@ -17,16 +17,21 @@ Both services support endpoint-based gRPC connectivity and optional TLS or mutua
 
 ### Notification service tuning
 
-The finality listener manager backing `notificationService` accepts various timeout configurations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
+`notificationService` configures the client used to subscribe to transaction status updates from the committer's notification service. It is what backs, for example, the finality listeners a view application registers via `fabric.Channel().Committer().AddFinalityListener(txID, listener)`: It subscribes to the transaction on the notification stream and invokes the listener's `OnStatus` callback once the committer reports an outcome.
 
 | Key | Default | What it controls |
 |-----|---------|------------------|
-| `requestTimeout` | `30s` | Sent to the committer as the outbound notification request's timeout, so it gives up and replies once it passes rather than the client aborting locally and marking transactions the committer may already know the outcome of as `Unknown`. |
-| `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned. |
-| `listenerTTL` | `2m` | How long a listener may wait for a notification before being settled locally with `Unknown`. Set it comfortably above `requestTimeout` — that timeout is documented non-strict, so a late notification can still arrive after it passes. Setting `listenerTTL: 0` explicitly disables local expiry. |
-| `sweepInterval` | `30s` | How often expired entries are collected. Worst-case entry lifetime is `listenerTTL + sweepInterval`. |
+| `endpoints[].address` | — | `host:port` of a notification service endpoint. At least one is required. |
+| `endpoints[].connectionTimeout` | `30s` | Minimum timeout for establishing the gRPC connection to that endpoint. |
+| `endpoints[].tls` | disabled | TLS settings for that endpoint: `enabled`, `rootCerts` (server TLS), plus `clientKey` / `clientCert` (mutual TLS) and `serverNameOverride`. |
+| `requestTimeout` | `30s` | How long the committer waits for a subscribed transaction before answering. It is sent to the committer as the subscription's timeout, so the committer reports the outcomes it does have and flags only the still-pending transactions as timed out — rather than the client giving up locally and treating the whole batch as `Unknown`. In practice this is how long a finality listener waits before it is settled. |
+| `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned so it cannot block the dispatcher. Callbacks that ignore context cancellation leak a goroutine. |
+| `listenerTTL` | `2m` | Local backstop: how long a listener may wait without hearing anything at all — a dead or silent stream — before FSC settles it locally with `Unknown`. Keep it comfortably above `requestTimeout`; the committer's timeout is documented non-strict, so a late notification can still arrive after it passes. `listenerTTL: 0` disables the local backstop entirely. |
+| `sweepInterval` | `30s` | How often expired listeners are collected. A listener's worst-case lifetime is `listenerTTL + sweepInterval`. Ignored when `listenerTTL` is `0`. |
 
-Example:
+Note that `Unknown` does not mean the transaction failed — only that there is no outcome yet for it within the configured bounds. A caller that needs certainty should query the transaction status directly.
+
+Complete example:
 
 ```yaml
 fabric:
@@ -34,13 +39,28 @@ fabric:
     notificationService:
       endpoints:
         - address: 127.0.0.1:5516
+          connectionTimeout: 30s
+          tls:
+            enabled: true
+            rootCerts:
+              - /path/to/ca.crt
+            # for mutual TLS, also set:
+            # clientKey: /path/to/client.key
+            # clientCert: /path/to/client.crt
       requestTimeout: 30s
       handlerTimeout: 5s
       listenerTTL: 2m
       sweepInterval: 30s
 ```
 
-Omitting any of these keys keeps the default shown above. Only `listenerTTL: 0` is treated specially — it disables local expiry rather than falling back to the default.
+Invalid timeouts are omitted, in which case the default above applies. Two zero values are meaningful rather than "unset":
+
+- `listenerTTL: 0` disables the local expiry backstop, leaving the committer's
+  reply as the only thing that ever settles a listener.
+- `requestTimeout: 0` delegates the subscription timeout to the committer's own
+  configuration rather than to the `30s` above.
+
+`handlerTimeout: 0` and `sweepInterval: 0` have no such meaning and fall back to their defaults.
 
 ## Related Documentation
 
--- a/docs/platform/fabric-x/configuration.md
+++ b/docs/platform/fabric-x/configuration.md
@@ -13,7 +13,33 @@ The Fabric-x platform currently documents configuration for these services:
 - `fabric.<network>.notificationService`
 - `fabric.<network>.queryService`
 
-Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings.
+Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings, plus a `requestTimeout` for outbound gRPC calls (default `30s`).
+
+### Finality listener tuning (`notificationService` only)
+
+The finality listener manager backing `notificationService` accepts three additional durations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
+
+| Key | Default | What it controls |
+|-----|---------|------------------|
+| `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned. |
+| `listenerTTL` | `2m` | How long a listener may wait for a notification before being settled locally with `Unknown`. Set it comfortably above the committer's own notification timeout — that timeout is documented non-strict, so a late notification can still arrive after it passes. Setting `listenerTTL: 0` explicitly disables local expiry. |
+| `sweepInterval` | `30s` | How often expired entries are collected. Worst-case entry lifetime is `listenerTTL + sweepInterval`. |
+
+Example:
+
+```yaml
+fabric:
+  <network>:
+    notificationService:
+      endpoints:
+        - address: 127.0.0.1:5516
+      requestTimeout: 30s
+      handlerTimeout: 5s
+      listenerTTL: 2m
+      sweepInterval: 30s
+```
+
+Omitting any of these keys keeps the default shown above. Only `listenerTTL: 0` is treated specially — it disables local expiry rather than falling back to the default.
 
 ## Related Documentation
 
--- a/docs/platform/fabric-x/configuration.md
+++ b/docs/platform/fabric-x/configuration.md
@@ -15,14 +15,15 @@ The Fabric-x platform currently documents configuration for these services:
 
 Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings, plus a `requestTimeout` for outbound gRPC calls (default `30s`).
 
-### Finality listener tuning (`notificationService` only)
+### Notification service tuning
 
-The finality listener manager backing `notificationService` accepts three additional durations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
+The finality listener manager backing `notificationService` accepts various timeout configurations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
 
 | Key | Default | What it controls |
 |-----|---------|------------------|
+| `requestTimeout` | `30s` | Sent to the committer as the outbound notification request's timeout, so it gives up and replies once it passes rather than the client aborting locally and marking transactions the committer may already know the outcome of as `Unknown`. |
 | `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned. |
-| `listenerTTL` | `2m` | How long a listener may wait for a notification before being settled locally with `Unknown`. Set it comfortably above the committer's own notification timeout — that timeout is documented non-strict, so a late notification can still arrive after it passes. Setting `listenerTTL: 0` explicitly disables local expiry. |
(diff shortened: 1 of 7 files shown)


docs/platform/fabric-x/configuration.md
```diff
@@ -13,7 +13,33 @@ The Fabric-x platform currently documents configuration for these services:
...
+
+### Finality listener tuning (`notificationService` only)
+
+The finality listener manager backing `notificationService` accepts three additional durations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
+
+| Key | Default | What it controls |
+|-----|---------|------------------|
+| `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned. |
+| `listenerTTL` | `2m` | How long a listener may wait for a notification before being settled locally with `Unknown`. Set it comfortably above the committer's own notification timeout — that timeout is documented non-strict, so a late notification can still arrive after it passes. Setting `listenerTTL: 0` explicitly disables local expiry. |
+| `sweepInterval` | `30s` | How often expired entries are collected. Worst-case entry lifetime is `listenerTTL + sweepInterval`. |
```
mbrandenburger: I think we should also explain `requestTimeout` in more detail in this table as it is only mentioned very briefly above. This also then make the example configuration "complete". WDYT?

docs/platform/fabric-x/configuration.md
```diff
@@ -13,7 +13,33 @@ The Fabric-x platform currently documents configuration for these services:
 - `fabric.<network>.notificationService`
 - `fabric.<network>.queryService`
 
-Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings.
+Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings, plus a `requestTimeout` for outbound gRPC calls (default `30s`).
+
+### Finality listener tuning (`notificationService` only)
+
+The finality listener manager backing `notificationService` accepts three additional durations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
```
mbrandenburger:
```suggestion
The finality listener manager backing `notificationService` accepts various timeout configurations. They control the local expiry backstop that settles a finality listener with `Unknown` if the committer never notifies it (see `platform/fabricx/core/finality/nlm.go`):
```

docs/platform/fabric-x/configuration.md
```diff
@@ -13,7 +13,33 @@ The Fabric-x platform currently documents configuration for these services:
 - `fabric.<network>.notificationService`
 - `fabric.<network>.queryService`
 
-Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings.
+Both services support endpoint-based gRPC connectivity and optional TLS or mutual TLS settings, plus a `requestTimeout` for outbound gRPC calls (default `30s`).
+
+### Finality listener tuning (`notificationService` only)
```
mbrandenburger: The heading is a bit misleading as it introduces a name that is not part of the actual config. That is, a user searching for finality listener in a config file will not find anything. 

I suggest we call this subsection "Notification service tuning" and then explain all the possible configuration noobs. To give an example where these settings are used, we can mention the finality listener, which can be used in a view application.

platform/fabricx/core/finality/nlm.go
```diff
@@ -275,8 +274,10 @@ func (n *notificationListenerManager) AddFinalityListener(txID driver.TxID, list
 		TxStatusRequest: &committerpb.TxIDsBatch{
 			TxIds: txIDs,
 		},
-		// TODO: set a proper timeout
-		Timeout: durationpb.New(10 * time.Second),
+		// Timeout deliberately left unset: notify.proto has the committer apply
+		// its own configured default when this field is absent.
+		// The committer operator is in a better position to know
+		// the right timeout for their network than FSC's client code is
```
mbrandenburger: I think we should use our `requestTimeout` here. If we do not set the timeout, the committer uses it's internal max-timeout. By setting the timeout, we would tell the committer to return a response once the timeout has passed. 

See docs: https://github.com/hyperledger/fabric-x-committer/blob/main/docs/notification-service.md#22-subscribing-to-transaction-status-updates

Let's think about the following example: we set our client timeout to 1second and the committers timeout is set to 10seconds. We asks for some notifications which where a single tx needs 4 seconds to be completed. By not setting the timeout in the request, our client will just fire timeout and abort the grpc request and mark all tx statuses as unknown. In that situation we client marks tx statuses even though some of them are already known to the committer. By setting the timeout to 1second, the committer will respond with all status but just mark the last tx as unknown. 

I believe this example shows that setting the timeout as client makes sense to get more accurate information. However, the right timing might be difficult as we would need to wait longer than requestTimeout to locally abort the grpc call. WDYT?

Looking at the code more closely, it seems that we are not using the `requestTimeout` from the config at all anymore - though we should?!

platform/fabricx/core/finality/provider.go
```diff
@@ -144,19 +156,25 @@ func newNotifiWithGRPC(network string, grpcClientProvider GRPCClientProvider) (*
...
+	handlerTimeout, listenerTTL, sweepInterval := DefaultHandlerTimeout, DefaultListenerTTL, DefaultSweepInterval
+	if cfg != nil {
+		listenerTTL = cfg.ListenerTTL
+		if cfg.HandlerTimeout > 0 {
+			handlerTimeout = cfg.HandlerTimeout
+		}
+		if cfg.SweepInterval > 0 {
+			sweepInterval = cfg.SweepInterval
+		}
+	}
```
mbrandenburger: I believe this should not happen here. We should construct a valid config outside this method (or even package) and just consume the parameters here. `cfg *config.Config` could also be passed as value rather than as reference, so we do not have to deal with defaults and nil checks at all here.

platform/fabricx/core/committer/config/config.go
```diff
@@ -15,12 +15,38 @@ import (
...
+const DefaultHandlerTimeout = 5 * time.Second
+
+// DefaultListenerTTL bounds how long a finality listener may wait for a
+// notification that may never arrive. See finality.DefaultListenerTTL for the
+// rationale.
+const DefaultListenerTTL = 2 * time.Minute
+
+// DefaultSweepInterval is how often expired finality listener entries are
+// collected. See finality.DefaultSweepInterval for the rationale.
+const DefaultSweepInterval = 30 * time.Second
```
mbrandenburger: Note that we also configure these defaults in the `finality` package. Do we need these defaults here? Or better, should we have the defaults here and not in the finiality package at all?

platform/fabricx/core/finality/provider.go
```diff
@@ -104,16 +104,28 @@ func (p *Provider) NewManager(network, channel string) (ListenerManager, error)
...
+	// is nil in some unit tests that never reach here for a real network; nil cfg
+	// is what newNotifiWithGRPC below maps to today's hardcoded defaults.
+	var cfg *config.Config
+	if p.configProvider != nil {
+		var err error
+		cfg, err = p.configProvider.NotificationServiceConfig(network)
+		if err != nil {
+			return nil, errors.Wrapf(err, "get notification service config [network=%s]", network)
+		}
+	}
```
mbrandenburger: this should better happen before we acquire `p.managersMu.Lock()` as parsing the config might be "expensive".

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for working on this PR. Please see my comments for further improvements.

platform/fabricx/core/finality/nlm.go
SaidAltury-ibm: Done in 44329da7 — the outbound `NotificationRequest` now sets `Timeout: durationpb.New(n.requestTimeout)`, so the committer replies once it passes instead of us aborting the gRPC call locally and marking already-known transactions as `Unknown`.

platform/fabricx/core/committer/config/config.go
SaidAltury-ibm: Agreed, done in 44329da7 — the Default* constants now live only in `committer/config`; `finality` consumes them via `config.Config` instead of keeping its own copies.

platform/fabricx/core/finality/provider.go
SaidAltury-ibm: Fixed in 44329da7. `newNotifiWithGRPC` now takes `cfg config.Config` by value instead of `*config.Config`, and the caller always passes a fully-resolved config (either from `ServiceConfigProvider` or `config.DefaultConfig()`), so there's no nil/zero-value handling left downstream.

docs/platform/fabric-x/configuration.md
SaidAltury-ibm: Renamed to "Notification service tuning" in 44329da7.

docs/platform/fabric-x/configuration.md
SaidAltury-ibm: Applied in 44329da7 (reworded to also mention requestTimeout, since the intro now covers all four keys, not just the local-expiry ones).

docs/platform/fabric-x/configuration.md
SaidAltury-ibm: Added in 44329da7 — the table now has a requestTimeout row.

platform/fabricx/core/committer/config/config.go
```diff
@@ -67,16 +113,34 @@ type ServiceBackend interface {
...
+// unlike ListenerTTL, they have no "zero disables it" meaning, so a zero value
+// would otherwise hand every listener an already-expired context. Callers can
+// use every field as-is, with no further nil or zero-value handling.
 func NewNotificationServiceConfig(configService ServiceBackend) (*Config, error) {
 	config := &Config{
 		RequestTimeout: DefaultRequestTimeout,
+		HandlerTimeout: DefaultHandlerTimeout,
+		ListenerTTL:    DefaultListenerTTL,
+		SweepInterval:  DefaultSweepInterval,
 	}
```
mbrandenburger: should we use `config := DefaultConfig()` here?

docs/platform/fabric-x/configuration.md
```diff
@@ -13,7 +13,34 @@ The Fabric-x platform currently documents configuration for these services:
...
+    notificationService:
+      endpoints:
+        - address: 127.0.0.1:5516
+      requestTimeout: 30s
+      handlerTimeout: 5s
+      listenerTTL: 2m
+      sweepInterval: 30s
+```
+
+Omitting any of these keys keeps the default shown above. Only `listenerTTL: 0` is treated specially — it disables local expiry rather than falling back to the default.
```
mbrandenburger: ### Notification service tuning

`notificationService` configures the client used to subscribe to transaction status updates from the committer's notification service. It is what backs, for example, the finality listeners a view application registers via `fabric.Channel().Committer().AddFinalityListener(txID, listener)`: It subscribes to the transaction on the notification stream and invokes the listener's `OnStatus` callback once the committer reports an outcome.

| Key | Default | What it controls |
|-----|---------|------------------|
| `endpoints[].address` | — | `host:port` of a notification service endpoint. At least one is required. |
| `endpoints[].connectionTimeout` | `30s` | Minimum timeout for establishing the gRPC connection to that endpoint. |
| `endpoints[].tls` | disabled | TLS settings for that endpoint: `enabled`, `rootCerts` (server TLS), plus `clientKey` / `clientCert` (mutual TLS) and `serverNameOverride`. |
| `requestTimeout` | `30s` | How long the committer waits for a subscribed transaction before answering. It is sent to the committer as the subscription's timeout, so the committer reports the outcomes it does have and flags only the still-pending transactions as timed out — rather than the client giving up locally and treating the whole batch as `Unknown`. In practice this is how long a finality listener waits before it is settled. |
| `handlerTimeout` | `5s` | How long a single `OnStatus` callback may run before it is abandoned so it cannot block the dispatcher. Callbacks that ignore context cancellation leak a goroutine. |
| `listenerTTL` | `2m` | Local backstop: how long a listener may wait without hearing anything at all — a dead or silent stream — before FSC settles it locally with `Unknown`. Keep it comfortably above `requestTimeout`; the committer's timeout is documented non-strict, so a late notification can still arrive after it passes. `listenerTTL: 0` disables the local backstop entirely. |
| `sweepInterval` | `30s` | How often expired listeners are collected. A listener's worst-case lifetime is `listenerTTL + sweepInterval`. Ignored when `listenerTTL` is `0`. |

Note that `Unknown` does not mean the transaction failed — only that there is no outcome yet for it within the configured bounds. A caller that needs certainty should query the transaction status directly.

Complete example:

```yaml
fabric:
  <network>:
    notificationService:
      endpoints:
        - address: 127.0.0.1:5516
          connectionTimeout: 30s
          tls:
            enabled: true
            rootCerts:
              - /path/to/ca.crt
            # for mutual TLS, also set:
            # clientKey: /path/to/client.key
            # clientCert: /path/to/client.crt
      requestTimeout: 30s
      handlerTimeout: 5s
      listenerTTL: 2m
      sweepInterval: 30s
```

Invalid timeouts are omitted, in which case the default above applies. Two zero values are meaningful rather than "unset":

- `listenerTTL: 0` disables the local expiry backstop, leaving the committer's
  reply as the only thing that ever settles a listener.
- `requestTimeout: 0` delegates the subscription timeout to the committer's own
  configuration rather than to the `30s` above.

`handlerTimeout: 0` and `sweepInterval: 0` have no such meaning and fall back to their defaults.

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for updating the PR. I think this is almost ready. Please check my suggestion to update the docs.

platform/fabricx/core/finality/provider.go
SaidAltury-ibm: Fixed in 39e58988 — config resolution now happens before acquiring `managersMu`, and the locking itself is back to a single `Lock()`/`defer Unlock()` spanning the existence-check, construct, and register steps (no more two-phase lock).

mbrandenburger (approved):

PR #1662, opened 2026-08-20
fix(fabricx): bound finality handler concurrency with errgroup SetLimit

### LINKED ISSUE #1627: fabricx: notification dispatcher spawns two unbounded goroutines per finality notification
## Summary

For every `(txID, listener)` pair in an inbound notification, the dispatcher spawns two goroutines
with no worker pool, semaphore, or queue. The 5 s handler timeout bounds only the outer goroutine;
the inner one is documented to leak outright if a listener ignores its context. Under sustained
load a single misbehaving listener converts the notification rate directly into unbounded goroutine
growth.

## Problem statement

`platform/fabricx/core/finality/nlm.go:144-162`:

```go
for _, c := range calls {
	go func() {                                              // outer
		timeoutCtx, cancel := context.WithTimeout(gCtx, n.handlerTimeout)
		defer cancel()

		done := make(chan struct{})
		go func() {                                          // inner
			c.handler.OnStatus(timeoutCtx, c.txID, c.status, "")
			close(done)
		}()

		select {
		case <-done:
			// Handler completed within timeout
		case <-timeoutCtx.Done():
			logger.Warnf("OnStatus handler timed out for txID=%s (timeout=%s)", c.txID, n.handlerTimeout)
		}
	}()
}
```

The design intent — never block the dispatcher on a slow listener — is right, and the leak is
acknowledged in the code at `nlm.go:141-143`:

```go
// Invoke each handler in its own goroutine with a timeout.
// If a handler ignores the context and never returns, the goroutine
// will leak — but the dispatcher remains unblocked.
```

The problem is that the chosen mechanism converts listener misbehaviour into an unbounded
resource leak rather than into backpressure.

### Two regimes

**Slow listeners — bounded.** If listeners take the full `DefaultHandlerTimeout` (5 s,
`nlm.go:33`), concurrency settles at roughly `notification_rate × 5 s`. At the ~300 tx/s described
in #1599 that is ~1 500 outer + ~1 500 inner ≈ 3 000 goroutines. At a few KB of stack each that is
on the order of 10 MB — undesirable but self-limiting.

**Blocked listeners — unbounded.** If a listener never returns (blocked on a full unbuffered
channel, a stalled store call, a contended lock), the outer goroutine exits after 5 s and logs a
warning, but the inner goroutine is pinned for the lifetime of the process, holding its stack plus
`c.handler`, `c.txID` and `timeoutCtx`. At ~300 tx/s that is on the order of 10^6 leaked goroutines
per hour. The only signal is a warning log emitted at the notification rate, which is easy to miss.

These figures are order-of-magnitude estimates derived from reading the code, not measurements.

### Secondary point: burst behaviour

`calls` is assembled by draining `handlers` under the lock and spawning afterwards
(`nlm.go:126-139`). That is good for lock hold time, but it means one notification response
carrying N txIDs spawns 2N goroutines in a single burst with no pacing.

### What is *not* a problem

The timeout contexts do not accumulate: `defer cancel()` always runs in the outer goroutine, so each
child context is detached from `gCtx`. The leak is purely the inner goroutines and what they
capture.

## Proposed fix

Replace spawn-per-call with a fixed-size worker pool reading from a bounded channel. There is
already an in-tree pattern for exactly this on the generic side —
`platform/common/core/generic/committer/finality.go`:

```go
// :23
defaultEventQueueSize = 1000

// :56
eventQueue: make(chan driver.FinalityEvent[V], defaultEventQueueSize),

// :76-80
func (c *FinalityManager[V]) Run(ctx context.Context) {
	for i := 0; i < c.eventQueueWorkers; i++ {
		go c.runEventQueue(ctx)
	}
	go c.runStatusListener(ctx)
}
```

Concretely for `notificationListenerManager`:

1. Add a buffered `callQueue chan handlerCall` and a `handlerWorkers int`, both defaulted alongside
   `DefaultHandlerTimeout` (`nlm.go:33`) and wired in `newNotifiWithGRPC`
   (`provider.go:147-153`).
2. In `listen()`, start `handlerWorkers` goroutines in the existing `errgroup` (`nlm.go:61`) so they
   are torn down with the stream. Each worker loops over `callQueue` and invokes
   `OnStatus` with the same per-call timeout as today.
3. The dispatcher (`nlm.go:144-162`) then only enqueues, so goroutine count is `O(workers)` rather
   than `O(notifications)`.

This bounds concurrency and gives a place to apply backpressure. To preserve the current
"never block the dispatcher" property, a full queue should not block — prefer dropping with a
counter/metric and a rate-limited warning over unbounded spawning, since a full queue already means
listeners are not keeping up.

Keeping the timeout requires care: a worker that calls a blocking listener directly is itself stuck,
which is what the pool is meant to prevent. Two workable options:

- **Accept it**: workers block, the queue applies backpressure, and a stuck listener degrades
  finality notification throughput rather than exhausting memory. Simplest, and arguably the correct
  failure mode.
- **Keep the watchdog**: workers retain the inner-goroutine pattern so they stay responsive, but
  because worker count is capped, the number of leaked inner goroutines is capped at `handlerWorkers`
  concurrent stuck calls rather than growing with the notification rate.

Either is a large improvement over the status quo; the first is simpler and I'd suggest starting
there.

Whichever is chosen, a gauge for in-flight handler invocations and a counter for timeouts/drops
would make the failure visible in Grafana next to the heap panels already used in #1599.

## Testing

- With a listener that blocks indefinitely, deliver N notifications and assert
  `runtime.NumGoroutine()` stays bounded (today it grows linearly with N).
- Assert a slow-but-completing listener still receives `OnStatus` and does not stall the dispatcher
  for other txIDs.
- Assert workers are torn down when `listen()` returns (no goroutine growth across
  stream restarts).

## Code references

- `platform/fabricx/core/finality/nlm.go:144-162` — the spawn-per-call loop (two goroutines per call)
- `platform/fabricx/core/finality/nlm.go:141-143` — the comment acknowledging the leak
- `platform/fabricx/core/finality/nlm.go:29-33` — `DefaultHandlerTimeout`, and its note that ignoring cancellation leaks goroutines
- `platform/fabricx/core/finality/nlm.go:106-139` — dispatcher and the `calls` batch assembled under lock
- `platform/fabricx/core/finality/nlm.go:54-61` — `listen()` and the `errgroup` that should own the workers
- `platform/fabricx/core/finality/provider.go:147-153` — `newNotifiWithGRPC`, where pool size would be wired
- `platform/common/core/generic/committer/finality.go:23,45-60,76-90` — the existing bounded worker-pool pattern to follow

SaidAltury-ibm:
This PR addresses the unbounded goroutine growth reported in #1627.

The notification dispatcher spawned two goroutines per `(txID, listener)` pair: an
outer one to enforce `handlerTimeout` and an inner one to run `OnStatus`. The timeout
bounded only the outer goroutine, so a listener that ignored context cancellation left
the inner one pinned for the lifetime of the process, turning the notification rate
directly into a memory leak.

Listener callbacks now run on a bounded pool: an `errgroup` whose `SetLimit` is the new
`handlerWorkers` setting (default 16), fed by a buffered queue sized by the new
`handlerQueueSize` (default 1000). The two do different jobs. `handlerWorkers` caps how
many callbacks run at once, which is what stops a misbehaving listener from growing
goroutines without end. The queue absorbs bursts: a notification response can carry far
more transactions than there are slots, so without a buffer everything past the limit
would be dropped even though the listeners are healthy and about to free their slots.
The dispatcher never blocks — it shares a goroutine with the stream reader and the
expiry sweeper — so it only enqueues, and a dedicated feeder moves work into the pool as
slots free. Once the queue is full, callbacks are dropped with one aggregated warning
per batch and those listeners are settled with `Unknown` by the `listenerTTL` sweeper.

The callbacks live in a **separate** errgroup from the three stream goroutines, for
three reasons: `Wait()` returns only once every goroutine in the group has returned, so
a stuck listener would hang `listen()` and node shutdown with it; `SetLimit` applies to
the whole group, so handlers filling it would starve the stream goroutines; and errgroup
cancels its context on the first error, so a handler error would tear down the stream.
For the same reason the handler group's own `Wait()` is bounded by `handlerTimeout` on
teardown, and `settleAllAndClear` — which runs after that wait, when slots may still be
held — invokes listeners directly with a per-listener timeout rather than through the
pool.

Because a stuck listener now degrades throughput rather than leaking memory, "`OnStatus`
must observe `ctx.Done()` and return promptly" becomes a requirement rather than advice,
so that warning is now on the `FinalityListener` interface itself.

Tested with eight tests in `nlm_group_test.go` covering the limit, slot reuse, full
delivery of a burst larger than the limit, drops under saturation, and both teardown
paths, plus constructor and config tests for the new keys. Verified with `-race` over
repeated runs; `make lint`, `govet`, `ineffassign` and `staticcheck` report no issues.

Closes #1627

### DIFF
--- a/docs/platform/fabric-x/README.md
+++ b/docs/platform/fabric-x/README.md
@@ -52,10 +52,20 @@ Obtain a `ListenerManager` via `finality.GetListenerManager(sp, network, channel
 
 The `OnStatus` callback receives one of: `fdriver.Valid` (committed), `fdriver.Invalid` (rejected), or `fdriver.Unknown` (undetermined / timeout).
 
+> [!WARNING]
+> `OnStatus` implementations MUST observe `ctx.Done()` and return promptly.
+> Callbacks run on an `errgroup` limited to `handlerWorkers` (default 16) concurrent
+> calls, so one that blocks forever holds a slot indefinitely — and once all slots are
+> held, finality notifications stop being delivered. Hand slow work to your own queue
+> and return. See [the notification service tuning
+> guide](configuration.md#notification-service-tuning) for the full failure mode.
+
 ### Limitations
 
 - **No automatic reconnection**: if the stream breaks, the manager is removed and registered listeners are lost. A new manager is created on the next `GetListenerManager` call.
-- **Handler timeout**: callbacks that exceed 5 seconds (`DefaultHandlerTimeout`) are abandoned with a warning. Handlers that ignore context cancellation will leak a goroutine but won't block other listeners.
+- **Bounded handler concurrency**: callbacks run on an `errgroup` with `SetLimit(handlerWorkers)` (default 16), started via `TryGo`. `handlerTimeout` (default 5s) only cancels the callback's context — it cannot force a return, so a callback that ignores cancellation holds its slot indefinitely. With every slot held, callbacks are dropped with a warning and the affected listeners are settled with `Unknown` by the `listenerTTL` sweeper. This bounds the blast radius of a misbehaving listener to throughput rather than memory; see [notification service tuning](configuration.md#notification-service-tuning).
+- **No queue in front of the limit**: `TryGo` never blocks the dispatcher, so a notification batch larger than `handlerWorkers` loses its overflow even when every listener is fast. Raise `handlerWorkers` above the largest expected batch if that matters; see [bursts and `handlerWorkers`](configuration.md#bursts-and-handlerworkers).
+- **Teardown does not wait on stuck handlers**: in-flight callbacks get a bounded window to finish after the stream stops; one still inside an unresponsive listener can outlive the stream, so a single bad listener cannot hang node shutdown. Such callbacks are capped at `handlerWorkers`.
 
 ### Configuration
 
@@ -66,6 +76,7 @@ fabric:
   <network>:
     notificationService:
       requestTimeout: 30s
+      handlerWorkers: 16
       endpoints:
         - address: "committer.example.com:9090"
           connectionTimeout: 5s
--- a/docs/platform/fabric-x/README.md
+++ b/docs/platform/fabric-x/README.md
@@ -53,18 +53,17 @@ Obtain a `ListenerManager` via `finality.GetListenerManager(sp, network, channel
 The `OnStatus` callback receives one of: `fdriver.Valid` (committed), `fdriver.Invalid` (rejected), or `fdriver.Unknown` (undetermined / timeout).
 
 > [!WARNING]
-> `OnStatus` implementations MUST observe `ctx.Done()` and return promptly.
-> Callbacks run on an `errgroup` limited to `handlerWorkers` (default 16) concurrent
-> calls, so one that blocks forever holds a slot indefinitely — and once all slots are
-> held, finality notifications stop being delivered. Hand slow work to your own queue
-> and return. See [the notification service tuning
+> `OnStatus` implementations MUST observe `ctx.Done()` and return promptly. At most
+> `handlerWorkers` (default 16) callbacks run at once, so one that blocks forever
+> occupies a slot indefinitely — and once all slots are occupied, finality
+> notifications stop being delivered. Hand slow work to your own queue and return.
+> See [the notification service tuning
 > guide](configuration.md#notification-service-tuning) for the full failure mode.
 
 ### Limitations
 
 - **No automatic reconnection**: if the stream breaks, the manager is removed and registered listeners are lost. A new manager is created on the next `GetListenerManager` call.
-- **Bounded handler concurrency**: callbacks run on an `errgroup` with `SetLimit(handlerWorkers)` (default 16), started via `TryGo`. `handlerTimeout` (default 5s) only cancels the callback's context — it cannot force a return, so a callback that ignores cancellation holds its slot indefinitely. With every slot held, callbacks are dropped with a warning and the affected listeners are settled with `Unknown` by the `listenerTTL` sweeper. This bounds the blast radius of a misbehaving listener to throughput rather than memory; see [notification service tuning](configuration.md#notification-service-tuning).
-- **No queue in front of the limit**: `TryGo` never blocks the dispatcher, so a notification batch larger than `handlerWorkers` loses its overflow even when every listener is fast. Raise `handlerWorkers` above the largest expected batch if that matters; see [bursts and `handlerWorkers`](configuration.md#bursts-and-handlerworkers).
+- **Bounded handler concurrency**: at most `handlerWorkers` (default 16) callbacks run concurrently, buffered by a queue of `handlerQueueSize` (default 1000) so bursts larger than the limit are still delivered in full. `handlerTimeout` (default 5s) only cancels the callback's context — it cannot force a return, so a callback that ignores cancellation occupies its slot indefinitely. Once every slot is occupied and the queue fills, callbacks are dropped with a warning and the affected listeners are settled with `Unknown` by the `listenerTTL` sweeper. This bounds the blast radius of a misbehaving listener to throughput rather than memory; see [notification service tuning](configuration.md#notification-service-tuning).
 - **Teardown does not wait on stuck handlers**: in-flight callbacks get a bounded window to finish after the stream stops; one still inside an unresponsive listener can outlive the stream, so a single bad listener cannot hang node shutdown. Such callbacks are capped at `handlerWorkers`.
 
 ### Configuration
@@ -77,6 +76,7 @@ fabric:
     notificationService:
       requestTimeout: 30s
       handlerWorkers: 16
+      handlerQueueSize: 1000
       endpoints:
         - address: "committer.example.com:9090"
           connectionTimeout: 5s
--- a/platform/fabricx/core/committer/config/config_test.go
+++ b/platform/fabricx/core/committer/config/config_test.go
@@ -59,6 +59,7 @@ func TestNewNotificationServiceConfig(t *testing.T) {
 		cfg, err := config.NewNotificationServiceConfig(fakeConfigService)
 		require.NoError(t, err)
 		require.Equal(t, config.DefaultHandlerTimeout, cfg.HandlerTimeout)
+		require.Equal(t, config.DefaultHandlerWorkers, cfg.HandlerWorkers)
 		require.Equal(t, config.DefaultListenerTTL, cfg.ListenerTTL)
 		require.Equal(t, config.DefaultSweepInterval, cfg.SweepInterval)
 	})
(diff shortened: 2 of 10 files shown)


platform/fabricx/core/finality/provider.go
```diff
@@ -173,11 +179,19 @@ func newNotifiWithGRPC(network string, grpcClientProvider GRPCClientProvider, cf
...
+			handlerWorkers, network, config.DefaultHandlerWorkers)
+		handlerWorkers = config.DefaultHandlerWorkers
+	}
+
 	nlm := &notificationListenerManager{
 		notifyClient:   notifyClient,
 		requestQueue:   make(chan *committerpb.NotificationRequest),  // Queue for outgoing requests to the committer
 		responseQueue:  make(chan *committerpb.NotificationResponse), // Queue for incoming responses/notifications
 		handlers:       make(map[driver.TxID]*handlerEntry),          // Map: txID -> listeners + local expiry deadline
+		handlerWorkers: handlerWorkers,
```
mbrandenburger:
```suggestion
		handlerWorkers: cfg.HandlerWorkers,
```

platform/fabricx/core/finality/provider.go
```diff
@@ -173,11 +179,19 @@ func newNotifiWithGRPC(network string, grpcClientProvider GRPCClientProvider, cf
 	// Create the gRPC client stub for the Notifier service
 	notifyClient := committerpb.NewNotifierClient(cc)
 
+	handlerWorkers := cfg.HandlerWorkers
+	if handlerWorkers <= 0 {
+		logger.Warnf("invalid handlerWorkers=%d for network=%s, falling back to %d",
+			handlerWorkers, network, config.DefaultHandlerWorkers)
+		handlerWorkers = config.DefaultHandlerWorkers
+	}
```
mbrandenburger: Config validation already happens in `committer/config/config.go`?

platform/fabricx/core/finality/provider_test.go
```diff
@@ -587,3 +588,52 @@ func TestGetListenerManager(t *testing.T) {
...
+		t.Parallel()
+		cfg := config.DefaultConfig()
+		cfg.HandlerWorkers = 3
+
+		nlm, err := newNotifiWithGRPC("network1", &mockGRPCClientProvider{}, cfg)
+		require.NoError(t, err)
+		require.Equal(t, 3, nlm.handlerWorkers)
+	})
+
+	t.Run("non-positive handlerWorkers does not produce an unusable manager", func(t *testing.T) {
```
mbrandenburger: Wondering if that test is redundant with `negative handlerWorkers falls back to default` in `config_test.go`

platform/fabricx/core/committer/config/config_test.go
```diff
@@ -93,6 +95,25 @@ func TestNewNotificationServiceConfig(t *testing.T) {
...
+		fakeConfigService.UnmarshalKeyStub = func(key string, rawVal any) error {
+			if cfg, ok := rawVal.(**config.Config); ok {
+				(*cfg).HandlerWorkers = -1
+			}
+			return nil
+		}
+
+		cfg, err := config.NewNotificationServiceConfig(fakeConfigService)
+		require.NoError(t, err)
+		// errgroup.SetLimit panics on a negative limit, so this must be sanitized.
```
mbrandenburger: I think this is not true with the version of `errgroup` we are using. Can you please double check this. Also note that in one of your tests, there is a NoPanic assertion, which contradicts this statement here.

```
// SetLimit limits the number of active goroutines in this group to at most n.
// A negative value indicates no limit.
// A limit of zero will prevent any new goroutines from being added.
```

docs/platform/fabric-x/README.md
```diff
@@ -52,10 +52,20 @@ Obtain a `ListenerManager` via `finality.GetListenerManager(sp, network, channel
...
+> and return. See [the notification service tuning
+> guide](configuration.md#notification-service-tuning) for the full failure mode.
+
 ### Limitations
 
 - **No automatic reconnection**: if the stream breaks, the manager is removed and registered listeners are lost. A new manager is created on the next `GetListenerManager` call.
-- **Handler timeout**: callbacks that exceed 5 seconds (`DefaultHandlerTimeout`) are abandoned with a warning. Handlers that ignore context cancellation will leak a goroutine but won't block other listeners.
+- **Bounded handler concurrency**: callbacks run on an `errgroup` with `SetLimit(handlerWorkers)` (default 16), started via `TryGo`. `handlerTimeout` (default 5s) only cancels the callback's context — it cannot force a return, so a callback that ignores cancellation holds its slot indefinitely. With every slot held, callbacks are dropped with a warning and the affected listeners are settled with `Unknown` by the `listenerTTL` sweeper. This bounds the blast radius of a misbehaving listener to throughput rather than memory; see [notification service tuning](configuration.md#notification-service-tuning).
```
mbrandenburger: I suggest to remove implementation details. I believe it is not relevant in the docs here that we use `errgroup`.

mbrandenburger: Moreover, I believe we need some additional mechanism here to allow some back pressure. 

If we get a batch of notifications that is larger than the number of `handlerWorkers`, this will cause dropping callbacks already. Thus, a more robust way would be to combine our `errgroup` (to limit the number of workers) with a queue to buffer some notifications (tolerate burst notifications better). 

I am wondering if `errgroup` is indeed the "right" tool if combined with a queue to feed the worker. HAHA I know I suggested the `errgroup`.

WDYT?

platform/fabricx/core/finality/nlm_group_test.go
```diff
@@ -0,0 +1,441 @@
...
+		require.EventuallyWithT(t, func(collect *assert.CollectT) {
+			assert.Equal(collect, int32(1), stuck.inFlight.Load())
+		}, timeout, tick, "the stuck callback should be in flight")
+
+		cancel()
+		select {
+		case <-listenErr:
+		case <-time.After(2 * time.Second):
+			t.Fatal("listen() did not return: teardown blocked on a callback that ignores its context")
+		}
```
mbrandenburger, line 387: Additionally we should check for `t.Context` cancellation. Please check the other tests as well.

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for creating this PR. Please see my comments below.

docs/platform/fabric-x/README.md
SaidAltury-ibm: You're right to doubt it
I think The right tool is N long-lived workers reading the queue. concurrency bounded by construction, no limiter, no polling,  right now  only SetLimit is used form errgroup and  this means an idle stream holds zero goroutines, where N workers implementation park N goroutines on idle

Issue #1385, opened 2026-05-03
Nil Pointer Panic in Generic Membership Service


Rozerxshashank:
The `Service` struct in `platform/fabric/core/generic/membership/membership.go` relies on `channelResources`, which is initialized asynchronously via an update. Public methods like `IsValid`, `GetVerifier`, `GetMSPIDs`, and `OrdererConfig` attempt to access `c.resources().MSPManager()` without verifying if `resources()` is `nil`.

Rozerxshashank:
Yes working

Rozerxshashank:
@mbrandenburger can you please manually assign me this issue. Thanks!

PR #1665, opened 2026-08-24
fix(ordering): report unset consensus type instead of a bespoke error

### LINKED ISSUE #1664: `ordering.Service` cannot report that it has no broadcaster yet
### Description

`ordering.Service` gets its broadcaster from the channel configuration, on the same asynchronous path as #1385: the committer feeds the config block to `MembershipService.Update`, reads it back with `OrdererConfig`, and passes the consensus type to `Service.Configure` → `SetConsensusType`. Until that lands, `Broadcast` fails with a bespoke string:

```go
// platform/fabric/core/generic/ordering/ordering.go:112
o.BroadcastMutex.RLock()
broadcaster := o.Broadcaster
o.BroadcastMutex.RUnlock()
if broadcaster == nil {
    return errors.Errorf("cannot broadcast yet, no consensus type set")
}
```

Two problems:

1. **The startup window is not detectable.** After #1663 a caller can recognise the same condition on the membership path with `errors.Is(err, driver.ErrNotInitialized)`. Here it can only string-match, so a node that is merely still coming up is indistinguishable from one that is permanently misconfigured — and retry logic cannot tell them apart.

2. **The locking invariant is unenforceable.** `Broadcaster` and `BroadcastMutex` are both exported. Nothing stops a caller inside or outside the package from reading or writing `Broadcaster` without the mutex; the existing unit tests already do exactly that. The guarded-field pattern is hand-rolled here for the third time in the codebase.

### Implementation Steps

Hold the broadcaster in the `configstate.Holder` introduced by #1663, and unexport it. `Get` then returns the `driver.ErrNotInitialized`-wrapped error, the mutex disappears, and the "read only under the lock" rule is enforced by the type rather than by convention.

### Additional Information

_No response_

mbrandenburger:
Closes #1664

### Description

`ordering.Service` gets its broadcaster from the channel configuration, on the same asynchronous path as #1385: the committer feeds the config block to `MembershipService.Update`, reads it back with `OrdererConfig`, and passes the consensus type to `Configure` → `SetConsensusType`. Until that lands, the service has no broadcaster, and `Broadcast` reported it like this:

```go
o.BroadcastMutex.RLock()
broadcaster := o.Broadcaster
o.BroadcastMutex.RUnlock()
if broadcaster == nil {
    return errors.Errorf("cannot broadcast yet, no consensus type set")
}
```

Two problems. The startup window is not detectable — after #1663 a caller recognises the same condition on the membership path with `errors.Is(err, driver.ErrNotInitialized)`, but here it can only string-match, so a node that is merely still coming up looks the same as one that is permanently misconfigured. And the locking invariant is unenforceable: `Broadcaster` and `BroadcastMutex` are both exported, so nothing requires the mutex to be taken — the unit tests assigned the field directly, and did.

### Changes

- **`platform/fabric/core/generic/ordering`**: the broadcaster moves into the `configstate.Holder` added by #1663 and is unexported along with its mutex. `Broadcast` reaches it through `Get`, so a service that cannot broadcast now says which of the two reasons applies:

```
awaited:  cannot broadcast: network [x] ordering configuration not loaded: not initialized
refused:  cannot broadcast: network [x] ordering configuration rejected: no broadcaster found for consensus [kafka]: configuration rejected
```

- `SetConsensusType` selects inside `Holder.Update`. Because `Update` leaves the held value alone when its function fails, an unsupported consensus type can no longer drop a working service back into the uninitialized state.
- `Broadcasters` stays exported: it is written only by `NewService` and read-only afterwards, so it is not the guarded-invariant problem.
- **`platform/fabric/driver`**: new `ErrConfigRejected`. `ErrNotInitialized` is documented as a transient startup condition its callers should retry, so it cannot also stand for a configuration that arrived and was refused — retrying that never helps. The two differ by what unblocks the caller rather than how long they last: `ErrNotInitialized` clears when the configuration arrives, `ErrConfigRejected` when a later update is accepted, and neither means the node is finished.
- **`platform/fabric/core/configstate`**: `NewHolder` took a channel name and hardcoded `"channel [%s] configuration not loaded"`. An ordering service does not hold a channel configuration, so the parameter becomes a `subject` noun phrase and the message becomes `"%s not loaded"`. The two membership call sites pass `"channel [x] configuration"`, leaving their error text unchanged. `Update` additionally records *why* an update was refused while nothing has been accepted, so `Get` can report `ErrConfigRejected` with that reason wrapped. Once a configuration is in force the holder keeps answering from it and a later refusal reaches the updater alone — already the behaviour, now documented.
- **Tests**: `newTestService()` collapses fifteen copies of the same eight-line constructor call, and `setBroadcaster` registers a fake under a test consensus type and calls `SetConsensusType`, so the tests go through the real selection path instead of writing the field. New cases cover broadcasting before a consensus type is set, a rejected `SetConsensusType` keeping the previous broadcaster, broadcasting after a rejected consensus type, and recovery from one. On the holder: that a rejection is not reported as a startup race, that its reason survives, that a later success clears it, that a refusal after a success does not mask the value in force, and that `TryGet` ignores rejections.

Membership accessors now report `ErrConfigRejected` where they reported `ErrNotInitialized` for a config block that failed to parse or validate. That narrows a sentinel added in #1663, which is not in any release tag yet. `Broadcaster` and `BroadcastMutex` had no references anywhere else in the repository.

### Verification

```
go build ./...                                                    ok
go test ./platform/...                                            ok
go test -race -count=2 (ordering, configstate, membership x2)     ok
golangci-lint run ./platform/fabric/... ./platform/fabricx/...    0 issues
```

### DIFF
--- a/platform/fabric/core/generic/ordering/ordering.go
+++ b/platform/fabric/core/generic/ordering/ordering.go
@@ -8,13 +8,13 @@ package ordering
 
 import (
 	"context"
-	"sync"
 
 	common2 "github.com/hyperledger/fabric-protos-go-apiv2/common"
 
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/proto"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/configstate"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/metrics"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/grpc"
@@ -49,15 +49,30 @@ type BroadcastFnc = func(ctx context.Context, env *common2.Envelope) error
 
 type GetEndorserTransactionServiceFunc = func(channelID string) (driver.EndorserTransactionService, error)
 
+// Service broadcasts envelopes to the ordering service of a network using the
+// broadcaster for the consensus type in force.
+//
+// Which consensus type that is comes from the channel configuration, which is
+// loaded asynchronously after the service is built: the committer feeds it to
+// MembershipService.Update, reads it back with OrdererConfig, and passes it here
+// via Configure. Until that happens the service has no broadcaster and Broadcast
+// reports driver.ErrNotInitialized.
 type Service struct {
 	GetEndorserTransactionService GetEndorserTransactionServiceFunc
 	SigService                    driver.SignerService
 	ConfigService                 driver.ConfigService
 	Metrics                       *metrics.Metrics
 
-	Broadcasters   map[ConsensusType]BroadcastFnc
-	BroadcastMutex sync.RWMutex
-	Broadcaster    BroadcastFnc
+	// Broadcasters holds one broadcaster per supported consensus type. It is
+	// populated by NewService and not written afterwards.
+	Broadcasters map[ConsensusType]BroadcastFnc
+
+	// broadcaster holds the broadcaster selected by SetConsensusType. It is
+	// unexported and reached only through configstate.Holder so that the
+	// "always read under the lock" rule cannot be sidestepped, and so that the
+	// not-yet-selected case reaches the caller as an error rather than as a nil
+	// function value.
+	broadcaster *configstate.Holder[BroadcastFnc]
 }
 
 func NewService(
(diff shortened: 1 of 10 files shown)


platform/fabric/core/generic/ordering/ordering.go
```diff
@@ -109,28 +123,28 @@ func (o *Service) Broadcast(ctx context.Context, blob any) error {
...
-	o.Broadcaster = broadcaster
-	return nil
+
+	return o.broadcaster.Update(func(BroadcastFnc, bool) (BroadcastFnc, error) {
+		broadcaster, ok := o.Broadcasters[consensusType]
+		if !ok {
+			return nil, errors.Errorf("no broadcaster found for consensus [%s]", consensusType)
+		}
+		return broadcaster, nil
+	})
```
SaidAltury-ibm, line 164: When the channel config carries an unsupported consensus type, this returns an error and Update correctly leaves the holder unloaded — but that discards the distinction between "not told yet" and "told something invalid". Downstream, Broadcast (127–130) can then only report ErrNotInitialized, whose contract at driver/errors.go:19 says it is "a transient startup condition, not a permanent failure" that callers "should […] retry". So a permanently misconfigured node instructs a conforming caller to retry forever — which is the exact confusion #1664 set out to remove. Could we record the rejected-config case on the Service (or add an ErrMisconfigured-style sentinel) so Broadcast can distinguish the two? WYT?

mbrandenburger: Oh this is a good catch. Let me work on this!

SaidAltury-ibm (approved):
Thank you @mbrandenburger for this great work!

PR #1666, opened 2026-08-25
fix(fabricx): bound finality handler concurrency with fixed worker pool (#1666)

### LINKED ISSUE #1627: fabricx: notification dispatcher spawns two unbounded goroutines per finality notification

SaidAltury-ibm:
Fixes the unbounded goroutine growth in #1627.

The notification dispatcher spawned two goroutines per `(txID, listener)` pair: an outer one to enforce `handlerTimeout`, an inner one to run `OnStatus`. The timeout bounded only the outer goroutine, so a listener that ignored cancellation left the inner one pinned for the life of the process — notification rate became a memory leak.

## What changed

Callbacks now run on a **fixed pool**. `listen()` starts `handlerWorkers` goroutines (default 16) draining a queue of `handlerQueueSize` (default 1000). Concurrency is bounded by construction: that many goroutines exist, so at most that many callbacks run. `handlerWorkers` caps concurrency; the queue absorbs bursts, since one committer response can name far more transactions than there are workers.

**The queue belongs to `listen()`, not to the manager.** Workers are `for c := range q` — no `select`, no context in the loop. `close(q)` ends the range only once the buffer is empty, so draining the backlog at shutdown is guaranteed rather than attempted, and there is no cancellable context in scope to hand a listener by mistake. A second `listen()` call gets its own queue, so no callback can outlive the stream that produced it.

**Hand-off is all-or-nothing per transaction.** `handOff` checks `cap(q)-len(q)` and either queues every listener waiting on a txID or none; callers delete the map entry only on success. This is what makes delivery exactly-once: the retry path knows the entry, not which of its listeners already ran, so a partial hand-off would settle those a second time on the next sweep. `cap-len` is a trustworthy lower bound because workers only consume, and the three producers all run on the dispatcher goroutine.

**Nothing the committer told us is lost.** A transaction whose callbacks could not be queued keeps its listeners *and* the status the committer sent, and the sweeper retries it with that status — not downgraded to `Unknown`. The remembered status is itself the "due" signal, so such an entry is retried whatever its deadline says, and the sweeper no longer returns early when `listenerTTL: 0` — that setting disables the local backstop, not the redelivery of a result we were given.

**Teardown has one deadline for the whole thing**, not one per listener: settle what is left into the queue, close it, wait once. With 300 pending listeners that all ignore their context, teardown takes 151ms and leaves 9 goroutines, against one goroutine per listener before.

Because a stuck listener costs throughput rather than memory, "`OnStatus` must observe `ctx.Done()` and return promptly" is now a requirement, and that warning sits on the `FinalityListener` interface.

## Configuration

Two new `notificationService` keys, both with defaults, both clamped to at least 1:

| Key | Default | Meaning |
|-----|---------|---------|
| `handlerWorkers` | 16 | how many `OnStatus` callbacks may run concurrently |
| `handlerQueueSize` | 1000 | how many pending invocations are buffered while every worker is busy |

Existing deployments need no change.

## Trade-offs worth knowing

- **The pending-listener map grows while the pool is congested.** That is the price of never dropping a result: with every worker wedged, 500 registrations retain 500 map entries. It is a strictly better failure mode than unbounded goroutine growth — no stacks, and the entries drain as soon as a worker frees — but it is not free, and `AddFinalityListener` has no admission control.
- **Blocked callbacks remain out of scope.** Nothing can force a callback to return, so enough stuck listeners still stop delivery on that stream, and listeners queued behind one are not settled at teardown. The pool bounds the blast radius; it cannot remove it.
- **Size `handlerQueueSize` above the most listeners you ever register for one txID.** Hand-off is per transaction, so an entry with more listeners than the whole queue can only be settled when the stream stops.
- An idle stream now parks `handlerWorkers` + 4 goroutines where the original held 4. Fixed and bounded, unlike what it replaces.

## Testing

`TestHandlerPool` and `TestHandOff` — 24 subtests covering the concurrency cap, bursts larger than the pool, the deferral path with and without `listenerTTL`, exactly-once delivery under a saturated queue, and every teardown path; plus config tests for the new keys and their zero/negative fallbacks.

Each invariant was verified red-green — breaking the hand-off atomicity, the sweeper's `listenerTTL: 0` path, the remembered status, the entry delete, the callback context, the teardown bound, or the workers' drain each fails its test. `TestHandOff` is a unit test on purpose: healthy workers drain about as fast as the dispatcher fills, so the integration-level test does not reliably reproduce a split hand-off.

Green under `-race -count=3` on `./platform/fabricx/core/finality/...` and `./platform/fabricx/core/committer/config/...`; `golangci-lint` clean.

Closes #1627


platform/fabricx/core/finality/nlm.go
```diff
@@ -191,11 +257,31 @@ func (n *notificationListenerManager) settleAllAndClear(ctx context.Context, sta
 
 	for _, p := range batch {
 		for _, h := range p.listeners {
-			n.invokeHandler(ctx, h, p.txID, status)
+			n.callHandlerBounded(ctx, handlerCall{handler: h, txID: p.txID, status: status})
```
mbrandenburger: I have the feeling we should also obey our parameter `handlerWorker` here during the cleanup instead of serializing the teardown. WDYT?

platform/fabricx/core/finality/nlm.go
```diff
@@ -89,6 +106,46 @@ func (n *notificationListenerManager) listen(ctx context.Context) error {
...
+	// would hand in-flight callbacks a dead context, and callHandler's timeout would
+	// expire instantly.
+	poolCtx, stopHandlers := context.WithCancel(context.WithoutCancel(ctx))
+	defer stopHandlers()
+
+	var handlerPool sync.WaitGroup
+	for range n.handlerWorkers {
+		handlerPool.Add(1)
+		go func() {
+			defer handlerPool.Done()
```
mbrandenburger: I suggest to use modern wg style here ...

```go
for range n.handlerWorkers {
    wg.Go(func() {
```

No need to manual call `.Add(1)` and `.Done()`

platform/fabricx/core/finality/nlm.go
```diff
@@ -89,6 +106,44 @@ func (n *notificationListenerManager) listen(ctx context.Context) error {
...
+			for {
+				select {
+				case c := <-n.callQueue:
+					n.callHandler(poolCtx, c)
+				case <-poolCtx.Done():
+					// Drain before exiting. dispatch already deleted these callbacks'
+					// handlers entries, so settleAllAndClear will not settle them:
+					// returning here loses the notification outright. Bounded by the
+					// queue's length at this instant.
+					drainCtx := context.WithoutCancel(poolCtx) // poolCtx is already done
```
mbrandenburger: I believe we have one special case here we need to cover. In the case where `poolCtx` is already done and `callQueue` still has items, the select case of all our worker will randomly pick on or the other cases. If `poolCtx.Done()` is picked, we are all good, however, if `n.CallQueue` is selected, we call `n.CallHandler(poolCtx, c)` with a closed context - and this is the problem.

What we could do is to create the `drainCtx` before the loop and use it in both branches and use `poolCtx` just for the select

platform/fabricx/core/finality/nlm.go
```diff
@@ -402,9 +553,13 @@ func (n *notificationListenerManager) dispatch(ctx context.Context, resp *commit
...
 
+	dropped := 0
 	for _, c := range calls {
-		n.invokeHandler(ctx, c.handler, c.txID, c.status)
+		if !n.enqueueHandler(c) {
+			dropped++
+		}
 	}
+	n.logDrops(dropped, len(calls))
 }
```
mbrandenburger: I am wondering if we are missing something here. Could it happen that the TTL sweeper does not catch handler if they are not enqueued here?

platform/fabricx/core/finality/nlm.go
```diff
@@ -459,9 +615,14 @@ func (n *notificationListenerManager) sweepExpired(ctx context.Context) {
 
 	logger.Debugf("Settling %d expired finality listener(s) with Unknown", len(batch))
 
+	dropped, total := 0, 0
 	for _, e := range batch {
 		for _, h := range e.listeners {
-			n.invokeHandler(ctx, h, e.txID, fdriver.Unknown)
+			total++
+			if !n.enqueueHandler(handlerCall{handler: h, txID: e.txID, status: fdriver.Unknown}) {
```
mbrandenburger: I think here we have a similar issue to the one mentioned in the `dispatch` method. Should we add the handler back to `n.handlers` in case we cannot `enqueue` the handler for cleanup?

platform/fabricx/core/finality/nlm_group_test.go
```diff
@@ -0,0 +1,485 @@
...
+		i := int(idx.Add(1)) - 1
+		if i < len(responses) {
+			return responses[i], nil
+		}
+		<-ctx.Done()
+		return nil, ctx.Err()
+	}
+}
+
+func TestHandlerGroup(t *testing.T) {
```
mbrandenburger: I suggest to rename the file into `nlm_handler_test.go` or merge into `nlm_test.go` but considering the size of this test already having a dedicated file is fine.

platform/fabricx/core/finality/nlm_group_test.go
```diff
@@ -0,0 +1,485 @@
...
+		require.EventuallyWithT(t, func(collect *assert.CollectT) {
+			assert.Equal(collect, int32(1), stuck.inFlight.Load())
+		}, timeout, tick, "the stuck callback should be in flight")
+
+		cancel()
+		select {
+		case <-listenErr:
+		case <-time.After(2 * time.Second):
+			t.Fatal("listen() did not return: teardown blocked on a callback that ignores its context")
+		}
```
mbrandenburger: should we also obey the `t.Context.Done()` in these tests.

platform/fabricx/core/finality/nlm_group_test.go
```diff
@@ -0,0 +1,485 @@
...
+
+	t.Run("Teardown_Drains_Queued_Callbacks", func(t *testing.T) {
+		// Callbacks already sitting in callQueue when the stream stops must still be
+		// delivered. dispatch has removed their handlers entries by then, so
+		// settleAllAndClear will not settle them: dropping them here loses the
+		// notification outright, with no Unknown fallback.
+		t.Parallel()
+
+		nlm, fakeStream := setupTest(t)
+		nlm.handlerTimeout = config.DefaultHandlerTimeout
```
mbrandenburger: I am wondering we we want to use a shorter timeout for the test here. Letting the unit test wait for 5s during a teardown test might be a bit long

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for working on this PR. Solid work! Please see my comments below.

platform/fabricx/core/finality/nlm.go
SaidAltury-ibm: i see what you are saying here but what if all of the workers are blocked?

mbrandenburger: I was thinking that we might want to re-insert the handler on failed enqueue back into `n.handlers`. This would allow `sweepExpired` to catch the handler and sweep it. However, you are also right what if all our workers are blocked when sweeping?!

mbrandenburger (approved):
Thank @SaidAltury-ibm for working on this. I've added a few corrections. Thank

PR #1673, opened 2026-08-25
fix(fabricx)!: synchronize provider base context and stop panicking

### LINKED ISSUE #1671: fabricx ledger and finality providers race on `baseCtx` and panic when uninitialized
### Description

`ledger.Provider` and `finality.Provider` each take their base context from the SDK during startup, through `Initialize`, and each guards it the same way:

```go
// platform/fabricx/core/ledger/provider.go:65
func (p *Provider) Initialize(ctx context.Context) {
    p.initOnce.Do(func() {
        p.baseCtx = ctx
        ...
    })
}

// platform/fabricx/core/ledger/provider.go:76
func (p *Provider) NewLedger(network, channel string) (driver.Ledger, error) {
    if p.baseCtx == nil {
        panic("programming error: Provider is not initialized. ...")
    }
    ...
}
```

`finality.Provider.NewManager` (`provider.go:109`) is identical.

**1. The `sync.Once` does not synchronize the readers.** `Once.Do(f)` orders the completion of `f` before the return of *`Do`* — that edge exists only for goroutines that call `Do`. `NewLedger`, `newLedger` (`provider.go:99`), `Context` (`provider.go:104`) and `NewManager` never call it, so their reads of `baseCtx` have no synchronization with the write inside it at all. The race detector confirms it:

```
WARNING: DATA RACE
Write at 0x00c0001d1ca0 by goroutine 19:
  ...(*Provider).Initialize.1() provider.go:67
  sync.(*Once).doSlow()
Previous read at 0x00c0001d1ca0 by goroutine 20:
  ...(*Provider).NewLedger() provider.go:76
```

Today's wiring happens to serialize them — `platform/fabricx/sdk/dig/sdk.go:102-103` calls both `Initialize`s before `p.SDK.Start(ctx)` — so this is not currently reachable in production. But that makes the guard incoherent either way: if the ordering is guaranteed, the `initOnce` and the nil check are both dead weight; if it is not, the read races. Nothing in the type enforces the ordering the panic is checking for.

**2. Panicking is the wrong report.** Both methods already return an error. Since #1663 the equivalent condition on the membership path is reported as `driver.ErrNotInitialized` and is detectable with `errors.Is`, so a caller racing startup can retry. Here it takes down the node.

### Implementation Steps

Hold the context in the `configstate.Holder` (#1663, extended in #1665). `Update` expresses first-write-wins, `Get` returns the `ErrNotInitialized`-wrapped error, and one lock covers both the write and every read.

**Breaking:** `ledger.Provider.Context` would return `(context.Context, error)`, and `NewLedger`/`NewManager` would return an error where they panicked.

### Additional Information

_No response_

mbrandenburger:
Closes #1671

## Description

The ledger and finality providers each take their base context from the SDK during startup, via Initialize, and each guarded it with `sync.Once` plus a nil check in the reader. Neither half worked.

Once.Do orders its write only against goroutines that also call Do, and none of the readers do, so every read of baseCtx raced the write; the race detector reports it. Today's wiring calls both Initialize methods before SDK.Start, so the race is not reachable in production, but nothing in the type enforces that ordering - which is precisely what the nil check was trying to check for.

And the nil check panicked, on methods that already return an error. Since `driver.ErrNotInitialized` and can be recognised with errors.Is, so a caller racing startup can retry rather than lose the node.

Both providers now hold the context in a configstate.Holder. Update expresses first-write-wins, Get returns the ErrNotInitialized-wrapped error, and one lock covers the write and every read.

**BREAKING CHANGE**: `ledger.Provider.Context` returns `(context.Context, error)`, and `ledger.Provider.NewLedger` and `finality.Provider.NewManager` report an error wrapping `driver.ErrNotInitialized` where they previously panicked.

### DIFF
--- a/platform/fabricx/core/finality/provider.go
+++ b/platform/fabricx/core/finality/provider.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/configstate"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabricx/core/committer/config"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services"
 )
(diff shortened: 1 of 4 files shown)


platform/fabricx/core/finality/provider.go
```diff
@@ -92,11 +102,26 @@ func (p *Provider) resolveConfig(network string) (config.Config, error) {
...
+// initialized while every listener goroutine ran on nil, which surfaces as a
+// panic inside gRPC with nothing pointing back here. The provider stays
+// uninitialized instead, so NewManager keeps reporting it.
+//
 // IMPORTANT: This method MUST be called once during service setup before calling NewManager.
 func (p *Provider) Initialize(ctx context.Context) {
-	p.initOnce.Do(func() {
-		p.baseCtx = ctx
+	if ctx == nil {
+		logger.Error("Provider.Initialize called with a nil context, ignoring it")
```
SaidAltury-ibm: Inconsistent log messages part 1

platform/fabricx/core/ledger/provider.go
```diff
@@ -62,27 +71,46 @@ func NewProvider(grpcClientProvider GRPCClientProvider, queryServiceProvider que
...
+//
+// A nil context is ignored: installing one would leave the provider looking
+// initialized while every ledger made its RPC calls on nil, which surfaces as a
+// panic inside gRPC with nothing pointing back here. The provider stays
+// uninitialized instead, so NewLedger and Context keep reporting it.
 func (p *Provider) Initialize(ctx context.Context) {
-	p.initOnce.Do(func() {
-		p.baseCtx = ctx
+	if ctx == nil {
+		logger.Error("Ledger Provider.Initialize called with a nil context, ignoring it")
```
SaidAltury-ibm, line 81: Inconsistent log message part 2

SaidAltury-ibm (approved):
Thank you @mbrandenburger i have added 2 comments to alight logs please look. i am approving this PR

PR #1674, opened 2026-08-25
refactor(msp): hold the default identity behind a guarded holder

### LINKED ISSUE #1672: `msp.service` cannot report that no identity loader has supplied a default yet
### Description

The default identity of a local MSP manager is not available when the service is built. It arrives while `loadLocalMSPs` walks the configured MSPs and an identity loader calls `SetDefaultIdentity`. Two fields track it:

```go
// platform/fabric/core/generic/msp/service.go:39
defaultIdentityMutex   sync.RWMutex
defaultIdentity        view.Identity
defaultSigningIdentity driver.SigningIdentity
```

**1. Nil conflates two different answers.** `DefaultIdentity()` returns a bare `view.Identity`, so `nil` means either "no loader has run yet" or "this network has no default identity". This is the same conflation #1385 fixed for `GetMSPIDs`, and `loadLocalMSPs` has to re-read the field under a read lock just to test it for nil and synthesize an error from it (`service.go:404`).

**2. The locking is hand-rolled and inconsistently applied.** Two fields that are only ever written together are guarded by a mutex each accessor has to remember to take. Separately, `defaultMSP` is declared among the `mspsMutex` fields (`service.go:50`) but guarded by `defaultIdentityMutex` (`service.go:122`, `:379`), and read outside both at `service.go:390`.

Note: this is **not** a live data race. `Load()` and `Refresh()` both hold `mspsMutex` across the whole of `loadLocalMSPs`, and the only other reader of `defaultMSP` is `SetDefaultIdentity`, which runs from an identity loader inside that same critical section. The second mutex is redundant rather than wrong.

### Implementation Steps

Hold both halves in one `configstate.Holder` (#1663), so they are installed as a unit and the lock is taken by the holder rather than by convention, and let `loadLocalMSPs` take its error from `Get`. Drop the redundant second mutex and group `defaultMSP` with the lock that actually covers it.

That leaves problem 1 in place at the API boundary. Fixing it means `DefaultIdentity()` and `DefaultSigningIdentity()` returning an error on `driver.LocalMembership` and `fabric.LocalMembership` — roughly forty call sites across the root and integration modules, including the sample views under `integration/fabric/atsa/views/`. Worth doing, but as its own change.

### Additional Information

_No response_

mbrandenburger:
Closes #1672

### Description

The default identity of a local MSP manager is not available when the service is built: it arrives while `loadLocalMSPs` walks the configured MSPs and an identity loader calls `SetDefaultIdentity`. Two fields tracked it, guarded by a hand-rolled RWMutex, with each accessor expected to remember the lock.

Both now live in a single `configstate.Holder`. The identity and the signing identity that goes with it are installed as one value, so a reader can no longer observe one without the other, and the lock is taken by the holder rather than by convention. `loadLocalMSPs` takes its "no default identity" error from the holder instead of re-reading the field to test it for nil, so it can now say *why* there is none — never offered, or offered and refused.

`defaultMSP` is held the same way. It is read by `SetDefaultIdentity`, which is on the exported `driver.Manager` interface, and identity loaders are a dependency-injection extension point (`group:"identity-loaders"`) — so a loader outside this repository can retain a `Manager` and call it from a goroutine owning none of our locks. Guarding it with `mspsMutex` is not an option, because the in-tree loaders are already inside that critical section when they call it and `sync.RWMutex` is not reentrant. A holder carries its own lock and is safe for any caller.

### Behaviour changes

Two configurations that used to start now fail at startup instead. Both turn a latent nil into an explicit error, but they are breaks:

- **An identity with no signing identity is refused.** `loadLocalMSPs` only ever checked `defaultIdentity`, never `defaultSigningIdentity`, so a loader supplying one without the other passed startup and handed `nil` to whoever called `DefaultSigningIdentity` — e.g. `platform/fabricx/core/channel/provider.go:84`. The pair is documented as meaningless apart; it is now validated as a pair.
- **An empty MSP identifier no longer nominates the default.** `defaultMSP` was `""` until `loadLocalMSPs` resolved it, and MSP identifiers are nowhere validated as non-empty, so `SetDefaultIdentity("", ...)` compared equal and installed itself. The holder distinguishes "not resolved yet" from a real identifier.

Otherwise unchanged: `DefaultIdentity` and `DefaultSigningIdentity` still return `nil` before a loader has supplied one. Telling that apart from a network with no default identity needs those accessors to return an error, which is a break across roughly forty call sites and is left for its own change.

### Also in this PR

- **`configstate.Holder` gains `Reset`.** `Refresh` rebuilds everything derived from the configured MSPs, and the default identity is derived from them, so it has to be cleared too — otherwise a reload that cannot produce one keeps serving the identity of an MSP it just dropped, and reports success. `Update` alone cannot express "start again". Mirrors `lazy.Holder.Reset`.
- **The `mspsMutex` comment now records `AddMSP`'s calling convention as a known hole rather than a rule.** `AddMSP` has the same shape as the problem fixed above — exported on `driver.Manager`, no lock of its own, writes four maps — and is the more dangerous of the two, since a concurrent map write is fatal rather than merely wrong. It predates this PR and wants its own fix.

### DIFF
--- a/platform/fabric/core/generic/msp/service.go
+++ b/platform/fabric/core/generic/msp/service.go
@@ -55,6 +55,26 @@ type service struct {
 	// it apart from a network with no default.
 	defaults *configstate.Holder[defaultIdentity]
 
+	// defaultMSP holds the identifier of the MSP whose identity becomes the
+	// network default. It is resolved by loadLocalMSPs, from the configured
+	// value or, when nothing is configured, from the first MSP in the list, so
+	// it is not the same as the configured value that Config.DefaultMSP
+	// reports.
+	//
+	// Held rather than guarded by mspsMutex because SetDefaultIdentity reads it
+	// and is reachable through the exported driver.Manager interface: identity
+	// loaders are a dependency-injection extension point, so a loader outside
+	// this repository can hold a Manager and call it from a goroutine that owns
+	// no lock of ours. A holder is safe for any caller, whereas taking
+	// mspsMutex there would deadlock the in-tree loaders, which are already
+	// inside that critical section.
+	//
+	// Get also distinguishes "not resolved yet" from a real identifier, which
+	// an empty string cannot: MSP identifiers are nowhere validated as
+	// non-empty, so before loadLocalMSPs has run an empty id would otherwise
+	// compare equal and install itself as the default.
+	defaultMSP *configstate.Holder[string]
+
 	signerService       driver.SignerService
 	binderService       driver.BinderService
 	deserializerManager driver.DeserializerManager
@@ -62,14 +82,12 @@ type service struct {
 	KVS                 KVS
 	config              driver.Config
 
-	// mspsMutex guards the fields below. defaultMSP is the one worth spelling
-	// out: it is written by loadLocalMSPs under the write lock and read by
-	// SetDefaultIdentity, which an identity loader calls from inside that same
-	// critical section, on the same goroutine. So that read takes no lock, and
+	// mspsMutex guards the fields below. They are written by loadLocalMSPs and
+	// by AddMSP, which an identity loader calls from inside that same critical
+	// section on the same goroutine, so AddMSP takes no lock of its own and
 	// must not start taking one — sync.RWMutex is not reentrant and it would
-	// deadlock. AddMSP already relies on this same calling convention.
+	// deadlock.
 	mspsMutex           sync.RWMutex
-	defaultMSP          string
 	identityLoaders     map[string]driver.IdentityLoader
 	msps                []*driver.MSP
 	mspsByName          map[string]*driver.MSP
@@ -90,6 +108,7 @@ func NewLocalMSPManager(
 ) *service {
 	s := &service{
 		defaults:            configstate.NewHolder[defaultIdentity]("default identity"),
+		defaultMSP:          configstate.NewHolder[string]("default MSP"),
 		config:              config,
 		KVS:                 KVS,
 		signerService:       signerService,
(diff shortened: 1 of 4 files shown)


platform/fabric/core/generic/msp/service.go
```diff
@@ -35,17 +36,38 @@ type KVS interface {
...
+	defaultViewIdentity view.Identity
+	KVS                 KVS
+	config              driver.Config
+
+	// mspsMutex guards the fields below. defaultMSP is the one worth spelling
+	// out: it is written by loadLocalMSPs under the write lock and read by
+	// SetDefaultIdentity, which an identity loader calls from inside that same
+	// critical section, on the same goroutine. So that read takes no lock, and
+	// must not start taking one — sync.RWMutex is not reentrant and it would
+	// deadlock. AddMSP already relies on this same calling convention.
```
SaidAltury-ibm: but SetDefaultIdentity is exported so i can be called from outside right?

SaidAltury-ibm (commented):
Thank you @mbrandenburger, please review my comment

platform/fabric/core/generic/msp/service.go
mbrandenburger: Addressed in the fixup. `defaultMSP` is now a `configstate.Holder[string]` instead of a field read without a lock, so `SetDefaultIdentity` is safe for any caller again — the property the old `defaultIdentityMutex` gave it, which I shouldn't have dropped. 

Putting a second mutex back wasn't an option: taking one in `SetDefaultIdentity` would deadlock the in-tree loaders, since they call it from inside `mspsMutex`. A holder carries its own lock, so it's safe regardless of what the caller holds, and the order stays `mspsMutex` → holder, never the reverse. 

It also closes a smaller hole in the same place: an unresolved `defaultMSP` was `""`, and MSP identifiers are nowhere validated as non-empty, so a caller passing an empty `id` before `loadLocalMSPs` had run compared equal and installed itself as the default — `Get` now reports that state rather than answering with a value. `TestSetDefaultIdentityIsSafeWithoutHoldingMspsMutex` pins the guarantee; I checked it does fail under `-race` against the unguarded version, so it isn't passing vacuously.

SaidAltury-ibm (approved):
LGTM

PR #1676, opened 2026-08-25
test(nwo): run the nwo unit tests in CI

### LINKED ISSUE #1675: test(ci): nwo unit tests are never executed
### Description

`make unit-tests` has never run a single test under `integration/nwo/`. `GO_PACKAGES` in the root `Makefile` is:

```make
GO_PACKAGES = $$(go list ./... | grep -Ev '/(integration/)'; go list ./integration/nwo/...)
```

The second `go list` runs from the root module and fails:

```
pattern ./integration/nwo/...: directory prefix integration/nwo does not contain main module or its selected dependencies
```

It writes that to stderr and contributes nothing to stdout, so `TEST_PKGS` quietly comes back holding only root-module packages and `make unit-tests` still reports success. **55 packages containing 23 test files never run in the `utest` job.**

Fixing the package list alone is not enough: `integration/` is a separate module, so `go test` invoked from the root module cannot build those packages either. nwo needs its own `go test -C integration` invocation.

### Implementation Steps

- [ ] Remove the broken `go list ./integration/nwo/...` leg from `GO_PACKAGES`
- [ ] Add a `NWO_PACKAGES = ./nwo/...` variable and a second `go test -C integration ...` invocation to the `unit-tests` target
- [ ] Make the recipe exit non-zero if either invocation fails, while still running both
- [ ] Leave `unit-tests-postgres` unchanged (no Postgres tests under nwo, and that job has no Fabric binaries)
- [ ] Confirm the nwo suite is green in CI

### Additional Information

Surfaced while reviewing #1650, which adds tests under `integration/nwo/fabric/network` that would otherwise never run in CI.

mbrandenburger:
Closes #1675 

`TEST_PKGS` appended `go list ./integration/nwo/...`, which fails from the root module ("directory prefix integration/nwo does not contain main module"), so none of the nwo unit tests ever reached `make unit-tests`. nwo lives in the integration module and needs its own `go test -C` invocation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,9 @@ postgres:
 # include the checks target
 include $(TOP)/checks.mk
 
-GO_PACKAGES = $$(go list ./... | grep -Ev '/(integration/)'; go list ./integration/nwo/...)
+# we use a multi-module repo structure here and therefore need to carefully collect packages for unit tests
+GO_PACKAGES = $$(go list ./...)
+NWO_PACKAGES = ./nwo/...
 GO_PACKAGES_SDK = $$(go list ./... | grep '/sdk/dig$$')
 GO_TEST_PARAMS ?= -race -cover
 TEST_PKGS ?= $(GO_PACKAGES)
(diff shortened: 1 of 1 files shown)


SaidAltury-ibm (approved):

PR #1678, opened 2026-08-25
ci(codeql): build for extraction instead of running the test suites

### LINKED ISSUE #1677: ci(codeql): stop running the unit test suites in the CodeQL build
### Description

The CodeQL workflow's manual build step runs `make unit-tests unit-tests-postgres`, re-running on every push exactly what the `utest` and `utest-postgres` jobs in `tests.yml` already run in parallel.

CodeQL builds its database from what it observes being **compiled**, so the test packages only need to be built, not executed. In the latest run, `Build manually` took **6m53s** of an **8m10s** job, against **53s** of actual analysis — roughly 85% of the job is duplicated test execution.

With nothing executing, the `make -j2 install-fabric-bins pull-images-database` setup step is no longer needed either.

### Implementation Steps

- [ ] Replace `make unit-tests unit-tests-postgres` with a compile-only build: `go build ./...` plus `go test -run '^$'` over the root module and `./nwo/...` (matches the set `make unit-tests` covers; `-run '^$'` builds every test binary and runs no test)
- [ ] Drop the fabric-binaries / database-images setup step
- [ ] Add a concurrency group so a new push to a PR cancels the superseded analysis
- [ ] After the first run, compare the lines-of-code and alert count reported by `Perform CodeQL Analysis` against a current run to confirm the database did not shrink

### Additional Information

- Concurrency groups for the remaining workflows (notably `tests.yml`, where the `itest`
  matrix makes a superseded push far more expensive) are intentionally left to a follow-up.

mbrandenburger:
Closes #1677

## Description

The CodeQL job ran `make unit-tests unit-tests-postgres` inside its manual build step, which re-ran on every push what the utest and utest-postgres jobs in tests.yml already run in parallel. CodeQL builds its database from what it observes being compiled, so the test suites only needed to be compiled, not executed: `Build manually` accounted for 6m53s of an 8m10s job, against 53s of actual analysis.

Compile the same set `make unit-tests` covers -- the root module plus nwo, which lives in the integration module -- with `go test -run '^$'`, which builds every test binary and runs no test. With nothing executing, the fabric binaries and database images the step used to install are no longer needed either.

Also add a concurrency group so a new push to a pull request cancels the analysis of the commit it replaced. Non-pull-request events fall back to run_id, which is unique per run, so pushes to main and the weekly scheduled run are never cancelled by, or queued behind, another run.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,11 +18,19 @@ on:
   schedule:
     - cron: '39 22 * * 5'
 
+# A newer push to the same PR makes the in-flight analysis obsolete, so cancel
+# it. Non-PR events (main pushes, the weekly schedule) fall back to run_id, which
+# is unique per run, so they each get their own group and are never cancelled by
+# -- or queued behind -- another run.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   analyze:
     name: Analyze
     runs-on: 'ubuntu-latest'
-    timeout-minutes: 60
+    timeout-minutes: 30
     permissions:
       security-events: write
       packages: read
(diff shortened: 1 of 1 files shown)


SaidAltury-ibm (approved):

PR #1680, opened 2026-08-26
ci: consolidate checks on golangci-lint and lint every module

### LINKED ISSUE #1679: ci: consolidate checks on golangci-lint and lint every module
### Description

`make checks` runs golangci-lint plus standalone `govet`, `ineffassign` and `staticcheck` — all three of which golangci-lint already enables — while `golangci-lint.yml` runs the same tool again over the whole tree. Only `licensecheck` is unique to it.

The resulting `checks` job gates `utest`, `utest-postgres` and the whole `itest` matrix, adding ~2m36s to the critical path of a ~13m run. It passed 20 of the last 20 runs.

Neither path lints anything outside the root module: `integration/` (333 Go files), the libp2p host and `cc/query` are unchecked today.

### Implementation Steps

- [ ] Drop the targets golangci-lint already covers; keep it as the single linter
- [ ] Replace `addlicense` with golangci-lint's `goheader`
- [ ] Run golangci-lint once per module, with the module list discovered not hardcoded
- [ ] Delete the `checks` job and the `needs: checks` gates

### Additional Information

_No response_

mbrandenburger:
Closes #1679 

### Description

Makes golangci-lint the single linter and runs it over every Go module.

- **`checks.mk`** — `checks` is now just `lint`. `govet`, `ineffassign`, `staticcheck` and `gocyclo` are dropped: golangci-lint already enables them. `lint`, `lint-auto-fix` and `lint-fmt` run once per module, collecting failures instead of stopping at the first. New `list-go-modules` target is the single source of truth for the module list.
- **`.golangci.yml`** — adds `goheader`, replacing `addlicense`. The template accepts the SPDX header and the long-form Apache notice; the copyright line is optional, a license marker is not. Generated `out/` trees are excluded.
- **`golangci-lint.yml`** — a discovery job feeds a per-module matrix, mirroring the existing `itest-list` → `itest` pattern.
- **`tests.yml`** — the `checks` job and its gates are gone; test jobs start immediately.
- **`tools/`** — `addlicense`, `staticcheck`, `ineffassign` and `gocyclo` removed.

`tools/` itself is not linted: its `//go:build tools` tag leaves no buildable package and
golangci-lint fails there with `no go files to analyze`.

### Testing

- Full lint (what CI runs) is clean in all four modules: root, `integration`, `cc/query`, libp2p host.
- `make checks` passes, and fails correctly when a license header is removed.
- `goheader` verified against a scratch module: missing header and wrong license are both caught; a correct header and a counterfeiter-generated file pass.
- Linting `integration/` for the first time surfaced 4 `thelper` misses, fixed here.


SaidAltury-ibm (approved):

PR #1682, opened 2026-08-26
ci(tests): cancel superseded pull request runs

### LINKED ISSUE #1681: ci(tests): cancel superseded pull request runs
### Description

`tests.yml` has no concurrency group, so pushing to a pull request leaves the previous run going to completion on a commit nobody will look at again. The workflow fans out to 19 simultaneous jobs (`checks`, both utest jobs, and a ~17-job `itest` matrix) over about 13 minutes, which is most of the standard 20-job allowance — so a superseded run delays every other pull request in the org.

Sampling recent runs, 15 of 25 pull-request runs were followed by another run on the same branch, each a cancellation candidate whenever the next push lands inside that 13-minute window.

`codeql-analysis.yml` already got this treatment in #1678; this applies the same pattern.

### Implementation Steps

- [ ] Add a workflow-level concurrency group keyed on the pull request number
- [ ] Fall back to `run_id` for pushes to `main` and manual dispatch, so those are never cancelled or queued
- [ ] Confirm Coveralls tolerates the parallel builds that cancelled runs leave unclosed

### Additional Information

_No response_

mbrandenburger:
Closes #1681 

### Description

Adds a workflow-level concurrency group to `tests.yml`, so a new push to a pull request cancels the run it supersedes. This is the follow-up to #1678, which applied the same pattern to `codeql-analysis.yml`.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
  cancel-in-progress: true
```

It matters more here than it did for CodeQL: this workflow starts 19 jobs at once and runs for ~13 minutes, so a superseded run occupies most of the concurrency allowance while producing nothing.

Pushes to `main` and `workflow_dispatch` runs key off `run_id`, which is unique per run, so they are never cancelled or queued behind each other.

### Testing

YAML parses and all six jobs are unchanged; cancellation behaviour can only be exercised by pushing.

### Note

Cancelling skips `publish-coverage`, so the Coveralls parallel build for the superseded commit is never closed. The next push opens its own build and `fail-on-error: false` means nothing fails, but it is worth a look at the Coveralls project after the first cancelled run.

The remaining workflows (the bot ones, `git.yml`, `gofix.yml`) still have no concurrency group. They are cheap, so they are left for a later pass.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,16 @@ on:
       - '.gitignore'
   workflow_dispatch:
 
+# A newer push to the same pull request makes the in-flight run obsolete, so
+# cancel it -- this workflow fans out to the whole itest matrix, so a superseded
+# run holds a large share of the concurrency allowance. Non-pull-request events
+# (pushes to main, manual dispatch) fall back to run_id, which is unique per run,
+# so they each get their own group and are never cancelled by -- or queued behind
+# -- another run.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   utest:
     runs-on: ubuntu-latest


SaidAltury-ibm (approved):
thank you @mbrandenburger , just keep an eye on Coveralls afterward.
LGTM

PR #1684, opened 2026-08-26
refactor(common): move the config holder to deferred, and generalize it

### LINKED ISSUE #1683: move the config holder to deferred, and generalize it
### Description

`configstate.Holder` was introduced to hold channel configuration for the two membership services, and its name, package doc and location all said so. It has since spread to seven call sites, and five of them hold something else.

### Implementation Steps

- [ ] Move Holder
- [ ] Update usage

### Additional Information

_No response_

mbrandenburger:
Closes #1683

### Description 

`configstate.Holder` was introduced to hold channel configuration for the two membership services, and its name, package doc and location all said so. It has since spread to seven call sites, and five of them hold something else: a broadcast function, a default identity, a default MSP identifier, and two base contexts. The package doc claiming it "holds channel configuration" was false for most of its users, and `platform/fabricx` reached into `platform/fabric/core` for what is really a concurrency helper.

What the type actually is: a value supplied after the owning service was constructed, which readers must not see early. It moves to `platform/common/utils/deferred`, beside `lazy.Holder`, which is the same shape pulled rather than pushed - lazy computes its value from a provider on first use, deferred is handed one from outside. Each now points at the other so the choice is visible from either side.

The two sentinels move with it, as `deferred.ErrNotLoaded` and `deferred.ErrRejected`, and `driver.ErrNotInitialized` and `driver.ErrConfigRejected` become aliases of them. That inverts the dependency - the package is now a leaf rather than importing `platform/fabric/driver` - while keeping the Fabric contract documented where `driver.ChannelMembership` already describes it. They are the same error values, so every existing errors.Is keeps matching.

No behaviour change: the sentinel messages are unchanged, so error text is byte-identical.


SaidAltury-ibm (approved):

PR #1694, opened 2026-08-30
build: exclude test-only helper suites from coverage denominator

### LINKED ISSUE #1693: build: exclude test-only helper suites from the coverage denominator
## Problem

`scripts/filter-coverage.sh` excludes mocks and fakes from the coverage report on the
principle that test scaffolding is not shipped code. It does not apply that principle to
shared test helpers.

Helpers that other packages import cannot live in a `_test.go` file, so conformance
suites, benchmark bodies and node harnesses sit in plain `.go` files — and are counted as
production code. Today that is **14 files, 3,498 relevant lines, ~1,134 of them "missed"**,
which distorts every package percentage in the report.

Two examples of how misleading this is:

- `platform/view/services/storage/driver/sql/common/bench.go` reports **0 %** simply
  because `go test` does not run `Benchmark*` bodies without `-bench`.
- `platform/view/services/comm/io/test_utils.go` reports **0 %** because its only caller
  sits outside the root module's test run.

Neither is a testing gap. Both are measurement artifacts.

## Proposal

Adopt one naming convention and filter on it.

1. Rename the four shared helpers that do not follow it (no import changes — Go imports
   reference packages, not filenames):

   | Current | New |
   |---|---|
   | `platform/common/core/generic/vault/helpers.go` | `vault_test_utils.go` |
   | `platform/fabric/services/db/driver/sql/common/helpers.go` | `kvstore_test_utils.go` |
   | `platform/view/services/storage/driver/sql/common/helpers.go` | `kvstore_test_utils.go` |
   | `platform/view/services/storage/driver/sql/common/bench.go` | `bench_test_utils.go` |

2. Add one rule to `filter-coverage.sh`, alongside the existing mock/fake exclusions:

   ```awk
   /[^\/]*test_utils[^\/]*\.go:/ { next }
   ```

3. Document the convention in `docs/agents/testing.md`.

Ten helpers already follow the `*test_utils*.go` naming and are picked up by the same rule
without changes.

## Why a naming convention rather than a pattern on `helpers.go`

A blanket `helpers.go` / `bench.go` pattern would be wrong. Both names are also used by
production code:

- `platform/fabric/services/state/helpers.go` — production, no `testing` import
- `platform/fabric/services/storage/vault/helpers.go` — production, 83.3 % covered

Excluding those would drop shipped code from the report, which is worse than the problem
being fixed. Renaming the test-only files makes the filter a single unambiguous pattern.

## Files deliberately NOT excluded

`platform/common/services/logging/logger.go` (`NewTestLogger`) and
`platform/common/utils/assert/retry.go` (`EventuallyWithRetry`) each mix production code
with a single test helper. Excluding them would remove real production code, and it changes
the result by only 0.02 pp. They stay; the documented guidance is to split such helpers into
a `*_test_utils.go` file instead.

## Effect on the reported number

Measured against Coveralls build 81440537:

| | Before | After |
|---|---:|---:|
| Coverage | 82.21 % | 83.59 % |
| Covered / relevant | 33,406 / 40,635 | 31,042 / 37,137 |
| Missed | 7,229 | 6,095 |

**This tests nothing.** The reported figure moves ~1.4 pp purely by removing non-shipped
code from the denominator, and no coverage credit should be claimed for it. Note also that
most of these files were already well covered (the vault suite at 100 %, the view suite at
98.2 %), so the change removes covered lines as well as missed ones — it stops measuring
scaffolding in both directions rather than deleting red lines.

The value is that package percentages become meaningful, which is a precondition for
planning any real testing work against them.

## Acceptance criteria

- [ ] Shared test helpers follow the `*_test_utils.go` convention.
- [ ] `filter-coverage.sh` excludes them by pattern.
- [ ] No live helper suite is deleted or made unreachable; `make unit-tests` still passes.
- [ ] Production files named `helpers.go` remain in the report.
- [ ] Convention documented in `docs/agents/testing.md`.
- [ ] No coverage-delta claim in the PR description beyond noting the denominator change.

SaidAltury-ibm:
filter-coverage.sh already excludes mocks and fakes on the principle that test scaffolding is not shipped code, but not shared test helpers. Helpers imported by other packages cannot live in _test.go files, so conformance suites, benchmark bodies and node harnesses sit in plain .go files and are counted as production code: 14 files, 3,498 relevant lines.

Adopt one naming convention and filter on it. Rename the four helpers that did not follow it and match *test_utils*.go in the awk filter; ten helpers already named that way are picked up by the same rule. Go imports reference packages rather than filenames, so the renames change no call sites.

A pattern on helpers.go/bench.go would be wrong: services/state/helpers.go and services/storage/vault/helpers.go are production code and must stay in the report.

logging/logger.go (NewTestLogger) and utils/assert/retry.go (EventuallyWithRetry) mix production code with a single test helper, so they are deliberately kept; excluding them would drop shipped code and changes the result by 0.02pp.

This changes measurement only. No test is added and no behaviour is tested; the reported figure moves purely by removing non-shipped code from the denominator.

Closes #1693


docs/agents/testing.md
```diff
@@ -52,3 +52,26 @@ never commit them (CI treats a stray focus as a failure). Prefer the
...
+**Name them `*_test_utils.go`.** That single convention is the whole mechanism:
+`scripts/filter-coverage.sh` matches the pattern and keeps them out of the coverage
+denominator, exactly as it already excludes mocks and fakes. Nothing enforces the
+name, and a misnamed helper is silently counted as production code — so apply the
+convention when you add one.
+
+A file counts as test-only when it imports `testing` and *every* exported function
+takes a `*testing.T/B/TB`. Do not rely on generic names for this —
+`platform/fabric/services/state/helpers.go` and
+`platform/fabric/services/storage/vault/helpers.go` are production code.
```
mbrandenburger: I have the feeling that `platform/fabric/services/storage/vault/helpers.go` is actually test helper code as all three `Open*Vault(...)` functions used only within tests, however, you are right, we do not import `testing` in that file. 

I suggest to rename `platform/fabric/services/storage/vault/helpers.go` -> `platform/fabric/services/storage/vault/vault_test_utils.go`

mbrandenburger: Moreover, there is another testing file in `platform/view/services/comm/host/websocket/testing.go` that does not follow the rule as specified here but is clearly only for testing.

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for this PR. Good work!

I think we have a few more places in the repo where we deal with test code which we may address in separate follow-up PRs.

For example:
- `platform/common/services/logging/logger.go` seems to have production and test code. This file can be split.
- `platform/fabric/core/msp/mgmt/testtools/config.go` provides more testing functions but not used at all. We could remove it - once checked that it is not used by Panurus.
- `platform/fabric/core/generic/ledger/kvledger/rwsetutil/` seems to be in a funky situation as `kvrwset_proto_util.go` is only used by it's tests. Probably dead code.

See also my comments below.

docs/agents/testing.md
SaidAltury-ibm: good catch, Thank you!

mbrandenburger (approved):

PR #1696, opened 2026-08-31
test(fabric): cover channel configuration updates

### LINKED ISSUE #1695: Fabric: No integration coverage for channel configuration changes
### Description

FSC reacts to a Fabric CONFIG block in `platform/fabric/core/generic/committer`: `HandleConfig` → `CommitConfig` commits it to the vault, hands the envelope to `MembershipService.Update`, then `applyConfigUpdates` re-reads the channel's orderer configuration and reconfigures the ordering service. On start-up, `ReloadConfigTransactions` (`generic/channel.go:60`) replays the stored `configtx_<n>` entries.

None of that is covered end to end. All nine Fabric suites were checked; none performs a channel configuration update. What runs today is incidental and unasserted: delivery seeks from `StartGenesis` (`delivery/delivery.go:34`), so every node replays block 0 — a config block — at start-up, and `stoprestart` re-runs the vault replay. A regression in membership refresh or orderer re-pointing would only surface if it broke the whole flow.

NWO also cannot currently produce a post-genesis config block on its own: `UpdateChannelAnchors` (`network_support.go:692`) has no callers, and anchor peers are baked into the genesis configtx instead (`topology/configtx_template.go:44-46`). So in every Fabric test the channel's config block set is `{genesis}` and never changes.

The machinery to close this already exists but has never had a caller — `fabric/network/configblock.go` and `update.go` arrived in the first code drop (6900cc85), vendored from Fabric's own nwo harness.

### Implementation Steps

- Create new integration test for config updates for the fabric platform

### Additional Information

_No response_

mbrandenburger:
Adds `integration/fabric/configupdate`, closing the gap in #1695

The suite runs the IOU flow, then changes the channel's orderer `BatchTimeout` (1s → 2s → 500ms) via orderer-signed config updates while the FSC nodes are running, re-running the flow after each change and again after restarting a node. Each transaction after a change is the assertion: it needs the nodes to still hold a usable view of the channel — its MSPs, to endorse and validate, and its ordering endpoints, to broadcast — which is exactly what `committer.HandleConfig` rebuilds. The restart additionally drives `ReloadConfigTransactions` over a sequence longer than the genesis block alone, for the first time in the test suite.

`BatchTimeout` is deliberate: it lives in the Orderer group that `applyConfigUpdates` re-reads, and changing it cannot partition the network or invalidate an identity, so a downstream failure points at FSC rather than the fixture.

The topology is IOU's, reusing its views and SDK verbatim, minus monitoring and tracing (nothing here asserts on them, and they cost container start-up).

**Side effect:** this gives the previously unused nwo config-block helpers their first caller — all of `network/update.go` (240 LOC) and 8 of the 12 functions in `network/configblock.go`. Still unreached: `UpdateConfig`, `UpdateOrdererConfigSession`, `UpdateConsensusMetadata`, `UpdateOrdererMSP`, and all of `network/orderer_client.go`.

**Known limitation:** this proves the nodes survive and keep transacting across a CONFIG block, not that they adopt *new* orderer endpoints — `applyConfigUpdates` returns early on an empty endpoint list (`committer/config.go:157-160`), so a silent no-op there would still pass. Catching that needs a multi-orderer Raft topology where the endpoint set actually changes; worth a follow-up.

### DIFF
--- a/integration/fabric/configupdate/configupdate_test.go
+++ b/integration/fabric/configupdate/configupdate_test.go
@@ -1,0 +1,94 @@
+/*
+Copyright IBM Corp All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package configupdate_test
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+
+	"github.com/hyperledger-labs/fabric-smart-client/integration"
+	"github.com/hyperledger-labs/fabric-smart-client/integration/fabric/configupdate"
+	"github.com/hyperledger-labs/fabric-smart-client/integration/fabric/iou"
+	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fsc"
+)
+
+var _ = Describe("EndToEnd", func() {
+	Describe("Channel Config Update With Websockets", Label("T1"), func() {
+		s := NewTestSuite(fsc.WebSocket, integration.NoReplication, true)
+		BeforeEach(s.Setup)
+		AfterEach(s.TearDown)
+		It("succeeded", s.TestSucceeded)
+	})
+})
+
+type TestSuite struct {
+	*integration.TestSuite
+}
+
+func NewTestSuite(commType fsc.P2PCommunicationType, nodeOpts *integration.ReplicationOptions, tlsEnabled bool) *TestSuite {
+	return &TestSuite{TestSuite: integration.NewTestSuite(func() (*integration.Infrastructure, error) {
+		return integration.Generate(StartPort(), integration.WithRaceDetection, configupdate.Topology(&configupdate.Opts{
+			CommType:        commType,
+			ReplicationOpts: nodeOpts,
+			TLSEnabled:      tlsEnabled,
+		})...)
+	})}
+}
+
+// TestSucceeded drives the IOU flow across two channel configuration changes and a node
+// restart. Each IOU transaction after a change is the assertion: it needs the FSC nodes to
+// still hold a usable view of the channel — its MSPs, for endorsement and validation, and
+// its ordering endpoints, for broadcast — which is precisely the state that
+// committer.HandleConfig rebuilds when a CONFIG block arrives.
+func (s *TestSuite) TestSucceeded() {
+	iou.InitApprover(s.II, "approver1")
+	iou.InitApprover(s.II, "approver2")
+
+	By("transacting on the channel's initial configuration")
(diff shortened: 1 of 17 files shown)


integration/fabric/configupdate/configupdate_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	By("restarting a node, so it replays the stored configuration transactions")
+	// On start-up a channel calls Committer.ReloadConfigTransactions, which walks the
+	// configtx_<n> entries in the node's vault and re-applies each one. With two updates
+	// committed above, this is the first time in the suite that the loop runs over a
+	// sequence longer than the genesis block alone.
+	s.II.StopFSCNode("borrower")
+	time.Sleep(3 * time.Second)
+	s.II.StartFSCNode("borrower")
+	time.Sleep(3 * time.Second)
```
SaidAltury-ibm, line 114: i think this can be flaky, WYT?

integration/fabric/configupdate/configupdate_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	iou.UpdateIOU(s.II, iouState, 3, "approver1")
+	iou.CheckState(s.II, "borrower", iouState, 3)
+	iou.CheckState(s.II, "lender", iouState, 3)
+
+	By("restarting a node, so it replays the stored configuration transactions")
+	// On start-up a channel calls Committer.ReloadConfigTransactions, which walks the
+	// configtx_<n> entries in the node's vault and re-applies each one. With two updates
+	// committed above, this is the first time in the suite that the loop runs over a
+	// sequence longer than the genesis block alone.
```
SaidAltury-ibm: can we assert that the node's vault should hold configtx_0, configtx_1 and configtx_2, and ReloadConfigTransactions should walk all three?

SaidAltury-ibm (commented):
Thank you @mbrandenburger for this PR. Great work!
i have left 2 comments, tell me WYT?

integration/fabric/configupdate/configupdate_test.go
mbrandenburger: ack, will add a view that can be called at each node to check the config sequence number.

integration/fabric/configupdate/configupdate_test.go
mbrandenburger: Agreed, will address this in a separate PR.

SaidAltury-ibm (approved):

integration/fabric/configupdate/configupdate_test.go
mbrandenburger: PR #1698 resolves this issue. We can rebase this PR and remove the sleep.

PR #1698, opened 2026-09-01
test(nwo): wait for an FSC node to exit instead of sleeping

mbrandenburger:
While working on #1696 we discussed that we should not use sleeps in integration tests to wait until a FSC node shutdown as completed as this might be flaky. This PR addresses this issue.

StopFSCNode sent SIGTERM and returned, so a caller that restarted the node raced the dying process for its ports and its vault. Callers papered over this with time.Sleep(3s) on either side of the restart: one to hope the node had gone, one to hope its replacement had come up. The second was already redundant -- StartFSCNode waits for the node to log its start check -- and the first was a guess.

Wait for the process to be gone, the way every other stop path in nwo already does with StopEventuallyTimeout, and drop both sleeps from the stoprestart suites.

Waiting means reading Runner.exitCode from the caller's goroutine while the runner's monitor writes it, which the race detector flags, so make it atomic. Two latent bugs fall out of pinning that value down: a cloned Runner never initialised it, so a restarted node's runner reported a clean exit before it had run, and a process whose OS exit status was -1 stored that as its exit code -- indistinguishable from "still running", which would have hung the new wait until it timed out.

### DIFF
--- a/integration/nwo/common/runner/runner.go
+++ b/integration/nwo/common/runner/runner.go
@@ -191,16 +191,17 @@ func (r *Runner) Run(sigChan <-chan os.Signal, ready chan<- struct{}) error {
 }
 
 func (r *Runner) monitorForExit(exited chan<- struct{}) {
-	err := r.Command.Wait()
+	if err := r.Command.Wait(); err != nil {
+		logger.Debugf("process [%s] exited with error: %s", r.Name, err)
+	}
 	status := r.Command.ProcessState.Sys().(syscall.WaitStatus)
 	if status.Signaled() {
 		r.exitCode.Store(int32(128 + int(status.Signal())))
 	} else {
 		exitStatus := status.ExitStatus()
-		// An exit status of -1 is how the OS reports "no status", and it is the
-		// same value this type uses for "still running": storing it would leave
-		// a dead process looking alive forever to anyone waiting on it.
-		if exitStatus == notExited && err != nil {
+		// Wait has returned, so the process is gone. ExitStatus reports -1 for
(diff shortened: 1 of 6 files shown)


integration/nwo/common/runner/runner.go
```diff
@@ -181,13 +194,16 @@ func (r *Runner) monitorForExit(exited chan<- struct{}) {
...
+		r.exitCode.Store(int32(128 + int(status.Signal())))
 	} else {
 		exitStatus := status.ExitStatus()
-		if exitStatus == -1 && err != nil {
-			r.exitCode = gexec.INVALID_EXIT_CODE
+		// An exit status of -1 is how the OS reports "no status", and it is the
+		// same value this type uses for "still running": storing it would leave
+		// a dead process looking alive forever to anyone waiting on it.
+		if exitStatus == notExited && err != nil {
+			exitStatus = gexec.INVALID_EXIT_CODE
```
SaidAltury-ibm: if ExitStatus() returns -1 while Wait() returned nil, the -1 is still stored, and the exact failure the comment describes still happens. The comment says the hazard is closed; the code closes it only when err != nil.

SaidAltury-ibm (commented):
thank ypu @mbrandenburger , i have left a comment please take a look and tell em WYT?

integration/nwo/common/runner/runner.go
mbrandenburger: good catch - let me fix that

SaidAltury-ibm (approved):

PR #1700, opened 2026-09-01
fix(collections): correct the iterator bugs the coverage tests pinned

### LINKED ISSUE #1699: test: increase coverage for collections/iterators
## Problem

`platform/common/utils/collections/iterators` is at **24% coverage**, the lowest of any
substantial package in the repo. Six of its files have no tests at all: `flatten.go`,
`reducers.go`, `filter.go`, `stream.go`, `predicates.go` and `map.go`.

The package is generic infrastructure used by around 20 others — `driver.TxStatusIterator`,
the `VaultRead` iterators, and the pagination types all build on it — so a defect here
propagates quietly into vault reads and paging.

## Proposal

Add behavioural unit tests for the package: success paths, error paths, and the edge cases
around empty iterators, exhaustion, and transformer/consumer failures.

Scope:

- `flatten.go` — `Flatten`, `FlattenValues` and both `Next()` implementations
- `utils.go` — `ReadAllPointers`, `ReadAllValues`, `ReadFirst`, `Copy`, `GetUnique`,
  `GetFirst`, `ForEach`, `Reduce`, `ReduceValue`
- `reducers.go` — `NewReducer`, `ToSet`, `ToFlattened`, `ToMaxBy`
- `filter.go`, `map.go`, `slice.go`, `stream.go`, `predicates.go`, `empty.go`

Conventions: `package iterators_test`, `testify/require`, table-driven cases where they are
homogeneous, and `tb testing.TB` + `tb.Helper()` for fixture builders, per
`docs/agents/testing.md`. The package takes no `context`, so `t.Context()` does not apply
here; where a test does need one elsewhere, use `t.Context()` rather than
`context.Background()`.

## Acceptance criteria

- [ ] Package coverage above 95%.
- [ ] Both `Flatten` implementations covered, including the empty-inner-group and
      transformer-error paths.
- [ ] Error paths asserted for every helper that can fail.
- [ ] `go test -race` clean; `make checks` passes.

SaidAltury-ibm:
collections/iterators was at 24.0% despite being the substrate under 21 packages, including the vault driver's TxStatusIterator and VaultRead iterators and the pagination types. Six of its files had no coverage at all: flatten.go, reducers.go, filter.go, stream.go, predicates.go and map.go.

Add behavioural tests for the whole package: 99.4% of statements.

Writing them surfaced three bugs, fixed here rather than pinned as contracts:

- `ReadFirst` advanced the iterator before re-checking the limit, so it read `limit+1` elements and dropped that read's error — over a SQL-backed iterator, a swallowed query failure.
- `ToMaxBy` compared keys against a zero-valued running max, so an iterator whose keys all sat below the key type's zero reduced to the zero value instead of its maximum.
- `CheckJaegerTraces` flattens with `FlattenValues`, which reads a chunk carrying no spans as exhaustion — one such chunk truncated the asserted span set.

Two of the new tests could not fail, and now discriminate:
`TestCopyIsIndependent` built its two readers from separate sources, so it would have passed even if `Copy` returned its input; `TestReadFirstStopsEarly` planted its error exactly where the over-read discarded it, and now asserts the read count. `failingIterator` gained a call counter and a sticky failure.

Godoc on the touched combinators now states what callers must honour rather than how exhaustion is signalled internally, records that `Map` and `Flatten` close their source, and moves `Map`'s sentinel caveat out of a test comment.

closes #1699

### DIFF
--- a/platform/common/utils/collections/iterators/flatten_test.go
+++ b/platform/common/utils/collections/iterators/flatten_test.go
@@ -1,0 +1,229 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package iterators_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/iterators"
+)
+
+// flattenedPointers and flattenedValues are near-duplicate implementations that
+// can drift apart, so the cases that apply to both are declared once here and run
+// against each variant.
+//
+// The two differ only in what they yield: Flatten passes the element type through
+// (so a []*int group yields *int), while FlattenValues yields a pointer to each
+// element (a []int group yields *int). Both are driven below over groups of int,
+// which is the shape that lets one table cover them.
+type flattenVariant struct {
+	name string
+	// flattenInts builds an iterator over the given groups, yielding one *int per
+	// element, whatever the underlying representation.
+	flattenInts func(groups [][]int) iterators.Iterator[*int]
+	// flattenFailing does the same over a source that fails at failAt.
+	flattenFailing func(groups [][]int, failAt int) iterators.Iterator[*int]
+	// flattenBadTransformer returns an iterator whose transformer always fails.
+	flattenBadTransformer func(groups [][]int, err error) iterators.Iterator[*int]
+}
+
+var flattenVariants = []flattenVariant{
+	{
+		name: "FlattenValues",
+		flattenInts: func(groups [][]int) iterators.Iterator[*int] {
+			return iterators.FlattenValues(iterators.Slice(groups),
+				func(g []int) ([]int, error) { return g, nil })
+		},
+		flattenFailing: func(groups [][]int, failAt int) iterators.Iterator[*int] {
+			return iterators.FlattenValues(newFailing(groups, failAt),
+				func(g []int) ([]int, error) { return g, nil })
+		},
+		flattenBadTransformer: func(groups [][]int, err error) iterators.Iterator[*int] {
+			return iterators.FlattenValues(iterators.Slice(groups),
+				func([]int) ([]int, error) { return nil, err })
+		},
+	},
+	{
+		name: "Flatten",
+		flattenInts: func(groups [][]int) iterators.Iterator[*int] {
+			return iterators.Flatten(iterators.Slice(toPointerGroups(groups)),
+				func(g []*int) ([]*int, error) { return g, nil })
+		},
+		flattenFailing: func(groups [][]int, failAt int) iterators.Iterator[*int] {
+			return iterators.Flatten(newFailing(toPointerGroups(groups), failAt),
+				func(g []*int) ([]*int, error) { return g, nil })
+		},
+		flattenBadTransformer: func(groups [][]int, err error) iterators.Iterator[*int] {
+			return iterators.Flatten(iterators.Slice(toPointerGroups(groups)),
+				func([]*int) ([]*int, error) { return nil, err })
+		},
+	},
+}
+
+func toPointerGroups(groups [][]int) [][]*int {
+	out := make([][]*int, 0, len(groups))
+	for _, g := range groups {
+		out = append(out, ptrs(g...))
+	}
+	return out
+}
+
+// drainPointers reads a flattened iterator to exhaustion. Exhaustion is signalled
+// by a nil item, matching what ReadAllPointers expects.
+func drainPointers(tb testing.TB, it iterators.Iterator[*int]) []int {
+	tb.Helper()
+	out := make([]int, 0)
+	for {
+		item, err := it.Next()
+		require.NoError(tb, err)
+		if item == nil {
+			return out
+		}
+		out = append(out, *item)
+	}
+}
+
+func TestFlattenBothVariants(t *testing.T) {
+	t.Parallel()
+
+	for _, v := range flattenVariants {
+		t.Run(v.name, func(t *testing.T) {
+			t.Parallel()
+
+			t.Run("several groups", func(t *testing.T) {
+				t.Parallel()
+				it := v.flattenInts([][]int{{1, 2}, {3}, {4, 5, 6}})
+				require.Equal(t, []int{1, 2, 3, 4, 5, 6}, drainPointers(t, it))
+			})
+
+			t.Run("single group", func(t *testing.T) {
+				t.Parallel()
+				require.Equal(t, []int{42}, drainPointers(t, v.flattenInts([][]int{{42}})))
+			})
+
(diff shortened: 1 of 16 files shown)


platform/common/utils/collections/iterators/flatten_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+	transformErr := errors.New("cannot transform")
+	it := iterators.FlattenValues(
+		iterators.Slice([][]int{{1}}),
+		func([]int) ([]int, error) { return nil, transformErr },
+	)
+
+	_, err := it.Next()
+	require.ErrorIs(t, err, transformErr)
+	require.Contains(t, err.Error(), "failed transforming")
```
mbrandenburger: We should use `require.ErrorContains` for those checks. Please check all new tests.

platform/common/utils/collections/iterators/flatten_test.go
```diff
@@ -0,0 +1,242 @@
...
+// expect {1, 2, 3}.
+func TestFlattenStopsAtEmptyInnerGroup(t *testing.T) {
+	t.Parallel()
+
+	groups := [][]int{{1, 2}, {}, {3}}
+	it := iterators.FlattenValues(iterators.Slice(groups), func(g []int) ([]int, error) { return g, nil })
+
+	require.Equal(t, []int{1, 2}, drainPointers(t, it),
+		"an empty inner group currently ends iteration; the trailing group is not reached")
+}
```
mbrandenburger: Probably the godoc of `FlattenValues` should be updated to clearly tell the user about this behaviour.

platform/common/utils/collections/iterators/combinators_test.go
```diff
@@ -0,0 +1,482 @@
...
+
+// TestEmptyHasNext reaches HasNext, which the empty iterator exposes on the
+// concrete type rather than through the Iterator interface.
+func TestEmptyHasNext(t *testing.T) {
+	t.Parallel()
+
+	it, ok := iterators.Empty[int]().(interface{ HasNext() bool })
+	require.True(t, ok, "the empty iterator is expected to expose HasNext")
+	require.False(t, it.HasNext())
+}
```
mbrandenburger: It seems that this test gives us coverage for actual dead code. Can you please double check if there is a caller in production that has uses `empty.HasNext()` (checking panurus as well) - if that is the case we can remove dead code. WDYT?

platform/common/utils/collections/iterators/reducers_test.go
```diff
@@ -0,0 +1,158 @@
...
+		iterators.ToMaxBy(func(r *record) (int, error) { return r.size, nil }),
+	)
+	require.NoError(t, err)
+	require.NotNil(t, max)
+	require.Equal(t, "largest", max.name)
+}
+
+// TestToMaxByFirstElementWins covers the strict `<` comparison: a later element
+// equal to the current max does not replace it.
+func TestToMaxByTiePrefersFirst(t *testing.T) {
```
mbrandenburger: godoc mismatches actual test name

platform/common/utils/collections/iterators/utils_test.go
```diff
@@ -0,0 +1,335 @@
...
+	t.Parallel()
+
+	got, err := iterators.ReadFirst[int](newFailing(ptrs(t, 1, 2, 3), 1), 3)
+	require.ErrorIs(t, err, errAt)
+	require.Nil(t, got)
+}
+
+// TestReadFirstDoesNotOverReadStopsEarly documents that ReadFirst stops pulling
+// once the limit is reached, rather than draining the source.
+func TestReadFirstStopsEarly(t *testing.T) {
```
mbrandenburger: godoc mismatches actual test name

platform/common/utils/collections/iterators/combinators_test.go
```diff
@@ -0,0 +1,482 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/iterators"
+)
+
+// --- Filter ---------------------------------------------------------------
```
mbrandenburger: I think we should follow the pattern to have one testfile for each go file as we are already introducing other testfiles here. WDYT?

platform/common/utils/collections/iterators/reducers_test.go
```diff
@@ -0,0 +1,158 @@
...
+	type record struct {
+		name string
+		size int
+	}
+	items := ptrs(t,
+		record{name: "small", size: 1},
+		record{name: "largest", size: 99},
+		record{name: "middle", size: 50},
+	)
+	max, err := iterators.Reduce(
```
mbrandenburger: the var name `max` shadows a go built-in function. Better name.

platform/common/utils/collections/iterators/flatten_test.go
```diff
@@ -0,0 +1,242 @@
...
+		func(g []*int) ([]*int, error) { return g, nil },
+	)
+
+	_, err := it.Next()
+	require.ErrorIs(t, err, errAt)
+}
+
+// TestFlattenPointersStopsAtEmptyInnerGroup is the pointer-variant counterpart of
+// TestFlattenStopsAtEmptyInnerGroup. flattenedPointers and flattenedValues are
+// near-duplicate implementations, so both need the same case to catch drift.
```
mbrandenburger: as you say in the comment, can we structure this test file a bit better to ensure that run all cases for both flatten values and flatten pointers (where applicable?).

platform/common/utils/collections/iterators/helpers_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/iterators"
+)
+
+// errAt is returned by the failing test iterators below, so assertions can match
+// on identity rather than on a message.
+var errAt = errors.New("iterator failure")
+
+// ptrs turns values into the pointer slice most of the package's helpers expect.
+func ptrs[T any](tb testing.TB, vs ...T) []*T {
+	tb.Helper()
```
mbrandenburger: it seems that we do not need `tb.Helper()` here at all as this function "cannot" fail.

mbrandenburger:
```suggestion
func ptrs[T any](vs ...T) []*T {
```

mbrandenburger: In addition we could add another helper

```go
func ptr[T any](v T) *T { return &v }
```

Example:
`v := ptrs(t, 42)[0]` -> `v :=ptr(42)`
`ptrs(t, []int{1, 2}, []int{3}, []int{4, 5})` -> `ptrs([]int{1, 2}, []int{3}, []int{4, 5})`

WDYT?

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for working on this. Great to see the coverage increase. Please see my comments below.

mbrandenburger (approved):

PR #1702, opened 2026-09-01
fix(state): reject state:"hash" on fields that cannot be hashed

### LINKED ISSUE #1701: test: increase coverage for fabric/services/state
## Problem

`platform/fabric/services/state` is at **82% coverage**, with 299 uncovered lines. The
integration suites drive the happy paths end-to-end, so what is left uncovered is mostly
error branches and boundary conditions that only a unit test can reach.

The largest gaps are in the state-assembly core:

| File | Missed |
|---|---:|
| `namespace.go` | 69 |
| `sbe.go` | 48 |
| `certification.go` | 40 |
| `vault/vault.go` | 39 |
| `recipients.go` | 60 |

## Proposal

Extend the existing tests in this package to cover the error paths, boundary conditions and
state transitions.

Scope:

- `namespace.go` — `AddOutput`, `AddInputByLinearID`, `GetInputAt`, `GetOutputAt`, `Delete`,
  `AddCommand`, `Commands`, `Inputs`, `Outputs`, `NumInputs`, `NumOutputs`
- `sbe.go`, `certification.go`, `vault/vault.go`

Cases to add: index out of range (negative, at count, past count), unmarshalling into a
mismatched state type, an underlying RWSet write failure, a missing or duplicate linear ID,
deleting an absent key, and interleaved add/delete/read sequences.

`recipients.go` is deliberately left out: its four `View.Call` implementations need session
plumbing, and whether this repo unit-tests views at all should be settled separately.

Conventions per `docs/agents/testing.md`: extend the existing `_test.go` files rather than
adding parallel ones, `testify/require`, table-driven cases where they are homogeneous,
`tb testing.TB` + `tb.Helper()` for fixture builders, and **`t.Context()` where a context is
needed — not `context.Background()`**.

## Acceptance criteria

- [ ] `namespace.go` above 90%.
- [ ] Package above 88%.
- [ ] Every index-bounds and type-mismatch branch covered, asserting the specific error.
- [ ] `go test -race` clean; `make checks` passes.

SaidAltury-ibm:
`fabric/services/state` was at 80.8%. The integration suites drive the happy paths end to end, so what was left uncovered was mostly error branches and boundary conditions that only a unit test can reach. Writing those tests turned up a bug, which this PR also fixes.

## The bug

`state:"hash"` on a field the tag handler cannot process failed unsafely in two different ways:

- a non-byte slice (`[]string`, `[]int`) panicked inside `reflect.Value.Bytes`;
- any other unhandled kind fell through the switch silently, so the value was written to the ledger **in the clear** while the caller believed it was hidden.

Both now return an error at write time. The check is whether a `[]byte` is assignable to the field, which is the precondition for both reflect calls in play (`Bytes` needs element kind uint8, `Set` needs assignability, and only assignability implies both). An element-kind check looks equivalent but still admits `[]MyByte`, which reads fine and then panics in `Set`.

`[]byte` and named types whose underlying type is `[]byte` remain supported. The only `state:"hash"` uses in this repo are `[]byte`, so nothing here changes behaviour.

### Compatibility

The read side is deliberately **not** a mirror of the write side. `unmarshalTags` runs on every read path, so refusing outright would make states already committed by an older release — value in the clear, no preimage recorded — permanently unreadable, with no migration path. It refuses only when a preimage is present, which means the field was hashed and cannot be verified here. New occurrences are prevented by the write-side check.

Slices are refused unconditionally: those panicked before anything was written, so no ledger can hold one.

Downstream code with `state:"hash"` on a non-`[]byte` field will now get an error from `AddOutput` where it previously got silent, unhidden data.

## Coverage

| | before | after |
|---|---|---|
| package | 80.8% | **86.9%** |
| `namespace.go` | 79.3% | 92.9% |
| `sbe.go` | 65.2% | 93.4% |
| `tags.go` | — | 94.2% |

New: `namespace_errors_test.go`, `tags_errors_test.go`, `sbe_test.go`, `trustedread_test.go`. All reuse the existing `common_test.go` fakes; no new test infrastructure.

## Also in here

- Godoc for the exported options in `opts.go`, which had none.
- A `TODO` recording that certified inputs and hash hiding do not compose: `GetInputAt` serves a certified input without consulting the field mapping, so a hash-hidden state read that way never recovers its preimage and the unmarshal fails. Pre-existing, and relevant to the ATSA hash-hiding work.
- Two subtests removed from `namespace_test.go` that the new files subsume.
- `TestNamespaceInterleavedAddDeleteRead` no longer asserts that re-writing a key appends another write entry. That holds for `testRWSet`, which appends on every `SetState`, but not for `vault.WriteSet`, which de-duplicates by key — so the test was pinning fake-only behaviour.

## Testing

`go test -race -cover ./platform/fabric/services/state/` passes at 86.9%.
`golangci-lint run ./platform/fabric/services/state/...` reports 0 issues. Root and `integration` modules both build.

Closes #1701

### DIFF
--- a/platform/fabric/services/state/namespace_errors_test.go
+++ b/platform/fabric/services/state/namespace_errors_test.go
@@ -1,0 +1,541 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package state
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	cdriver "github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
+)
+
+const errNS = "assetns"
+
+// seedOutput writes a state through the namespace so the RWSet has a write entry
+// to read back, and returns the id it was stored under.
+func seedOutput(tb testing.TB, tx *Transaction, h *House) string {
+	tb.Helper()
+	require.NoError(tb, tx.AddOutput(h))
+	return h.LinearID
+}
+
+// requirePanicsContaining asserts f panics and that the panic value mentions want.
+//
+// It matches a substring rather than the whole value on purpose: several of these
+// messages carry a long-standing "filed"/"failed" typo, and an exact match would
+// turn a spelling fix into a test failure. Prefer the injected cause as want, since
+// that is the part a behaviour change would actually drop.
+func requirePanicsContaining(tb testing.TB, want string, f func()) {
+	tb.Helper()
+	defer func() {
+		r := recover()
+		require.NotNil(tb, r, "expected a panic")
+		require.Contains(tb, fmt.Sprint(r), want)
+	}()
+	f()
+}
+
+// seedRead registers a read entry for key with the given raw value.
+func seedRead(tb testing.TB, rwset *testRWSet, key string, raw []byte) {
+	tb.Helper()
+	require.NoError(tb, rwset.SetState(errNS, cdriver.PKey(key), raw))
+	rwset.writes[errNS] = nil // keep the write list clean; we only want a read
+	require.NoError(tb, rwset.AddReadAt(errNS, key, nil))
+}
+
+// TestNamespaceIndexOutOfRange checks that an index the RWSet cannot serve is
+// reported rather than panicking.
+//
+// The three indices below all take the same `i < 0 || i >= len(...)` branch in the
+// RWSet, so this is one path, not three. They are kept as a table because the
+// interesting property is that GetInputAt and GetOutputAt both surface the failure
+// for a negative index, one past the end, and far past the end -- a regression that
+// silently clamped instead of erroring would still be caught.
+func TestNamespaceIndexOutOfRange(t *testing.T) {
+	t.Parallel()
+
+	for _, index := range []int{-1, 1, 99} {
+		t.Run(fmt.Sprintf("get output at %d", index), func(t *testing.T) {
+			t.Parallel()
+			tx, _, _ := newTestStateTransaction(errNS)
+			seedOutput(t, tx, &House{Address: "one", LinearID: "id-1"})
+
+			err := tx.GetOutputAt(index, &House{})
+			require.Error(t, err)
+			require.ErrorContains(t, err, "failed getting state")
+		})
+
+		t.Run(fmt.Sprintf("get input at %d", index), func(t *testing.T) {
+			t.Parallel()
+			tx, rwset, _ := newTestStateTransaction(errNS)
+			raw, err := json.Marshal(&House{Address: "one", LinearID: "id-1"})
+			require.NoError(t, err)
+			seedRead(t, rwset, "id-1", raw)
+
+			err = tx.GetInputAt(index, &House{})
+			require.Error(t, err)
+			require.ErrorContains(t, err, "failed getting state")
+		})
+	}
+}
+
+// TestNamespaceGetOutputAtLastValidIndex is the boundary companion to the
+// out-of-range cases: count-1 must succeed.
+func TestNamespaceGetOutputAtLastValidIndex(t *testing.T) {
+	t.Parallel()
+
+	tx, _, _ := newTestStateTransaction(errNS)
+	seedOutput(t, tx, &House{Address: "first", LinearID: "id-1"})
+	seedOutput(t, tx, &House{Address: "second", LinearID: "id-2"})
+	require.Equal(t, 2, tx.NumOutputs())
+
+	var got House
+	require.NoError(t, tx.GetOutputAt(1, &got))
+	require.Equal(t, "second", got.Address)
+}
+
+func TestNamespaceGetOutputAtUnmarshalMismatch(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(errNS)
+	require.NoError(t, rwset.SetState(errNS, "id-1", []byte("{not-json")))
+
+	err := tx.GetOutputAt(0, &House{})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "failed unmarshalling state")
+}
+
+func TestNamespaceGetOutputAtMappingError(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(errNS)
+	seedOutput(t, tx, &House{Address: "one", LinearID: "id-1"})
+	rwset.getStateMetadataErr = errors.New("metadata failed")
+
+	err := tx.GetOutputAt(0, &House{})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "failed getting mapping")
+}
+
+func TestNamespaceGetInputAtUnmarshalMismatch(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(errNS)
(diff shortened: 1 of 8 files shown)


platform/fabric/services/state/tags_errors_test.go
```diff
@@ -0,0 +1,241 @@
...
+	type intTagged struct {
+		Count int `state:"hash"`
+	}
+
+	n := &Namespace{}
+	dest, mapping, err := n.marshalTags(nil, &intTagged{Count: 7})
+	require.NoError(t, err)
+	require.Empty(t, mapping, "no preimage is recorded for an unhandled kind")
+	require.Equal(t, 7, dest.(*intTagged).Count, "the value passes through unchanged")
+}
```
mbrandenburger: I believe here we found a bug in the marshalTag implementation. I believe it should return an error we we cannot handle the type but set `state:"hash"`. Thus, we should reject unsupported kinds.

platform/fabric/services/state/tags_errors_test.go
```diff
@@ -0,0 +1,241 @@
...
+	require.ErrorContains(t, err, "not supported for string field")
+
+	err = n.unmarshalTags(nil, &stringTagged{Secret: "s"}, map[string][]byte{})
+	require.Error(t, err)
+	require.ErrorContains(t, err, "not supported for string field")
+}
+
+// TestMarshalTagsIgnoresUnhashableKinds records that a state:"hash" tag on a kind
+// the implementation does not handle (here an int) is silently skipped rather than
+// rejected: only string is an explicit error, and only slices are hashed.
```
mbrandenburger: > and only slices are hashed.

Could it be that marshalTags panics if we provide a non-byte slice, for instance `[]string`?

platform/fabric/services/state/namespace_errors_test.go
```diff
@@ -0,0 +1,529 @@
...
+
+	var got House
+	require.NoError(t, tx.AddInputByLinearID("id-1", &got))
+	require.Equal(t, "hidden", got.Address)
+	require.Equal(t, uint64(7), got.Valuation)
+}
+
+// TestNamespaceTagUnmarshalError covers the unmarshalTags failure branch, reached
+// by corrupting the tag mapping that a hash-hidden field depends on.
+func TestNamespaceTagUnmarshalError(t *testing.T) {
```
mbrandenburger: It seems that this test does not reach unmarshal and therefore redundant with `TestGetFieldMappingCorruptPayload`. Can you please check.

platform/fabric/services/state/namespace_errors_test.go
```diff
@@ -0,0 +1,529 @@
...
+func TestNamespaceAddOutputEmbeddingStateDerivesInnerID(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(errNS)
+	inner := &House{Address: "wrapped", LinearID: "inner-id"}
+	require.NoError(t, tx.AddOutput(&embeddingHouse{inner: inner}))
+
+	stored, err := rwset.GetState(errNS, "inner-id")
+	require.NoError(t, err)
+	require.NotEmpty(t, stored, "the output is keyed by the embedded state's linear id")
```
mbrandenburger: I am wondering if we have to check more concrete here what to expect instead of checking if it is not empty.

platform/fabric/services/state/namespace_errors_test.go
```diff
@@ -0,0 +1,529 @@
...
+func TestNamespaceAddOutputHashHidingStoresHashAndPreimage(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(errNS)
+	h := &House{Address: "secret", LinearID: "id-1"}
+	require.NoError(t, tx.AddOutput(h, WithHashHiding()))
+
+	stored, err := rwset.GetState(errNS, "id-1")
+	require.NoError(t, err)
+	require.NotContains(t, string(stored), "secret", "the raw state must not be written in the clear")
```
mbrandenburger, line 276: should we also check `require.NotEmpty(t, stored)` to avoid confusion with missing key?

platform/fabric/services/state/trustedread_test.go
```diff
@@ -0,0 +1,78 @@
...
+	err := (&TrustedReadCertifier{}).VerifyInputCertificationAt(tx.Namespace, 0, "id-1")
+	require.Error(t, err)
+	require.ErrorContains(t, err, "failed getting rw set")
+}
+
+func TestTrustedReadCertifierVerifyReadError(t *testing.T) {
+	t.Parallel()
+
+	tx, rwset, _ := newTestStateTransaction(trustedNS)
+	rwset.getReadAtErr = errors.New("read failed")
```
mbrandenburger: how is this injected error used here?

platform/fabric/services/state/namespace_errors_test.go
```diff
@@ -0,0 +1,529 @@
...
+
+		t.Run("get input at "+tc.name, func(t *testing.T) {
+			t.Parallel()
+			tx, rwset, _ := newTestStateTransaction(errNS)
+			raw, err := json.Marshal(&House{Address: "one", LinearID: "id-1"})
+			require.NoError(t, err)
+			seedRead(t, rwset, "id-1", raw)
+
+			err = tx.GetInputAt(tc.index, &House{})
+			require.Error(t, err)
```
mbrandenburger: does this really cover different paths?

platform/fabric/services/state/tags_errors_test.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+const tagNS = "tagns"
+
+func TestGetFieldMappingInvalidKey(t *testing.T) {
+	t.Parallel()
+
+	tx, _, _ := newTestStateTransaction(tagNS)
+
+	// An empty key cannot be turned into a composite mapping key.
```
mbrandenburger: I think the comment is not accurate. Can you please check if the input is a NUL byte and an empty key is provided that it. actually succeeds?!

platform/fabric/services/state/namespace_errors_test.go
```diff
@@ -0,0 +1,529 @@
...
+		commands := tx.Commands()
+		require.Equal(t, 2, commands.Count())
+		require.Equal(t, "create", commands.At(0).Name)
+		require.Equal(t, "transfer", commands.At(1).Name)
+	})
+}
+
+// TestNamespaceAddCommandRejectsCorruptHeader checks a malformed existing header
+// surfaces as an error rather than being silently replaced.
+func TestNamespaceAddCommandRejectsCorruptHeader(t *testing.T) {
```
mbrandenburger: Can you check if that test is a duplicate and already available in namespace_test.go?

platform/fabric/services/state/trustedread_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	err := (&TrustedReadCertifier{}).VerifyInputCertificationAt(tx.Namespace, 0, "id-1")
+	require.Error(t, err)
+	require.ErrorContains(t, err, "failed reading committed state")
+}
+
+// TestTrustedReadCertifierVerifyNoCommittedValue covers the guard against an empty
+// committed value, which would otherwise be accepted as a trivially valid read.
+func TestTrustedReadCertifierVerifyNoCommittedValue(t *testing.T) {
```
mbrandenburger: Can you check if that test is a duplicate already available in `certification_test.go`?

mbrandenburger (commented):
Thank you @SaidAltury-ibm for working on this. Great coverage improvement. Please see my comments below.

mbrandenburger (approved):

PR #1709, opened 2026-09-02
docs(agents): expand the Godoc conventions

mbrandenburger:
The Godoc section only said that exported identifiers need a comment. Spell out what a good comment looks like: caller-perspective wording, which non-obvious behavior is worth documenting, what to leave out, and the style rules (complete sentences, FSC terminology, Godoc links).


adecaro (approved):
LGTM

Issue #1479, opened 2026-05-23
[storage] Migrate query builder from view storage driver to fabric vault driver


sid200727:
### Description

As part of the query builder migration discussed in the FSC community, the SQL query builder currently lives in platform/view/services/storage/driver/sql/query/.
This issue tracks copying it into platform/fabric/services/db/driver/sql/query/ for the fabric vault driver and eventually removing it from the view storage driver entirely.
Related: hyperledger-labs/fabric-token-sdk#1734

### Implementation Steps

- [x] Copy query builder from platform/view/services/storage/driver/sql/query/ into platform/fabric/services/db/driver/sql/query/
- [x] Redirect vault.go imports to use local copy
- [ ] Remove query builder from platform/view/services/storage/driver/sql/query/ entirely

mbrandenburger:
@sid200727 can you elaborate why we should move the query builder into platform/fabric/services/db/driver/sql/query/? I was thinking that we want to remove it entirely from FSC.

sid200727:
Hi @mbrandenburger! The reasoning came from a discussion with Angelo, the fabric vault driver (platform/fabric/services/db/driver/sql/common/vault.go) currently depends on the query builder from the view package. Removing the query builder from FSC entirely without first giving the fabric vault driver its own copy would break that dependency.
The two-step approach he suggested was:

1. Copy the query builder into the fabric driver (this PR) so vault.go has a local copy
2. Then remove it from the view package entirely in a follow-up PR

This way each step is independently reviewable and nothing breaks between steps. 
Does that make sense, or would you like me to have a different approach?

sid200727:
This issue is still in progress.

mbrandenburger:
Completed with #1711

PR #1478, opened 2026-05-23
feat(storage): copy query builder into fabric vault driver and redirect imports

sid200727:
Closes #1479

Part of the query builder migration (2 of 3 PRs).
Following the same approach as fabric-token-sdk PR #1734, this PR copies the SQL query builder from platform/view/services/storage/driver/sql/query/ into platform/fabric/services/db/driver/sql/query/ for the fabric vault driver.
Changes:

-> Copied query builder (common, cond, select, insert, update, delete, pagination) into fabric vault driver
-> Redirected vault.go imports to use the local copy
-> Added test interpreters for each subpackage
-> Full build passes, all tests pass

Next PR: Remove the query builder from platform/view/services/storage/driver/sql/query/ entirely once this is merged.

PR #1258, opened 2026-04-21
fix(storage): migrate AuditInfoStore to squirrel PlaceholderFormat

sid200727:
Migrates AuditInfoStore from the deprecated CondInterpreter query builder 
to squirrel's PlaceholderFormat, following the same pattern established 
for SignerInfoStore and BindingStore (PR #1199) and SimpleKeyDataStore (PR #1235).

Changes:
- common/auditinfo.go: replace ci CondInterpreter with sq.PlaceholderFormat
- sqlite/auditinfo.go: pass sq.Question to NewAuditInfoStore, add RetryWriteDB
- postgres/auditinfo.go: pass sq.Dollar to NewAuditInfoStore
- auditinfo_test.go: update mock to use sq.Dollar and exact query matching

All tests pass locally:
- go test ./platform/view/services/storage/driver/sql/common/... 
- go test ./platform/view/services/storage/driver/sql/sqlite/... 

Part of #1159

PR #1711, opened 2026-09-03
refactor(storage): replace squirrel with an internal SQL builder

### LINKED ISSUE #1710: Remove `github.com/Masterminds/squirrel` dependency and the SQL query builder that wraps it
### How we got here

FSC has had two goes at the SQL-building problem, and this issue closes out both.

1. **An in-house query builder** landed in [#907](https://github.com/hyperledger-labs/fabric-smart-client/pull/907) ("SQL Query builder", commit `331877f0`, 2025-05-13), originally at `platform/view/services/db/driver/sql/query/` and moved to `platform/view/services/storage/driver/sql/query/` by the platform reorganisation in [#976](https://github.com/hyperledger-labs/fabric-smart-client/pull/976). It grew a `select`/`insert`/`update`/`delete` statement API, a condition tree (`cond`), and keyset pagination ([#977](https://github.com/hyperledger-labs/fabric-smart-client/pull/977)).

   Being our own builder, every gap became our ticket: [#988](https://github.com/hyperledger-labs/fabric-smart-client/issues/988) (`LEFT JOIN` unsupported) and [#1122](https://github.com/hyperledger-labs/fabric-smart-client/issues/1122) (`EXISTS`, `IS NULL`, `IS NOT NULL`).

2. **Replacing it with an external dependency** was then agreed in [#1159](https://github.com/hyperledger-labs/fabric-smart-client/issues/1159) ("Deprecate query builder"), off the back of [this discussion on #1122](https://github.com/hyperledger-labs/fabric-smart-client/issues/1122#issuecomment-3908220052). The reasoning was sound — lower our maintenance burden, give applications a more flexible query surface. `github.com/Masterminds/squirrel` v1.5.4 was adopted and the stores were migrated one at a time:

   | Store | Issue | PR |
   |---|---|---|
   | SignerInfoStore / BindingStore | [#1269](https://github.com/hyperledger-labs/fabric-smart-client/issues/1269) | [#1199](https://github.com/hyperledger-labs/fabric-smart-client/pull/1199) — commit `cfbdca2e`, first to add squirrel to `go.mod` |
   | SimpleKeyDataStore / KeyValueStore | [#1270](https://github.com/hyperledger-labs/fabric-smart-client/issues/1270) | [#1235](https://github.com/hyperledger-labs/fabric-smart-client/pull/1235) |
   | AuditInfoStore | [#1271](https://github.com/hyperledger-labs/fabric-smart-client/issues/1271) | [#1258](https://github.com/hyperledger-labs/fabric-smart-client/pull/1258) |
   | VaultStore (SELECTs) | [#1272](https://github.com/hyperledger-labs/fabric-smart-client/issues/1272) | [#1301](https://github.com/hyperledger-labs/fabric-smart-client/pull/1301) |

### Why that migration should be undone rather than finished

The migration stalled halfway, and from where it stopped the end state is worse than either starting point:

1. **The replacement dependency is not actively maintained.** squirrel's last release is v1.5.4 (2023-03-17). We pin it in the **root** module, so it ships to every FSC consumer, and a CVE or Go-version incompatibility in it would be ours to work around, not upstream's. The move was made to reduce maintenance exposure; it added some.

2. **The in-house builder never actually left.** It was not replaced by squirrel — it was rewritten *on top of* squirrel. We still own the `sql/query` tree (~1,850 non-test LOC: a builder, a condition tree, five statement packages, and a pair of per-dialect *interpreters* that translate FSC pagination and condition types into squirrel's). So we carry both the dependency and the wrapper, and #1159's goal — stop maintaining a query builder — was not reached.

3. **Almost none of squirrel is used.** Every statement FSC builds is static: fixed columns, fixed `WHERE` shapes, no dialect branching. What squirrel contributed in practice was placeholder formatting. Meanwhile the interpreter indirection means you cannot read the emitted SQL off the call site, which is a real cost when debugging a storage query.

The conclusion is that FSC never needed a query *builder*; it needed a statement *writer*. #1159's goal is kept — we stop maintaining a general builder — but its chosen means (adopt a third-party builder) is what this task reverses.

### Intended strategy

Replace squirrel with a minimal internal writer at `internal/storage/sqlbuild` (~275 non-test LOC) that renders statements directly and numbers `$N` placeholders as it writes them, then delete the `sql/query` tree and the per-dialect interpreters outright. `internal/` is deliberate: this is a private implementation detail, not API we want to support for third parties, and putting it there prevents a third attempt at a public builder.

No replacement library. Evaluating alternatives (bob, goqu, bun, jet) is explicitly out of scope: a third-party builder cannot replace the thing FSC was actually using it for, and each one re-creates problem 1.

### Constraints and tradeoffs

- **The emitted SQL must not change.** This is a dependency swap, not a behaviour change. Every affected query needs its current SQL pinned by a characterization test *before* the port, so the result is provably byte-identical — including squirrel's edge cases (`IN ()` rendering as `(1=0)`, an empty `AND` as `(1=1)`, single-element `AND` parenthesisation, `[]byte` treated as a scalar rather than a list).
- **Parameter ordering is the main risk.** Hand-written SQL means a hand-written argument slice, and a placeholder that drifts from its argument is a silent data-corruption bug, not a compile error. The builder must make the two inseparable by construction rather than relying on review.
- **Placeholder unification is a breaking API change.** Both dialects accept `$N` (`modernc.org/sqlite` maps `$N` to positional ordinals, and `common/binding.go` already hand-built `$N` for sqlite in production), so the `ph sq.PlaceholderFormat` parameter disappears from nine exported constructors. Needs a release note and its own root-module tag.
- **`LIMIT`/`OFFSET` stay literals**, matching what production squirrel emitted. The deleted interpreter bound them as parameters, but that path was test-only; preserve production behaviour and document the divergence.
- **Deleting the tree strands two things worth keeping** — the `pagination` package and the `FieldName` type — so lift both out before deleting rather than dropping them.

### Files involved

- **Add:** `internal/storage/sqlbuild/`
- **Remove:** `platform/view/services/storage/driver/sql/query/` in full; the per-dialect `conditions.go` and `pagination.go` in `sql/sqlite/` and `sql/postgres/`
- **Lift out before deleting:** `sql/query/pagination/` → `platform/view/services/storage/driver/pagination/`, and `sql/query/common.FieldName` → package `driver` (`driver.FieldName`)
- **Port:** the vault store (`platform/fabric/services/db/driver/sql/common/`) plus the key-value, identity/binding, audit-info, signer-info and simple-key-data stores under `platform/view/services/storage/driver/sql/common/`

### Implementation Steps

- [ ] Pin the exact SQL of every affected statement with `sqlmock` + `QueryMatcherEqual` characterization tests, against today's squirrel output — bugs included.
- [ ] Add `internal/storage/sqlbuild`: a statement builder whose placeholder counter and argument slice advance together, so a `$N` and its value are written in one operation and cannot drift.
- [ ] Add lazily-rendered `WHERE` conditions (`Eq`, `Gt`, `Gte`, `Lt`, `In`, `And`) reproducing squirrel's edge-case rendering exactly.
- [ ] Add pagination rendering (`ORDER BY` / `LIMIT` / `OFFSET`) with `LIMIT` and `OFFSET` as literals, matching production.
- [ ] Port each store to `sqlbuild`, one store per commit, with the characterization tests unchanged as the oracle.
- [ ] Route `driver.Pagination` through the new builder; delete the squirrel and interpreter renderers, keeping the runtime half of `pagination/` (`none`/`offset`/`keyset`/`page`).
- [ ] Lift `pagination/` and `FieldName` up one level, then delete the `sql/query` builder packages and the per-dialect condition and pagination interpreters.
- [ ] Drop the dependency: `./scripts/gomate.sh tidy` across all modules; confirm no `Masterminds/squirrel` reference remains in any `.go`, `go.mod` or `go.sum`.
- [ ] Verify: `make checks`, `make lint`, `make unit-tests`, `make unit-tests-postgres`, and the fabric-x integration suites.
- [ ] Write the release note for the removed exported API, the dropped `ph sq.PlaceholderFormat` parameters, and the new `pagination` / `FieldName` import paths.

### Additional Information

**Dependency:** `github.com/Masterminds/squirrel` v1.5.4, last released 2023-03-17.

**Supersedes [#1479](https://github.com/hyperledger-labs/fabric-smart-client/issues/1479) (open) and draft PR [#1478](https://github.com/hyperledger-labs/fabric-smart-client/pull/1478).** That issue plans to *copy* the query builder into `platform/fabric/services/db/driver/sql/query/` for the fabric vault driver and then remove it from the view storage driver. Its final step ("Remove query builder from `platform/view/services/storage/driver/sql/query/` entirely") is what this task does — but by deleting the tree rather than relocating it, which makes the copy in steps 1–2 unnecessary. Suggest closing or re-scoping #1479 and closing #1478 — though that is the call of the people who opened them. #1479 cites `fabric-token-sdk#1734` as related; that downstream need no longer applies (see below).

**Downstream impact — none, verified.** The downstream consumer is [Panurus](https://github.com/LFDT-Panurus/panurus) (formerly Token SDK), and it needs no source change:

- Panurus already carries its own fork of the query builder at `token/services/storage/db/sql/query` (2,126 non-test LOC, already extended beyond FSC's version with `common/expression.go` and `cond/exists.go`), and that fork is **squirrel-free** — Panurus de-squirreled it independently.
- It has zero imports of FSC's `storage/driver/sql/query` tree and zero call sites for any of the nine constructors whose signature changes.
- `go mod why` shows squirrel reaches Panurus *only* through FSC's `storage/driver/sql/common`, so this change removes squirrel from Panurus's dependency graph entirely. Its eight `// indirect` requirements go stale and `go mod tidy` drops them.

The fork is therefore permanent and accepted: we get zero cross-repo coordination in exchange for a duplicate that has already diverged.

**Release/tagging:** FSC tags each module separately, so the root-module API removal needs its own tag to be consumable downstream.

mbrandenburger:
Closes #1710

Supersedes #1479 and draft PR #1478 — see *Supersedes* below.

## What

Removes `github.com/Masterminds/squirrel` from FSC and replaces it with `internal/storage/sqlbuild`, a 275-line writer that renders SQL directly. The in-house query package that existed to bridge FSC types to squirrel types (`platform/view/services/storage/driver/sql/query/`, ~1,850 non-test LOC) goes with it, along with the per-dialect condition and pagination interpreters. What that tree still carried once squirrel left — the `pagination` package and the `FieldName` type — is lifted up one level into `platform/view/services/storage/driver` rather than dropped.

Net: **−1,347 non-test lines**, −1,055 test lines, one fewer dependency in the root module.

## Why

This finishes what #1159 set out to do, by the opposite route.

#1159 decided to stop maintaining an in-house query builder by adopting an external one, and squirrel v1.5.4 was migrated in across #1199, #1235, #1258 and #1301. From where that migration stopped, the end state is worse than either starting point:

- **squirrel is unmaintained** (last release v1.5.4, 2023-03-17) and we pin it in the *root* module, so it ships to every FSC consumer. The move was made to reduce maintenance exposure; it added some.
- **The in-house builder never left.** It was rewritten *on top of* squirrel, so we carried both the dependency and the wrapper — and #1159's actual goal went unmet.
- **Almost none of squirrel was used.** Every statement FSC builds is static — fixed columns, fixed `WHERE` shapes, no dialect branching — so what squirrel contributed in practice was placeholder formatting, while the interpreter indirection made the emitted SQL impossible to read off the call site.

FSC never needed a query *builder*; it needed a statement *writer*. `internal/` is deliberate — this is a private implementation detail, not API we want to support for third parties, and it prevents a third attempt at a public builder.

## How

The port is behaviour-preserving by construction, not by inspection:

1. **Characterization tests first.** Before touching any query, the exact SQL of every affected statement was pinned with `sqlmock` + `QueryMatcherEqual` against today's squirrel output — bugs included (32 statements: 19 queries, 13 execs). Those oracles were then left untouched for the whole migration; the only change to them across the port is the import and constructor line.
2. **A builder that cannot mis-number a placeholder.** `sqlbuild.Builder` holds the `$N` counter and the argument slice together, and writing a placeholder appends its value in the same operation. Conditions are lazy (`WriteTo(*Builder)`), so a placeholder's number is assigned at the instant it is written. The classic hand-written-SQL bug — argument order drifting from placeholder order — is not expressible in this API.
3. **Squirrel's edge cases reproduced deliberately**, re-derived from its v1.5.4 `expr.go`: `IN ()` renders `(1=0)`, an empty `And` renders `(1=1)`, a single-element `And` keeps its parentheses, and `[]byte` is a scalar rather than a list.
4. **One store per commit** (squashed here), each green against the untouched oracle, then the dependency dropped with `./scripts/gomate.sh tidy` across all modules.

## Notable behaviour notes

- **Placeholders are now `$N` for both dialects.** sqlite already received `$N` in production (`common/binding.go` hand-built them before this change), and `modernc.org/sqlite` maps `$N`/`?N` names to positional ordinals, so this unifies what already shipped rather than introducing new dialect behaviour.
- **`LIMIT`/`OFFSET` render as literals**, matching what production squirrel emitted. The deleted query package's interpreter bound them as parameters; that path was test-only. Documented on `WritePaging`.
- **`OffsetFromRaw` now re-validates**, the way `KeysetFromRaw` already did. A deserialized cursor is untrusted input, and a negative page size is the same value as `NoLimit` — it would have rendered *no* `LIMIT` clause, turning a page request into a full-table read. (On `main` this errored on Postgres via squirrel's `LIMIT 18446744073709551615`; without the fix it would have silently returned every row.)

## Breaking changes

FSC tags each module separately, so this needs its own root-module tag to be consumable downstream.

- Nine exported constructors no longer take a `ph sq.PlaceholderFormat` argument: `NewAuditInfoStore`, `NewBindingStore`, `NewKeyValueStore`, `NewSignerInfoStore`, `NewSimpleKeyDataStore` (in `…/storage/driver/sql/common`), and `NewEndorseTxStore`, `NewEnvelopeStore`, `NewMetadataStore`, `NewVaultStore` (in `platform/fabric/services/db/driver/sql/common`).
- **Removed:** `pagination.ApplyToSquirrel`; `NewConditionInterpreter` and `NewPaginationInterpreter` in both `sql/sqlite` and `sql/postgres`; the whole `platform/view/services/storage/driver/sql/query` tree (`select`, `insert`, `update`, `delete`, `cond`, `common`).
- **Moved** (the tree held nothing else once squirrel left):
  - `…/storage/driver/sql/query/pagination` → `…/storage/driver/pagination`
  - `…/storage/driver/sql/query/common.FieldName` → `…/storage/driver.FieldName` (same underlying `string` type)
- **Unchanged:** the runtime API of `pagination` (`None`, `Offset`, `Keyset`, `Page`) and the `driver.Pagination` interface — only the import path moves.

## Supersedes

**#1479** (open) plans to copy the query builder into `platform/fabric/services/db/driver/sql/query/` and then remove it from the view storage driver. Its final unchecked step is what this PR does — by deleting the tree rather than relocating it, which makes the copy in its first two steps unnecessary. Suggest closing or re-scoping #1479 and closing draft PR **#1478**. #1479 cites `fabric-token-sdk#1734` as related; that downstream need no longer applies — see below.

## Downstream impact: none (verified)

[Panurus](https://github.com/LFDT-Panurus/panurus) is the downstream consumer, and it needs **no source change**. Checked against the Panurus tree at `b6ebf6cad`:

- **Build parity.** Panurus's root module was built against FSC `main` and against this branch (via throwaway `go.work` files). Output is byte-identical — this branch introduces zero new build failures.
- **Full break surface covered.** An exported-API diff of all nine FSC packages Panurus imports shows only four differ: `driver` (*additive* — gains `FieldName`), `sql/common` (the five constructor signatures), and `sql/postgres` + `sql/sqlite` (lose the two interpreter constructors). Panurus has **zero call sites** for any changed constructor, and everything it uses from FSC's postgres/sqlite survives (`Config`, `Persistence`, `Opts`, `DbProvider`, `NewDbProvider`, `NewConfigProvider`, `ErrorMapper`, `StartPostgres`, `DefaultConfig`, `WithDBName`, `ContainerConfig`, `Open`).
- **Its interpreter calls are its own.** Panurus mentions `NewConditionInterpreter` in 35 files, but every call site resolves to its own fork at `token/services/storage/db/sql/{postgres,sqlite}`, not FSC's.
- **Zero imports** of the deleted `sql/query` tree, and zero imports of FSC's `pagination` package — so the move is invisible to it.
- **squirrel leaves its graph too.** `go mod why` shows squirrel reaches Panurus *only* through FSC's `storage/driver/sql/common`; its own forked builder is already squirrel-free. Its eight `// indirect` requirements go stale and `go mod tidy` drops them. That is the only follow-up on the Panurus side, and it is mechanical.

Unrelated but worth flagging for whoever bumps FSC in Panurus: Panurus does not compile against FSC `main` today, for two reasons predating this branch — `Session.SendWithContext`/`SendErrorWithContext` no longer exist, and `events.NewListenerManager` gained a `ctx` parameter and changed three types.

## Testing

- `make lint` / `make checks`: exit 0, "0 issues" across all four modules. (Note: clear `golangci-lint cache clean` first — the cache stores absolute paths and replays stale ones from other worktrees, silently dropping `//nolint` directives and generated-file exclusions.)
- `make unit-tests`: exit 0, 143 packages ok, no failures, no data races.
- `make unit-tests-postgres`: exit 0, 123 packages ok.
- Fabric-x integration suites: all five pass (`fabricx-iou`, `fabricx-atsa`, `fabricx-simple`, `fabricx-deployment`, `fabricx-multiendorsement`), ginkgo summaries confirmed rather than exit codes trusted.
- `go fix -diff ./...`: clean on the root module and on the new `sqlbuild` package, identical to `main`.
- No `Masterminds/squirrel` reference remains in any `.go`, `go.mod` or `go.sum`.

## Follow-ups (not in this PR — all pre-existing on `main`)

- A `pos`-keyed keyset over `GetAllTxStatuses` cannot advance: `pagination/page.go` builds the next cursor by reflecting a `Pos` field out of the returned rows, but `driver.TxStatus` has only TxID/Code/Message, so a non-empty page panics. Adding the column back would be worse — a fourth column that `rows.Scan(&s.TxID, &s.Code, &s.Message)` cannot absorb.
- Several `sql/common` constructors are called with `(readDB, writeDB)` swapped, and the unit tests cannot detect it because they pass the same `*sql.DB` as both handles.
- `pagination.NewPage` type-switches on `*keyset[int, any]` / `*keyset[string, any]` only and panics on any other keyset type.


mbrandenburger:
@sid200727 can you please have a look at this PR and do a review?

sid200727:
@mbrandenburger went through the diff, since I wrote the squirrel migrations this replaces (#1199, #1235, #1258, #1301), figured I'd check the things I have direct context on.

The characterization-test-first approach makes sense. Pinning the 32 statements before touching anything means regressions get caught mechanically, and ~100 lines of internal builder vs the full squirrel dependency is a straightforward win.

A few things I checked:

1. **$N placeholders for both dialects** - this lines up with what I saw during #1235. SQLite was already receiving $N via common/binding.go in production, so the dialect split I implemented never actually needed separate placeholder logic for the statements we used. The builder's single-counter Format() is the right call.

2. **Stores I migrated (SignerInfo, AuditInfo, Vault)** - their statement patterns are in the characterization tests under internal/storage/, looks correct against what I originally put in.

3. **Panurus downstream** - confirmed no impact. The FSC query builder got copied into panurus as an independent fork when Angelo asked me to do that, so panurus has no squirrel references left and this removal doesn't touch it.

4. **OffsetFromRaw re-validation** - good catch, a negative page size silently returning every row wouldn't have surfaced through my migrations since I wasn't adding validation at that layer.

5. **Squirrel removal** - looks complete, only go.sum transitive hashes remain which should clean up on the next go mod tidy.

One question: the follow-ups mention Panurus doesn't currently compile against FSC main (Session.SendWithContext gone, events.NewListenerManager signature changed). Should I start preparing for that on the Panurus side, or is a separate bump PR planned?

Overall this looks solid to me - removing squirrel was overdue.

mbrandenburger:
Thank you @sid200727 for your review.

I think you are right. Panurus needs to integrate FSC [v0.19.0](https://github.com/hyperledger-labs/fabric-smart-client/releases/tag/v0.19.0) and adopt the recent breaking changes. Please feel free to move forward with this and create an issue + PR on Panurus.

adecaro (commented):
## Review: does this address #1710?

Checked out at `de744f3b5af6f44055d28fc704cc1f3c40e4400e`.

```
$ grep -rn squirrel --include='*.go' --include=go.mod --include=go.sum .
(no hits outside a historical Godoc comment referencing squirrel's error message)
```

```
$ go build ./... && go vet ./... && gofmt -l .
(clean, all three)
$ golangci-lint run ./internal/... ./platform/fabric/services/db/driver/sql/... ./platform/view/services/storage/driver/...
0 issues.
$ go test ./internal/storage/... ./platform/fabric/services/db/driver/sql/... ./platform/view/services/storage/driver/...
ok (all packages)
```

```
$ git diff c3cea6b9..HEAD -- platform/view/services/storage/driver/sql/common/auditinfo_test.go
(only the squirrel import and the constructor call change; expected SQL/params/assertions untouched)
```

### What it fixes / adds

| Claim | Verified how |
|---|---|
| squirrel dependency fully removed | grepped every `.go`/`go.mod`/`go.sum` for `squirrel` |
| `sql/query` tree deleted, not just relocated | confirmed directory absent, zero remaining imports of the path |
| Characterization tests are the untouched oracle | diffed `auditinfo_test.go`: only import/constructor lines changed |
| squirrel edge cases reproduced (`IN ()`→`(1=0)`, empty `AND`→`(1=1)`, single-elem `AND` keeps parens, `[]byte` scalar) | read `cond.go` against its test cases in `cond_test.go` — all four present and matching |
| `LIMIT`/`OFFSET` stay literals, not parameters | read `paging.go`'s `WritePaging` |
| `pagination` + `FieldName` lifted out, not dropped | confirmed both exist with unchanged runtime API; mock move is a 0/0-diff rename |
| `OffsetFromRaw` re-validates like `KeysetFromRaw` | read `offset.go`/`keyset.go`; both route through validating constructors |
| `WriteTuples`' panic on empty rows is unreachable in production | checked every call site — all guard with a length check or an equivalent precondition |
| Downstream Panurus needs no change | not independently re-verified (would need a Panurus checkout); relying on @sid200727's confirmation, who wrote Panurus's own fork |

### Blocking

None.

### Non-blocking

None found beyond what @sid200727 already raised and got answered (Panurus follow-up bump, which the author confirmed is tracked separately against v0.19.0).

### Recommendation

Merge as-is. Build/vet/format/lint/tests are all clean across every touched module, CI (including the fabric-x integration suites) is green, and the behavior-preservation claims check out against the characterization tests and the new package's implementation directly rather than by inspection alone.

adecaro (approved):
LGTM

PR #1713, opened 2026-09-03
test(fabricx): cover channel configuration updates

### LINKED ISSUE #1712: fabricx: no integration coverage for channel configuration updates
### Description

`platform/fabricx/core/channel/config` runs a `ChannelConfigMonitor` that polls the committer's query service for the latest configuration transaction and, on seeing a new one, hands the envelope to `membership.Service.Update` and reconfigures the ordering service from the new Orderer group.

Nothing exercises that update path. Every fabricx suite starts on a channel's genesis configuration and never changes it, so the monitor's "a new configuration arrived" branch — including the cross-check between the committer's out-of-band `Version` and the sequence carried inside the envelope's signed content, and the capability check in `membership.capabilitiesSupported` — has never run in a test.

Fabric has this coverage, added in #1696. Fabricx does not.

### Why it matters

The monitor is the only thing that keeps a running FSC node's view of a channel in step with the channel itself. Two failure modes it is meant to prevent are both currently untested:

- a node silently *not* following a configuration change, so it goes on transacting against a configuration that is no longer in force;
- a node adopting a configuration it cannot support, rather than refusing it and continuing to serve the one it holds.

A regression in either would not be caught by any existing suite.

### What coverage should establish

- Every FSC node follows a configuration change made while the network is running, and can still transact afterwards.
- More than one update in sequence works, not just the first.
- A node that was down when an update landed converges on it. (Note fabricx has no vault replay on start-up — the monitor's first poll fetches only the latest configuration — so the property to test is convergence, not re-application of what the node already had.)
- A configuration the nodes cannot support is refused, and the configuration already in force keeps working.

### Implementation Steps

- A `integration/fabricx/configupdate` suite wired into `make integration-tests-*`.
- The specs fail if the config monitor stalls or dies — a spec asserting only that nodes *stay* at some sequence passes trivially when nodes receive nothing, so adoption of a good configuration must be proven on the same network first.
- Whatever submits the update lives in `integration/nwo/fabricx`: Fabric-X has no client-side component that computes, signs and broadcasts a channel configuration update, so the harness needs one.

### Additional Information

_No response_

mbrandenburger:
Closes #1712.

Adds an integration suite covering channel configuration updates on the fabricx platform — the counterpart to #1696 for fabric. Until now the fabricx channel config monitor's update path only ever ran for a channel's genesis configuration.

## The specs

**T1 — the nodes follow the configuration.** Changes `BatchTimeout` on a running network three times, asserting each time that every FSC node reports the new sequence and that the IOU flow still works. The third change lands while `borrower` is stopped, so it also covers a node converging on a configuration it missed.

**T2 — the nodes refuse what they cannot support.** Submits a configuration requiring an Application capability no fabric-x binary defines. The committer accepts it, having no capability check of its own; the nodes must refuse it and go on serving what they hold.

T2 applies an ordinary update first and asserts adoption. Without that, a dead config monitor would leave every node at sequence 0 regardless of the poisoned update and the spec would pass for the wrong reason — fault injection confirms it, since raising `pollInterval` to 24h fails both specs, and before that step existed T2 passed under the fault.

## Harness

`integration/nwo/fabricx/network/configupdate.go` reads the committer's configuration and computes, signs and broadcasts an update; Fabric-X has no client-side component that does this. Two gotchas are documented in place: the envelope is assembled by hand because `protoutil.CreateSignedEnvelope` leaves the channel header's `TxId` empty, which the committer's sidecar silently drops as `MALFORMED_MISSING_TX_ID`, and `UpdateConfig` takes a variadic `orgs` because an Application-group value needs peer-org signatures no orderer identity can supply.

## Testing

`make integration-tests-fabricx-configupdate` — `Ran 2 of 2 Specs in 187.909 seconds`,
`2 Passed | 0 Failed`. `make checks` and `make lint` — 0 issues across all four modules.
`make tidy` — no diff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,6 +209,7 @@ INTEGRATION_TARGETS += fabricx-atsa
 INTEGRATION_TARGETS += fabricx-simple
 INTEGRATION_TARGETS += fabricx-deployment
 INTEGRATION_TARGETS += fabricx-multiendorsement
+INTEGRATION_TARGETS += fabricx-configupdate
 
 .PHONE: list-integration-tests
 list-integration-tests: ## List all integration tests
--- a/integration/fabric/configupdate/configupdate.go
+++ b/integration/fabric/configupdate/configupdate.go
@@ -32,11 +32,11 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/protoutil"
 )
 
-// batchTimeoutKey is the key of the BatchTimeout value in the channel config's Orderer group.
-const batchTimeoutKey = "BatchTimeout"
+// BatchTimeoutKey is the key of the BatchTimeout value in the channel config's Orderer group.
+const BatchTimeoutKey = "BatchTimeout"
 
-// ordererGroupKey is the key of the Orderer group in the channel config.
-const ordererGroupKey = "Orderer"
+// OrdererGroupKey is the key of the Orderer group in the channel config.
+const OrdererGroupKey = "Orderer"
 
 // channelHandles are the nwo handles needed to submit a configuration update: the network,
 // the channel to update, an orderer to submit through, and a Fabric peer whose local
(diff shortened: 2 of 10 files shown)


Rozerxshashank (commented):
LGTM!

Rozerxshashank (approved):

adecaro (approved):
LGTM

PR #1716, opened 2026-09-03
test(integration): run libp2p in four targets instead of nine

### LINKED ISSUE #1715: Integration tests run libp2p in nine targets; the cheap libp2p unit tests never run at all
### Description

FSC ships two p2p comm implementations, `websocket` (primary) and `libp2p` (optional, its own nested Go module). Nine of seventeen integration targets assert both by declaring two `Describe` blocks over the same scenario.

That is more expensive than it looks. `integration.TestSuite` wires `BeforeEach(s.Setup)` / `AfterEach(s.TearDown)`, so **every `It` pays a full network generate/start/stop cycle**. A libp2p `Describe` in a Fabric suite is therefore an extra Fabric bootstrap — cryptogen, orderer, peers, channel create, chaincode install — not a cheap re-assertion.

Meanwhile the cheap parity check exists and never runs. Both hosts share a conformance suite (`comm.P2PLayerTestRound`, `SessionsTestRound`, `SessionsForMPCTestRound`, `SessionsMultipleMessagesTestRound`, `SessionsNodesTestRound`), and the libp2p side additionally covers 3-node bootstrap discovery. But `platform/view/services/comm/host/libp2p` is a nested module, so the root `go list ./...` excludes it and no Makefile target or workflow steps into it. **Those tests have never executed in CI.**

Some of the duplication buys nothing at all: `fabric/atsachaincode` registers no responders and never calls `GetSession`, so no FSC-to-FSC traffic crosses the p2p layer in either variant.

### Implementation Steps

Proposal: assert host parity in the unit tests that already exist, keep libp2p end-to-end only where it earns its cost, and label specs by comm type so a suite can be driven one transport at a time.

### Additional Information

_No response_

mbrandenburger:
Closes: #1715

## What

- Every `Describe` in the affected suites carries a comm-type label, so a suite can be run one transport at a time (`ginkgo --label-filter=libp2p`) and split into parallel CI entries.
- libp2p dropped from `fabric/atsa`, `fabric/atsachaincode`, `fabric/iouhsm`, `fabric/twonets`, `fabric/stoprestart`.
- `fsc/pingpong`'s libp2p sweep trimmed 6 → 2 specs; the other four vary the *client* API (REST, web client, artifact generate-vs-load, stream-vs-call), which is transport-agnostic and stays fully covered under websocket.
- `fabric-iou` split into `fabric-iou` (websocket) + `fabric-iou-libp2p`, so keeping both transports there doesn't set PR wall-clock. The websocket target uses `--label-filter='!libp2p'` so a spec added later without a label still runs somewhere.
- **`make unit-tests` now executes the libp2p host module** — the enabling change. `ok .../comm/host/libp2p`, 52.4% coverage.
- A fixed `time.Sleep(7 * time.Second)` in `TestLoadAndStreamWebsocket` replaced with a poll of the stream call it was waiting for.
- nwo harness defaults to websocket; four docs corrected (they claimed libp2p was "the default" — there is no default; `provider.NewHostProvider` rejects an unset `fsc.p2p.type`).

Net: 5 Fabric bootstraps and 4 nwo cycles removed. libp2p end-to-end goes from 9 targets to 4 (`fsc-pingpong`, `fsc-stoprestart`, `fsc-signedpingpong`, `fabric-iou-libp2p`).

## Why this is safe

Three of the five deletions are provably free: `atsachaincode` has no sessions at all; `iouhsm`'s HSM covers the *Fabric* identity while libp2p derives its peer key from `fsc.identity.key.file`; `fabric/stoprestart` duplicates `fsc/stoprestart`. `fabric/iou` retains the richest p2p flow — 4 nodes, recipient-identity exchange plus endorsement collection — under both transports.

Two residual gaps, both low risk: libp2p alongside two Fabric SDK graphs (`twonets`), and libp2p under `atsa`'s larger `services/state` payloads.

## Verified

pingpong 18/18 · iou 1/1 + 3/3 · signedpingpong 2/2 · stoprestart 4/4 · atsachaincode 1/1 · twonets 1/1 · `make unit-tests` green with libp2p executing · `make checks` (includes lint) 0 issues across all four modules · `make tidy` no `.mod`/`.sum` churn.

`fabric/atsa` and `fabric/iouhsm` are dry-run only — atsa fails locally on a pre-existing legacy-chaincode build issue unrelated to this branch, iouhsm needs SoftHSM. Both left to CI.

## Notes for review

- **`P2PCommunicationType` default flips to websocket.** Downstream consumers of `nwo` (Panurus) whose topologies omit the field will silently change transport.
- **`fsc/pingpong`'s committed testdata stays libp2p.** Moving it to websocket needs regenerated crypto: the fixture's signing cert has no IP SANs and the websocket host derives P2P TLS material from `fsc.identity.cert.file`. The node-based `Describe` is labelled `fsc.LibP2P` to match, with the reason recorded inline.
- Known follow-ups: gitignore `*.test` / `*.profile` / `pids.txt`; connection leak in `platform/view/services/web/client/client.go:131-141`; vestigial `Label("T1")` in `configupdate` / `runtimeconfig`; the seven one-element-loop simplifications flagged by the over-engineering review.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -117,7 +117,7 @@ jobs:
 
       - name: Set up monitoring
         # we only need this for a subset of integration tests
-        if: contains(fromJSON('["fabric-iou", "fabric-runtimeconfig", "fsc-stoprestart"]'), matrix.tests)
+        if: contains(fromJSON('["fabric-iou", "fabric-iou-libp2p", "fabric-runtimeconfig", "fsc-stoprestart"]'), matrix.tests)
         run: |
           make -j4 pull-images-monitoring
 
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,10 @@ include $(TOP)/checks.mk
 # we use a multi-module repo structure here and therefore need to carefully collect packages for unit tests
 GO_PACKAGES = $$(go list ./...)
 NWO_PACKAGES = ./nwo/...
+# The libp2p comm host is its own module, so the root `go list ./...` cannot see
+# it. Named here so `unit-tests` can step into it explicitly -- without this its
+# host-conformance tests (shared with the websocket host) never run.
+LIBP2P_HOST_MODULE = platform/view/services/comm/host/libp2p
 GO_PACKAGES_SDK = $$(go list ./... | grep '/sdk/dig$$')
 GO_TEST_PARAMS ?= -race -cover
 TEST_PKGS ?= $(GO_PACKAGES)
@@ -167,6 +171,7 @@ unit-tests: ## Run unit tests
 	rc=0; \
 	go test $(GO_TEST_PARAMS) $(GO_COVERPKG) --skip '(Postgres)' $(TEST_PKGS) || rc=1; \
 	go test -C integration $(GO_TEST_PARAMS) --skip '(Postgres)' $(NWO_PACKAGES) || rc=1; \
+	go test -C $(LIBP2P_HOST_MODULE) $(GO_TEST_PARAMS) ./... || rc=1; \
 	exit $$rc
 
 .PHONY: unit-tests-postgres
@@ -196,6 +201,9 @@ INTEGRATION_TARGETS += fabric-atsachaincode
 INTEGRATION_TARGETS += fabric-configupdate
 INTEGRATION_TARGETS += fabric-events
 INTEGRATION_TARGETS += fabric-iou
+# fabric/iou runs both comm types. They are split into two targets so CI runs
+# them as parallel matrix entries rather than serially in one job.
+INTEGRATION_TARGETS += fabric-iou-libp2p
 INTEGRATION_TARGETS += fabric-runtimeconfig
 INTEGRATION_TARGETS += fabric-stoprestart
 INTEGRATION_TARGETS += fabric-twonets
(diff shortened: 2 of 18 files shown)


adecaro (commented):
## Review: does this address #1715?

Checked out at `6525c09a2ffc89e76cf02c6fecb06d9243116581`.

```
$ go test ./platform/view/services/comm/host/libp2p/...
ok  	github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/host/libp2p	72.666s
```
This module previously had zero CI signal (nested Go module, invisible to root `go list ./...`); `Makefile`'s new `go test -C $(LIBP2P_HOST_MODULE) ./...` step in `unit-tests` is what makes it run now, and it passes.

```
$ cd integration/fabric/iou && ginkgo --dry-run --label-filter=libp2p .
Will run 1 of 4 specs
$ ginkgo --dry-run --label-filter='!libp2p' .
Will run 3 of 4 specs
```
Clean partition (no gap, no overlap) confirming the `Label(c)` scheme actually drives `--label-filter='!libp2p'` / `--label-filter=libp2p` the way the Makefile's new `INTEGRATION_FLAGS_fabric-iou{,-libp2p}` expects.

```
$ make list-integration-tests
... fabric-iou
fabric-iou-libp2p
...
```
Matches the CI matrix's `itest (fabric-iou-libp2p)` job, which is green.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| libp2p dropped from 5 targets (atsa, atsachaincode, iouhsm, twonets, stoprestart) | Read each diff + dry-ran ginkgo; each has a concrete, checkable rationale (e.g. `atsachaincode` never opens a session, so no p2p traffic crossed either transport) |
| libp2p host-conformance unit tests now actually run in CI | `Makefile` diff + local `go test` run, 72s (a real network bootstrap, confirming it wasn't run before) |
| `nwo` P2P default flips libp2p → websocket | Checked `platform/view/services/comm/provider/provider.go:30-36` — the root module's host provider only has a `websocket` case; the old `LibP2P` default made any `viewsdk.NewSDK`-based consumer (e.g. `integration/benchmark/node/benchmark.go`) fail at boot with "unknown p2p type: libp2p". This is a real fix, not just a doc correction |
| Flaky `time.Sleep(7s)` replaced with a poll | Read `TestLoadAndStreamWebsocket` — now `Eventually`-retries the actual `StreamCallView` call |

### Blocking

None.

### Non-blocking

- The PR body's own "Notes for review" already discloses the one thing I'd otherwise raise — that the `P2PCommunicationType` default flip affects any downstream `nwo` consumer (e.g. Panurus) whose topology omits the field. I independently found the same thing (grepped for `NewTopology()` callers that never set the field) and have nothing to add to that disclosure.
- Nit: the libp2p module's own `coverage.profile` (from the new `unit-tests` step) isn't uploaded to Codecov — only the root profile is. Not a regression, just a gap if libp2p coverage tracking matters later.
- Nit: the four dropped `fsc/pingpong` libp2p specs ("transport-agnostic, covered under websocket") are the least mechanically-provable trim in this PR — reasonable, but more of a judgment call than the session-free removals.
- Already acknowledged as out-of-scope follow-ups in the PR body: vestigial `Label("T1")` in `configupdate`/`runtimeconfig`, a connection leak in `platform/view/services/web/client/client.go:131-141`, gitignore entries for build artifacts.

### Recommendation

Merge as-is — build, vet, lint, and the newly-wired unit tests all pass, the label-filter partitioning behaves exactly as documented, and every trimmed spec has a rationale that held up under inspection.

adecaro (approved):
LGTM

PR #1718, opened 2026-09-04
fix(fabricx): report the status of an already-final transaction

### LINKED ISSUE #1714: fabricx: AddFinalityListener reports Unknown for an already-final transaction
### What happened?

The borrower's `"update"` view (`integration/fabric/iou/views/borrower.go:112`) never completes. It stalls while collecting endorsements: the lender's Org3 MSP identity, read back out of the IOU state, cannot be mapped onto the lender's FSC node, so the view opens a session nobody is listening on and waits for a reply that never arrives. It hangs rather than failing, so it burns a whole CI timeout when hit.

Fingerprint, logged repeatedly on the borrower:

```
failed to get identity for label Org3MSP<cert for lender@org3.example.com>: identity not found
```

The "label" there is the serialized MSP identity itself, passed where a node label is expected.

### How can we reproduce it?

In `integration/fabricx/iou/iou_test.go`, after a successful create:

```go
iouState := iou.CreateIOU(s.II, "", 10, "approver1")
iou.UpdateIOU(s.II, iouState, 5, "approver2")   // hangs here
```

Add a `SpecTimeout` first — `integration/fabric/iou`'s helpers call `CallView` with no context, so nothing bounds the wait and Ginkgo's 60-minute default is what stops it.

### Environment

- FSC version: main:
- Go Version: go1.26.5 darwin/arm64
- Operating system: MacOS

## Why this is invisible today

No fabricx suite exercises the update flow: `integration/fabricx/iou/commands_test.go` defines `UpdateIOU`/`UpdateIOUWithBorrower` and never calls them, and there were no other callers under `integration/fabricx/`. Four non-x suites use the same helpers routinely and pass. This is a coverage gap, not a regression.

## Root cause

`Context.GetSession` tries three candidate identities. Candidate 2 (`c.resolve(party)` → `resolverByIdentity`, `platform/view/services/endpoint/service.go:507`) misses — on fabricx there is no binding from the lender's Org3 identity to its FSC node. Candidate 3 passes the serialized identity as a *label* and produces the warning above.

`Sessions.GetFirstOpen` then skips the failing candidates **without updating `targetId`**, so it falls through to `createSession` with candidate 1: the raw, unresolvable identity. An unresolvable party therefore yields a *successfully created* session instead of an error, turning "I don't know who that is" into an unbounded stall.

`create` works because it begins with `ExchangeRecipientIdentities`, which binds the mapping at runtime. `update` has no such step and expects the stored identity to resolve already.

Ruled out: the fabricx driver does load the endpoint resolver service, the fabricx network does generate the resolver map, and the two IOU topologies place the same roles in the same organizations.

## Suggested fixes — probably both

1. **The binding.** Why is it absent on fabricx when the same resolver machinery runs on both platforms? Cheapest decisive experiment: dump the Org3 identity bytes stored as the state's owner and the bytes `resolversMap` is keyed on, and diff them — that map is keyed on exact bytes, so any difference in serialization or in which certificate is used is fatal.
2. **Fail fast.** `GetSession` should not create a session toward an identity nothing resolved. Worth doing regardless of (1): it turns this class of bug from a silent twenty-minute stall into an immediate, legible failure.

## Definition of done

- `iou.UpdateIOU` succeeds in a fabricx suite, exercised by a spec that actually calls it.
- The spec fails promptly, with a diagnosable message, if resolution regresses.

## References

- **#1713** — where this surfaced. Its `integration/fabricx/configupdate` suite is the first thing under `integration/fabricx/` that would have called `UpdateIOU`. That PR works around the stall rather than fixing it, using fresh `iou.CreateIOU` calls at distinct amounts, which exercise the same endorse / validate / broadcast / commit path and state query on both parties without depending on this one.
- A fuller analysis — the verified call chain, the ruled-out hypotheses, three ranked root-cause hypotheses — is on branch `fix/fabricx-iou-update-identity-resolution` at `docs/dev/issues/fabricx-iou-update-identity-resolution.md`.

mbrandenburger:
Closes #1714.

On fabricx, `AddFinalityListener` for a transaction the committer had already reported answered `Unknown` rather than that transaction's real status, breaking the contract in `platform/common/driver/committer.go` — *"OnStatus is called when the status of a transaction changes, or it is already valid or invalid."* fabric honours it; fabricx did not.

## The defect

The committer's notification stream reports a transaction once. `notificationListenerManager.dispatch` drops its handler entry on delivery, so a later registration sent a fresh `TxStatusRequest` the committer would not answer, and after `requestTimeout` the timeout verdict became `Unknown`. Nothing in the finality path could say otherwise — `vault.Vault.Statuses` already resolves status through the query service, but was wired only to `VaultService`.

It already fired on **every successful create**, in `ApproverView`, silently: the approver is a responder whose view completion nothing waits for. It failed a spec only on `update`, where the second registration sits on the borrower.

## The fix

`ResolvingListenerManager` wraps the notification manager: resolve status first, subscribe only when the transaction is not yet final, and treat a later `Unknown` from the notification path as a reason to ask again rather than as a verdict. It is composed in `channel/provider.go`, where both collaborators already existed, so no dependency plumbing changed and `vault.Vault` satisfies the new `TxStatusResolver` interface unchanged.

No polling and no cache. The query service is authoritative at every point, so there is no cache-expiry cliff; a cache would be a latency optimisation, and there is no evidence it is needed.


adecaro (commented):
## Review: does this address #1714?

Checked out at `3d99231da30b0acaff6b429cbc20095ddee64318` (up to date with `main`).

```
go test ./platform/fabricx/core/finality/...
ok  github.com/hyperledger-labs/fabric-smart-client/platform/fabricx/core/finality  4.439s   (15/15 new cases pass, incl. duplicate-registration, remove-after-settle, stale-proxy-after-re-register, resolver-error-degrades-to-notifications)
```

```
golangci-lint run ./platform/fabricx/core/channel/... ./platform/fabricx/core/finality/...
./fabric/common/views/... ./fabric/iou/... ./fabric/iou/views/... ./fabricx/iou/... ./fabricx/simple/views/...
0 issues.
```

```
go build ./...   (root and integration module)
go vet ./...      (changed packages, both modules)
gofmt -l -s       (changed files)
-> all clean
```

### What it fixes / adds

| Claim | Verified how |
|---|---|
| A finality listener registered after the committer already reported a tx's status got `Unknown` instead of the real status, violating the `OnStatus`/"already valid or invalid" contract | Read `platform/common/driver/committer.go`'s contract, `resolving.go`'s resolve-before-subscribe logic, and the wiring in `channel/provider.go` that puts `ResolvingListenerManager` on the `ch.Committer()` path |
| Fix resolves through the query-service-backed vault before subscribing, and re-resolves once on a later `Unknown` rather than treating it as a verdict | `resolving_test.go`: `TestResolvingSettlesAlreadyFinalWithoutSubscribing`, `TestResolvingReResolvesOnUnknown`, `TestResolvingPassesUnknownThrough` |
| `iou.UpdateIOU` now exercised on fabricx, using `approver1` per the namespace-policy constraint | `integration/fabricx/iou/iou_test.go` diff; matches #1714's definition of done exactly |
| Three near-duplicate integration `FinalityListener`s (which discarded unexpected statuses and slept inside `OnStatus`) replaced by one context/timeout-bound implementation | `integration/fabric/common/views/finality_listener.go` + its test file; call sites in `approver.go`/`borrower.go`/`create.go` now check the registration error and `defer` removal |

### Non-blocking

None — no findings survived review. Concurrency edge cases (stale proxy after re-registration, forget-on-deliver bookkeeping, re-resolve staying on the calling goroutine to avoid a teardown fan-out) are explicitly covered by tests, and the interface fit between `vault.Vault.Statuses` and `TxStatusResolver` is exact with no adapter needed.

### Recommendation

Merge as-is — matches the issue's revised root-cause analysis and definition of done, CI is green including `itest (fabricx-iou)`, and my own build/vet/lint/test verification found nothing to add.

adecaro (approved):
LGTM

PR #1720, opened 2026-09-04
fix(node): fix two profiling bugs found while covering error paths

### LINKED ISSUE #1719: test: node package coverage
### Description

This issue asks for improving `node` package test coverage

### Implementation Steps

- Identify missing tests
- Implement

### Additional Information

_No response_

mbrandenburger:
Closes #1719 

Raise node package coverage from 83% to 98% by testing the error paths that were previously unexercised. Two of those paths turned out to be buggy:

- profile: `Stop()` left runtime.MemProfileRate at the profiling rate instead of the value it found. With both memory profiles enabled, `startMemProfile("allocs")` captured the rate that `startMemProfile("heap")` had already overwritten, and its closer ran last, so the process was left at `DefaultMemProfileRate`. Only the profile that actually changes the rate now restores it.
- profile: `startCPUProfile` leaked the profile file when `pprof.StartCPUProfile` failed. Close it before returning.

New tests:

- node: both panic branches in `listen()`, via a helper that recovers; the nil-error cases reuse the same helper.
- start: the `profiler.Start()` failure path, reached by pointing `FSCNODE_CFG_PATH` at a regular file so `MkdirAll` fails.
- profile: the option-error return from New; the `os.Create` failure in every start*Profile plus its propagation out of Start, reached by planting a directory where each profile file would go; the "cpu profiling already in use" branch; and the Sync/Close/WriteTo error branches inside the closers, reached by calling `Stop()` twice, which also asserts the MemProfileRate round-trip.

profile.go reaches 100%. The two remaining uncovered statements (`os.Exit` in Node.Execute, and the `profile.New` error path in `serve()`, which is unreachable because `configPath` defaults to "./") would each need a test seam in production code, so they are left alone.

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 package node
 
 import (
-	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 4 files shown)


adecaro (commented):
## Review: does this address #1719?

Checked out at `3bd4368be69ff0e788a8d7a9da5a420a56b1ecec` (up to date with `main`).

```
$ go test -race -count=1 ./node/...
ok  	node
ok  	node/start
ok  	node/start/profile
ok  	node/version

$ go tool cover -func (node/start/profile)
100.0% of statements
```

Revert-test: reverted the `startMemProfile` rate-restoration fix and re-ran `TestStopTwice` — it fails exactly as expected (`expected: 524288, actual: 409`), confirming the new test catches the real bug being fixed.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Leaked CPU profile file handle when `pprof.StartCPUProfile` fails | Read `profile.go:134-137`; `os.Create`'d file was never closed on that error path before this fix |
| `runtime.MemProfileRate` left un-restored when both heap and allocs profiling are enabled (`WithAll()`) | Traced the double-closer-overwrite bug in the old code; reverted the fix and watched `TestStopTwice` fail with exactly that symptom |
| Node package coverage improvement (issue #1719) | `node/start/profile` now at 100.0% statement coverage; new tests cover option-validation errors, `os.Create` failures for all 5 profile kinds, double-start collision, `Stop()` idempotency, and both `listen()` panic paths |

### Blocking

None.

### Non-blocking

- Nit (`profile.go:161-169`): the mem-profile restore closer is appended before the write closer for the same profile type, so `Stop()` restores the rate before writing the profile to disk. Harmless — `WriteTo` dumps already-recorded samples regardless of current rate — just reads slightly inconsistently next to the CPU/mutex/block closers, which restore after writing. Not worth changing for this PR.

### Recommendation

Small, well-tested, correctly-scoped bug fix with a revert-verified regression test for each bug. Looks ready to merge as-is (pending maintainer review per repo process).

adecaro (approved):
LGTM

PR #1722, opened 2026-09-04
fix(fabricx): retry the block-index lookup after finality

### LINKED ISSUE #1721: fabricx-deployment and fabricx-simple are flaky: GetTransactionByID returns NotFound for a transaction that is already final
### Description

`integration/fabricx/simple/views.CreateView` queries the committer's block query service the instant its finality listener reports Valid. That fails intermittently:

```
Wait for txID=74b5... to be committed        09:37:05.745
Ledger info: height=3                        09:37:05.746
failed to get tx for txID [74b5...]: rpc error: code = NotFound
  desc = transaction ID [74b5...]: not found in index: tx not found
```

The committer reports finality from its notification service, but appends and indexes the block into its block store (`fabric-x-common/common/ledger/blkstorage`) independently. not found in index is blkstorage.ErrNotFound — the block store, not the query service. Height 3 means blocks 0..2 while the transaction lands in block 3, so at query time its block was not yet appended.

Instrumenting the view locally reproduces the lag on the first create of a run: attempt 1 fails at 10 ms, attempt 2 succeeds at 619 ms, and the transaction is then reported in block 3. Later creates in the same run usually succeed on the first attempt — it is a timing race, not a functional break.

The race is not new. `time.Sleep(5 * time.Second)` inside the old `utils.FinalityListener.OnStatus` had been giving the index time; `9701422c` (#1718) removed it because the committer contract forbids sleeping in OnStatus. The ledger-query block in that view had most likely never run without it.

Seen in CI on main after `9701422c`. Reproduction is probabilistic — the CI failure had 1 ms between finality and the query.

### Implementation Steps

- Confirm the issue
- Probably use a retry mechaism

### Additional Information

_No response_

mbrandenburger:
Closes #1721.

The committer reports finality from its notification service but indexes blocks into its block store independently, so `GetTransactionByID` can answer `NotFound` for a transaction that is already final. `CreateView` queried it 1 ms after finality.

- `integration/fabricx/simple/views/create.go` retries `GetTransactionByID` (30 × 500 ms) until the index catches up. `GetBlockNumberByTxID` reads the same txID index, so it needs no retry of its own.
- `platform/fabricx/core/ledger.GetTransactionByID` documents the lag, so the next caller does not write the same bug.

No production behaviour change — the platform still returns `NotFound` rather than blocking, because how long to wait is the caller's policy.

Testing: fabricx-deployment, fabricx-simple, fabricx-multiendorsement pass locally; make checks clean. Verified against the failure by instrumenting the retry: attempt 1 fails at 10 ms, attempt 2 succeeds at 619 ms.

### DIFF
--- a/integration/fabricx/simple/views/create.go
+++ b/integration/fabricx/simple/views/create.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	views2 "github.com/hyperledger-labs/fabric-smart-client/integration/fabric/common/views"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/services/logging"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
@@ -124,7 +125,9 @@ func (i *CreateView) Call(viewCtx view.Context) (any, error) {
 	}
 	logger.Infof("Ledger info: height=%v", info.Height)
 
-	pt, err := l.GetTransactionByID(tx.ID())
+	// The block index lags finality -- see ledger.GetTransactionByID.
+	pt, err := utils.NewTypedRetryRunner[fdriver.ProcessedTransaction](30, 500*time.Millisecond, false).
+		Run(func() (fdriver.ProcessedTransaction, error) { return l.GetTransactionByID(tx.ID()) })
 	if err != nil {
 		return nil, err
 	}
--- a/platform/fabricx/core/ledger/ledger.go
+++ b/platform/fabricx/core/ledger/ledger.go
@@ -58,6 +58,10 @@ func (c *ledger) GetLedgerInfo() (*driver.LedgerInfo, error) {
 }
 
 // GetTransactionByID returns the processed transaction for the given transaction ID.
+//
+// The committer indexes blocks into its block store independently of the finality it
+// reports, so a caller that queries right after finality must retry on
+// finality.TxNotFound.
 func (c *ledger) GetTransactionByID(txID string) (driver.ProcessedTransaction, error) {
 	env, err := c.client.GetTxByID(c.baseCtx, &committerpb.TxID{TxId: txID})
 	if err != nil {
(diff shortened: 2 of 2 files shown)


adecaro (commented):
## Review: does this address #1721?

Checked out at `1e856b610be290e6b3bf9c6ec4de208ed31d8661`.

Yes. #1721 diagnosed that `integration/fabricx/simple/views.CreateView` queried the committer's block-index (`ledger.GetTransactionByID`) immediately after finality was reported, but the committer indexes blocks into its block store independently of finality notification -- causing an intermittent `NotFound` (fabricx-simple / fabricx-deployment flakiness in CI). This PR wraps the lookup in a bounded retry and documents the lag on `ledger.GetTransactionByID`'s Godoc.

```
go build ./integration/fabricx/... && go build ./platform/fabricx/... && go vet ./platform/fabricx/...
(clean, no output)
```

```
gofmt -l integration/fabricx/simple/views/create.go platform/fabricx/core/ledger/ledger.go
(clean, no output)
```

Read `pkg/utils/retry.go`: `NewTypedRetryRunner(30, 500*time.Millisecond, false).Run(...)` retries on any non-nil error, fixed 500ms delay, up to 15s total -- comfortably above the 619ms lag from the issue's repro. Confirmed `GetTransactionByID` has exactly one caller in the repo, and the immediately-following `GetBlockNumberByTxID` call is transitively protected (it hits the same index, which is already populated by the time the retry above succeeds).

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Block-index lookup races finality notification | Confirmed by reading `ledger.GetTransactionByID` (wraps `finality.TxNotFound` when the committer's `GetTxByID` fails) and the issue's repro numbers |
| Retry gives the index enough time without reintroducing the banned `OnStatus` sleep | Retry lives in the test view's `Call`, not in `FinalityListener.OnStatus` -- confirmed by reading the diff context in `create.go` |
| Fix doesn't break the build / existing callers | `go build`/`go vet` clean; only caller of `GetTransactionByID` is the one updated |

### Blocking

None.

### Non-blocking

- Nit: the new Godoc on `ledger.GetTransactionByID` says a caller "must retry on `finality.TxNotFound`", but the retry added in `create.go` is unconditional on error type -- it retries on any error the call returns, not just that one. Harmless in this test view (worst case a permanent error burns the full ~15s before surfacing), just slightly more specific wording than the actual behavior.

### Recommendation

Merge as-is -- minimal, correctly targeted fix for a well-diagnosed flake, no production code path touched.

adecaro (approved):
There is a nit but LGTM

PR #1724, opened 2026-09-04
test(integration): keep the fabric and fabric-x tools apart under FAB_BINS

### LINKED ISSUE #1723: `make install-fabricx-tools` overwrites the fabric binaries in `$FAB_BINS`
### Description

`FAB_BINS` is a single directory, and both toolchains install into it:

| target | installs into | binaries |
|---|---|---|
| `make install-fabric-bins` | `$(FABRIC_BINARY_BASE)/bin` | `peer`, `orderer`, `configtxgen`, `cryptogen`, `osnadmin`, `discover`, … |
| `make install-fabricx-tools` | `GOBIN=$(FAB_BINS)` — *the same directory* | `fxconfig`, `configtxgen`, `cryptogen` |

Fabric-x ships a `configtxgen` and a `cryptogen` of its own, so whichever target ran last wins. A local checkout cannot have both platforms set up at the same time: running `make install-fabricx-tools` breaks every `fabric-*` suite, and re-running `make install-fabric-bins` to repair it does nothing.

### Why it does not heal

`ci/scripts/download_fabric.sh` gates the re-download on `bin/peer version`. The fabric-x install never touches `peer`, so the check passes and the script prints *"Fabric binaries already downloaded"* while `configtxgen` remains the fabric-x build. The only recovery is `rm -rf $(dirname $FAB_BINS)`, which is documented only under a different symptom (the missing `ccaas` external builder).

### How it fails

The mismatched `configtxgen` writes a genesis block the release orderer rejects, so `fabric-*` suites die in `BeforeEach` well after `cryptogen` has already run:

```
osnadmin channel join → Status: 400 "invalid join block: invalid block:
                        Header.DataHash is different from Hash(block.Data)"
peer channel fetch    → can't read the block: NOT_FOUND
```

Nothing points at the real cause. Confirming it takes a manual version comparison:

```console
$ $FAB_BINS/peer version | sed -n 2p
 Version: v3.1.4
$ $FAB_BINS/configtxgen --version | sed -n 2p
 Version: latest          # fabric-x development build
```

`fabricx-*` suites are unaffected — they use the committer all-in-one container and start no fabric orderer or peer.

### Why CI is green

Each matrix job in `.github/workflows/tests.yml` installs exactly one toolchain (`install-fabric-bins` for `fabric-*`, `install-fabricx-tools` for `fabricx-*`), so the two never share a directory there. This is a local-development-only problem.

### Implementation Steps

Install the fabric-x tools into a subdirectory of `FAB_BINS` and have the fabricx platform look there, so one exported variable serves both platforms and both install targets can be run.

### Additional Information

_No response_

mbrandenburger:
Closes #1723 

### Problem

Both toolchains ship a `configtxgen` and a `cryptogen`, and `install-fabricx-tools` installed its own on top of fabric's — `GOBIN` was `$(FAB_BINS)`, the very directory `install-fabric-bins` unpacks the fabric distribution into. Whichever target ran last won, so a local checkout could not have both platforms set up at once.

CI never noticed: each matrix job installs exactly one toolchain. Locally the state also never heals, because `download_fabric.sh` gates its re-download on `bin/peer version`, which the fabric-x install leaves untouched — so `install-fabric-bins` reports *"already downloaded"* while `configtxgen` stays the fabric-x build and every `fabric-*` suite fails generating artifacts.

### Fix

The fabric-x tools get their own subdirectory:

```
$FAB_BINS/            fabric distribution (peer, orderer, configtxgen, ...)
$FAB_BINS/fabric-x/   fabric-x tools (fxconfig, configtxgen, cryptogen)
```

Developers still export a single `FAB_BINS`, can run both install targets, and neither disturbs the other.

- `Makefile`: `FABRIC_X_BINS ?= $(FAB_BINS)/fabric-x` is the fabric-x `GOBIN`.
- `fxconfig.BinSubdir = "fabric-x"`, the single source of the name on the Go side.
- `fabric_network.Network.BinSubdir` is added to the `path.Join` already in `findCmdAtEnv`, so an empty value is exactly fabric's previous behaviour; the fabricx platform sets it in `New`. Going through the shared `Network` rather than overriding `ConfigTxGen` means every call site is covered, not just the one in `fabricx/network.GenerateArtifacts`.
- The `could not find %s` messages now name the directory that was actually searched instead of `$FAB_BINS` itself.

Deliberately a subdirectory, not a fallback: a network that declares one never reads the other toolchain's binaries, so a missing fabric-x tool is a clear error rather than a silent run of fabric's binary of the same name. The fabricx platform needs nothing else from the parent directory — it starts no peers or orderers, and `idemixgen` goes through `findOrBuild` and is built from `IBM/idemix`.

### Not included

Self-healing `download_fabric.sh` by probing `configtxgen` instead of `peer`: fabric's `configtxgen` has no `version` subcommand, so the probe would fail and re-download the 114 MB tarball on every install. Anyone whose `bin/` is already clobbered gets a troubleshooting entry with the one-time `rm -rf` instead.

### Testing

- New `integration/nwo/fabric/network/utils_test.go`: the subdirectory is honoured, the parent is not searched as a fallback, and an empty subdirectory is unchanged.
- `make install-fabricx-tools FAB_BINS=<tmp>/bin` puts all three tools in `bin/fabric-x/` and nothing at the root.
- `fabricx-simple` passes against a `bin/` containing **only** `fabric-x/` — proof the subdirectory is what resolves.
- Against one clean v3.1.4 distribution with both toolchains installed (`bin/configtxgen` → v3.1.4, `bin/fabric-x/configtxgen` → fabric-x): `fabric-stoprestart` 3/3 passed and `fabricx-simple` 1/1 passed from the same `FAB_BINS`.
- `make checks` clean.

### Docs

`docs/dev/development.md` (Fabric-x install, `FAB_BINS`, plus a troubleshooting entry for an already-clobbered `bin/`), `docs/agents/integration-tests.md` (which toolchain a target's platform prefix needs), and `integration/fabricx/atsa/README.md`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,11 @@ FABRIC_X_COMMITTER_VERSION ?= 1.0.4
 # need to install fabric binaries outside of fsc tree for now (due to chaincode packaging issues)
 FABRIC_BINARY_BASE ?= $(PWD)/../fabric
 FAB_BINS ?= $(FABRIC_BINARY_BASE)/bin
+# The fabric-x tools ship a configtxgen and a cryptogen of their own. Installing
+# them into $(FAB_BINS) would overwrite fabric's, so they get their own
+# subdirectory there and both toolchains can be installed at once. Keep in sync
+# with fxconfig.BinSubdir.
+FABRIC_X_BINS ?= $(FAB_BINS)/fabric-x
 
 # integration test options
 GINKGO_TEST_OPTS ?=
@@ -43,10 +48,10 @@ install-linter-tool: ## Install linter in $(go env GOPATH)/bin
 	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $$(go env GOPATH)/bin v2.12.2
 
 .PHONY: install-fabricx-tools
-install-fabricx-tools: ## Install fxconfig in $(FAB_BINS)
-	@env GOBIN=$(FAB_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/fxconfig@$(FABRIC_X_TOOLS_VERSION)
-	@env GOBIN=$(FAB_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/configtxgen@$(FABRIC_X_TOOLS_VERSION)
-	@env GOBIN=$(FAB_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/cryptogen@$(FABRIC_X_TOOLS_VERSION)
+install-fabricx-tools: ## Install the fabric-x tools in $(FABRIC_X_BINS)
+	@env GOBIN=$(FABRIC_X_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/fxconfig@$(FABRIC_X_TOOLS_VERSION)
+	@env GOBIN=$(FABRIC_X_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/configtxgen@$(FABRIC_X_TOOLS_VERSION)
+	@env GOBIN=$(FABRIC_X_BINS) go install $(GO_FLAGS) github.com/hyperledger/fabric-x/tools/cryptogen@$(FABRIC_X_TOOLS_VERSION)
 
 .PHONY: install-fabric-bins
 install-fabric-bins: ## Install fabric binaries in $(FABRIC_BINARY_BASE)
--- a/docs/agents/integration-tests.md
+++ b/docs/agents/integration-tests.md
@@ -91,5 +91,11 @@ make integration-tests                                  # all targets
 GINKGO_TEST_OPTS="--focus='IOU Life Cycle'" make integration-tests-fabric-iou
 ```
 
+A target's platform prefix decides which toolchain it needs: `fabric-*` targets
+need `make install-fabric-bins` (binaries in `$FAB_BINS`), `fabricx-*` targets
+need `make install-fabricx-tools` (tools in `$FAB_BINS/fabric-x`, kept apart
+because both toolchains ship a `configtxgen`). Installing both is safe, and
+`fsc-*` targets need neither.
+
 Prerequisites (Fabric binaries, Docker images) are covered in
 [`docs/dev/development.md`](../dev/development.md).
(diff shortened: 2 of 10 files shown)


adecaro (commented):
## Review: does this address #1723?

Checked out at `b6c9b10eea39062b5225afd5364f771169e7c0bf`, up to date with `main`.

```
$ go build ./nwo/... && go vet ./nwo/... && go test ./nwo/fabric/network/...
ok  	github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fabric/network	1.4s
```

```
$ golangci-lint run ./nwo/fabric/network/... ./nwo/fabricx/...
0 issues.
```

Revert-test: reverting the 5 non-test prod files back to base and re-running the new
`TestFindCmdAtEnv` fails to *build* (`findCmdAtEnv`/`BinSubdir` undefined) — confirms the test
is exercising the new code path, not passing vacuously.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Fabric-x tools install into `$FAB_BINS/fabric-x` instead of clobbering `$FAB_BINS` directly | Read `Makefile` diff (`FABRIC_X_BINS := $(FAB_BINS)/fabric-x`, `install-fabricx-tools` now uses `GOBIN=$(FABRIC_X_BINS)`) and `fxconfig.BinSubdir` constant kept in sync |
| fabricx platform reads binaries from that subdirectory | `fabricx/network/network.go` sets `fabricNetwork.BinSubdir = fxconfig.BinSubdir`; every fabric lookup site (`ConfigTxGen`, `Discover`, `Osnadmin`, `OrdererRunner`, `peerCommand`, `Idemixgen`) converted to the new `(n *Network)` receiver methods — grepped for leftover package-level calls, found none |
| No silent cross-toolchain fallback | New `TestFindCmdAtEnv` explicitly asserts a network with `BinSubdir` set does not find a binary that only exists at the top level of `$FAB_BINS` |
| Both install targets can now be run together | `install-fabric-bins` targets `$(FABRIC_BINARY_BASE)` and is untouched by this change — confirmed by reading the Makefile, the two targets no longer share a write path |

### Non-blocking

None.

### Recommendation

Clean, minimally-scoped fix that matches the issue's own implementation plan; all call sites
consistently migrated, new test covers the actual invariant (no fallback), docs updated
including a troubleshooting section for checkouts already in the broken state. CI is fully
green. No blocking or should-fix findings — happy to see this merged as-is.

adecaro (approved):
LGTM

PR #1727, opened 2026-09-05
test(ci): split unit tests per module and stop rebuilding what CI alr…

### LINKED ISSUE #1726: CI repeats work it already has, and unit tests cannot be scoped to one module
## Unit tests are all-or-nothing

`make unit-tests` is a single recipe spanning three modules (root, `integration/`, the libp2p comm host). A change to one of them costs a run over all 208 root packages plus the other two. There is no way to say "just the module I touched", and the `rc=0; … || rc=1` accumulator inside one recipe means the three failures arrive interleaved in one log with no target names attached.

In CI the same recipe runs as one job, so the three modules are serialised on one runner even though they share nothing.

## Coverage from two of the three modules is silently discarded

`go test -C <module>` chdirs before resolving `-coverprofile`, so the integration and libp2p profiles are written *inside* their module directories. The `utest` job uploads `./coverage.profile` — the root module's. The libp2p comm host has been tested in CI for a while with none of its coverage counted.

## Every fabric CI job rebuilds chaincode images from scratch

`make chaincode-images` is gated on `startsWith(matrix.tests, 'fabric-')`, so all 10 `fabric-*` jobs build all 4 images: 40 builds where 11 would do. `fabric-atsa` builds four and uses none (it packages from source).

Worse than the redundancy is that each build is cold. The shared Dockerfile compiles in a `golang:1.26-bookworm` stage, so every job re-pulls that image and re-downloads the module cache into an empty `--mount=type=cache`, while the runner's own Go cache — restored by `actions/setup-go` moments earlier — sits unused next to it. The final stage is `FROM scratch` plus one static binary, so the builder stage contributes nothing to the artefact.

## Every fabric CI job pulls 880MB of images one suite needs

`pull-images-fabric` fetches `fabric-ccenv` (725MB) and `fabric-baseos` (155MB). Both serve only the legacy source-packaging path — peers compile the chaincode in the first and run the result in the second, per `chaincode.builder` and `chaincode.golang.runtime` in `core_template.go`. Only `integration/fabric/atsa` calls `WithLegacyChaincode`; the other nine fabric jobs pull both for nothing.

Related: `NewPlatform` adds only `ccenv` to `RequiredImages` for that path, not `baseos`. `make pull-images-fabric` always fetches both so it does not bite in practice, but a machine with one and not the other gets an opaque peer-side chaincode-launch failure instead of nwo's "Require the following container images" error.

mbrandenburger:
Closes #1726 

This PR improves the make targets for unit testing and improve fabric chaincode container builds.

## Unit tests: one target per module group

```
make unit-tests               # all three, via $(MAKE) -k
make unit-tests-root          # the framework
make unit-tests-integration   # the nwo harness and the helpers the suites share
make unit-tests-extensions    # optional driver modules (today: the libp2p comm host)
```

`unit-tests` recurses with `-k` rather than listing the three as prerequisites, so one module's failure still does not hide the others — the behaviour the old `rc=0; … || rc=1` accumulator provided — while each failure now arrives labelled with its target.

Extension modules are a list (`EXTENSION_MODULES`), not a target apiece: adding a second optional driver is one word and needs no CI change. `TEST_PKGS` now scopes `unit-tests-root`, which is the module it was ever able to scope.

All five unit-test targets are consistent: `@env $(UNIT_TEST_ENV) go test …`. `unit-tests-postgres` had inlined its own `FABRIC_LOGGING_SPEC`, and `unit-tests-sdk` set no environment at all, so it logged at default verbosity.

## CI: a four-entry matrix instead of two duplicated jobs

`utest` becomes a matrix over root / postgres / integration / extensions, each on its own runner, replacing ~80 lines of near-duplicate YAML with one set of steps — matching the `itest` matrix convention already in the file.

Two entries deserve their comments:

- **integration reports no coverage.** `filter-coverage.sh` drops every path under `/integration/` by design, so a profile from that suite filters down to just its `mode:` line. Uploading it would add an empty flag to the parallel Coveralls build, reading as 0% rather than as absent.
- **extensions reports coverage that was previously discarded.** `go test -C` chdirs before resolving `-coverprofile`, so the profile lands inside the module; the entry uploads that path under its own flag. `$(GO_COVERPKG)` is applied in the extensions loop so each module credits its own packages the way root does (which moves libp2p from 52.4% package-scoped to 39.1% module-scoped — same code, honest denominator).

Only the integration module reads `FAB_BINS`, so `install-fabric-bins` is scoped to that entry instead of gating every unit-test run.

## Chaincode images: compile on the host, package with Docker

The image is `FROM scratch` plus one static binary, so the `golang:` builder stage bought each CI job a cold Go environment and nothing else. `build-image.sh` now compiles with the host toolchain — warm via `actions/setup-go` on CI, the developer's own cache locally — and Docker only wraps the binary.

Measured: **~4s for all four images** against ~3min of cold container per job. The build context shrinks from the whole repo to a temp dir holding one binary, so the `COPY . .` layer is gone too.

`--platform linux/$(go env GOARCH)` is new: the builder stage used to fix the image's architecture implicitly by pulling `golang:` for the host platform, and nothing in a scratch image pins it otherwise. The per-chaincode `Dockerfile` escape hatch is unchanged, still repo-root context.

`make chaincode-images` is unchanged for local use.

## Fabric images: pull ccenv and baseos for the one suite that needs them

Both serve only the legacy source-packaging path, so CI pulls them for
`fabric-atsa` rather than all ten fabric targets — 880MB × 9 jobs.

`requiredImagesFor` (extracted from `NewPlatform`) now requires **both** images for that path. It listed only `ccenv`, so a missing `baseos` surfaced as an opaque chaincode-launch failure rather than nwo's actionable error — which is also what makes the CI gating safe: a suite that starts packaging from source and is not listed fails with the missing image named.

## Verification

- `make checks`: 0 issues across all four modules.
- `make unit-tests-root` (scoped), `unit-tests-integration`, `unit-tests-extensions`, `unit-tests-sdk`: pass. `unit-tests-postgres` dry-run only (needs the DB image); its only change is the env line.
- `-k` aggregation: `make unit-tests-extensions EXTENSION_MODULES="nope1 nope2"` attempts both and exits non-zero.
- `make chaincode-images` for real: four single-layer `linux/arm64` images, 16.7–22.9MB, and `docker run --rm fsc-cc/base:latest` reaches the chaincode's own `ccid must be specified` — so nothing was missing from the empty base.
- The coverage claims were checked against `filter-coverage.sh` with synthesised profiles: an integration profile filters to its `mode:` line; a libp2p one keeps production paths and drops `/mock/`.
- `TestRequiredImagesFor` covers legacy / CCaaS-only / empty topologies.

Not run: `make integration-tests-fabric-atsa`, the one suite whose behaviour the last section touches. Worth a look in CI.

## Docs

`docs/dev/development.md`, `docs/agents/testing.md` and `AGENTS.md` cover the new targets. `docs/agents/integration-tests.md` said the legacy path needs "the `ccenv` image" — now both, with what each does.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,43 +28,51 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # One entry per `make unit-tests` module, on its own runner.
   utest:
+    name: utest-${{ matrix.suite }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - suite: root
+            target: unit-tests-root
+            coverage: true
+          - suite: postgres
+            target: unit-tests-postgres
+            coverage: true
+            database: true
+          # No coverage: filter-coverage.sh drops every path under /integration/.
+          - suite: integration
+            target: unit-tests-integration
+            fabric: true
+          - suite: extensions
+            target: unit-tests-extensions
+            coverage: true
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # ratchet:actions/checkout@v7.0.1
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # ratchet:actions/setup-go@v7.0.0
         with:
           go-version-file: "go.mod"
           cache-dependency-path: "**/*.sum"
-      - run: make install-fabric-bins
-      - name: unit-tests
-        run: |
-          GO_TEST_PARAMS="-race -coverprofile=coverage.profile" make unit-tests
-          ./scripts/filter-coverage.sh coverage.profile coverage.profile
-      - uses: coverallsapp/github-action@8d6379e14d29928660c4ba802d8e85393440b329 # ratchet:coverallsapp/github-action@v2.3.8
-        with:
-          file: coverage.profile
-          flag-name: utest
-          parallel: true
-          fail-on-error: false
 
-  utest-postgres:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # ratchet:actions/checkout@v7.0.1
-      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # ratchet:actions/setup-go@v7.0.0
-        with:
-          go-version-file: "go.mod"
-          cache-dependency-path: "**/*.sum"
-      - run: make pull-images-database
-      - name: unit-tests-postgres
-        run: |
-          GO_TEST_PARAMS="-race -coverprofile=coverage.profile" make unit-tests-postgres
-          ./scripts/filter-coverage.sh coverage.profile coverage.profile
+      - name: Set up the database
+        if: matrix.database
+        run: make pull-images-database
+
+      - name: Set up Fabric
+        if: matrix.fabric
+        run: make install-fabric-bins
+
+      - name: Run the unit tests
+        run: GO_TEST_PARAMS=-race COVERAGE=${{ matrix.coverage }} make ${{ matrix.target }}
+
       - uses: coverallsapp/github-action@8d6379e14d29928660c4ba802d8e85393440b329 # ratchet:coverallsapp/github-action@v2.3.8
+        if: matrix.coverage
         with:
           file: coverage.profile
-          flag-name: utest-postgres
+          flag-name: utest-${{ matrix.suite }}
           parallel: true
           fail-on-error: false
 
@@ -101,7 +109,12 @@ jobs:
       - name: Set up Fabric
         if: startsWith(matrix.tests, 'fabric-')
         run: |
-          make -j4 install-fabric-bins pull-images-fabric pull-images-database
+          make -j4 install-fabric-bins pull-images-database
+
+      - name: Set up the legacy chaincode builder
+        # pull-images-fabric serve only the legacy source-packaging which is used it fabric-atsa
+        if: matrix.tests == 'fabric-atsa'
+        run: make -j4 pull-images-fabric
 
       - name: Set up Fabric-x
         if: startsWith(matrix.tests, 'fabricx-')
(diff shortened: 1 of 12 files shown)


adecaro (approved):
LGTM

adecaro (commented):
## Review: does this address #1726?

Checked out at `623bc2e155e8ca73a10ef7d0fb8a35a5226fe9c2`, up to date with main.

```
$ make unit-tests-extensions COVERAGE=1
ok  .../platform/view/services/comm/host/libp2p  74.108s  coverage: 39.1% of statements in ./...
Filtered coverage written to .../coverage.profile
```
Matches the PR body's claimed 39.1% module-scoped figure exactly.

```
$ make unit-tests-extensions EXTENSION_MODULES="nope1 nope2"
go: chdir nope1: no such file or directory
go: chdir nope2: no such file or directory
make: *** [unit-tests-extensions] Error 1
```
Confirms the `-k`-based aggregation attempts every module and still exits non-zero on failure, as claimed.

```
$ make chaincode-images   # real docker build, warm go/module caches
... 4 images built in ~13s total
$ docker run --rm fsc-cc/atsachaincode:latest
panic: Error starting asset chaincode: ccid must be specified
```
Matches the PR's own verification ("nothing missing from the empty base").

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Per-module unit-test targets (`unit-tests-root/-integration/-extensions`), `-k` aggregation | Read `Makefile`, ran the failure-path repro above |
| `utest` CI matrix (root/postgres/integration/extensions) replaces two duplicated jobs | Read `.github/workflows/tests.yml`; all matrix jobs green in the run |
| libp2p coverage no longer silently discarded, now `-coverpkg` scoped per module | Ran `unit-tests-extensions COVERAGE=1` locally, got 39.1% as claimed |
| Chaincode images compiled with host toolchain instead of a cold Docker builder stage | Built all 4 images for real (~13s vs. the claimed ~3min cold); `docker run` sanity-checked one |
| `requiredImagesFor` now requires both `ccenv` and `baseos` for the legacy chaincode path; only `fabric-atsa` pulls them in CI | Read `platform.go`, ran `TestRequiredImagesFor`, confirmed `itest (fabric-atsa)` green in CI |
| Docs updated to match (`docs/dev/development.md`, `docs/agents/testing.md`, `AGENTS.md`, `docs/agents/integration-tests.md`) | Read each against the Makefile/code it describes — consistent |

One thing worth naming explicitly (the PR body already does): this does **not** reduce the *count* of redundant chaincode-image builds from #1726 (`fabric-atsa` still builds all 4 images and uses none) — it makes each build cheap enough (~4s total) that the redundancy stops mattering in practice. That's a reasonable, disclosed tradeoff rather than a gap.

### Blocking

None.

### Non-blocking

- `.github/workflows/tests.yml`, comment on the "Set up the legacy chaincode builder" step: "pull-images-fabric serve only the legacy source-packaging which is used it fabric-atsa" has two grammar slips (serve→serves, used it→used in). Cosmetic only.

### Recommendation

Merge as-is — verified builds, tests, lint, and the specific numeric/behavioral claims in the PR body all check out.

Issue #719, opened 2024-12-19
Decoupling the communication layer from the account identities


arner:
# Intro

The communication layer (platform/view/services/comm) is currently coupled with node identities (through the resolvers), and account identities are also coupled with node identities (through the endpointservice). This is a bit hard to understand and maybe is more coupling than necessary; we might be more flexible without it.

The communication layer needs to be able to have mutually authenticated connections to other nodes. This implies:
- A list of node identifiers
- a way to resolve them so we can route to them over the internet
- Some form of PKI?

A simple implementation could just DNS and TLS with client authentication. Like with Fabric, we could explicitly trust specific CAs or specific (self-signed?) certificates or even just public keys. Right now, for websockets we have a routing-config.yaml with hostnames next to the resolvers configuration which also has addresses, and I suspect it has something to do with the mixing between node identity and transport. The application layer should be able to trust that the connection to another node is secure and authenticated. Also consider the possibility that in the future we might need a more dynamic way to contact new parties without updating core.yaml.

Are we currently using the node identity for anything else than authentication of a connection?

## EndpointService bindings (account -> node)

Consider:
- Node A
  - account_alice
- Node B
  - account_bob

Alice wants to transfer 1 token to bob. If Node A doesn't have account_bob configured in its core.yaml, it has to create a binding before being able to request an ephemeral identity for account_bob:

```go
node := view.Identity("Node B")
rec := view.Identity("account_bob")
eps := viewregistry.GetEndpointService(ctx)
eps.Bind(node, rec)
recipient, err := ttx.RequestRecipientIdentity(ctx, rec)
```
 
RequestRecipientIdentity will then use the just created binding to look up Node B from account_bob, and put the bytes of the literal "account_bob" in the request so that Node B can look up the wallet / identity for account_bob. The address of Node B (as well as the node's public key) to set up the connection comes from the resolvers configuration. This is a bit cumbersome and there is a risk of bugs as well;  for instance if we create a wrong binding "Node A" -> "account_bob" this binding can be used by another thread.

In many usage scenarios, either Node A or Alice herself will already know that Bob is hosted by Node B. It would be more natural if we can just call RequestRecipientIdentity with the node id and the account alias.There is no technical need anymore for tight coupling between node identities and accounts, this can be left to the application.

```go
recipient, err := ttx.RequestRecipientIdentity(ctx, node, rec)
```

## Websockets / https

[...]

PR #1728, opened 2026-09-05
refactor(tls)!: resolve all TLS configuration through one path

### LINKED ISSUE #1112: Refactor TLS Configuration

mbrandenburger:
Closes #1112

## What this does

Adds `platform/view/services/tlsconfig`, which owns TLS configuration for the whole project, and moves every TLS-speaking surface onto it.

A `tls:` subtree unmarshals into pointer-typed wire structs, merges child over parent per field, and resolves to a single `grpc.SecureOptions`. Pointers are the point: they keep "absent" distinct from "false", which is what #1111 needed and what `GetBool` could never express.

Surfaces migrated: view gRPC, web, operations/metrics, websocket P2P, Fabric network defaults and per-peer/per-orderer blocks, the Fabric-x query and notification services, the OTLP trace exporter, and the FSCCLI client flags.

**Deleted:** `web.TLS`, `operations.TLS`, `MetricsOptions.TLS` and the Fabric-x `config.TLSConfig` — four duplicate types for one concept. Seven Fabric TLS accessors collapse into `NetworkClientTLS()`. The per-surface `x509` pool building and keypair reading (~120 lines) is gone; one `TLSConfig()` replaces four hand-rolled builders.

The refactor's own measure — no TLS setting is read by string path any more:

```console
# on main: 27 TLS settings read by string path
# on this branch: none
$ grep -rnE 'Get(Bool|Path|String|StringSlice)\("[^"]*tls\.' platform integration \
    --include="*.go" | grep -v _test | wc -l
0
```

## Breaking changes

This is a hard break, deliberately. Accepting both spellings indefinitely would preserve exactly the ambiguity that caused #1111. **Removed keys are a startup error naming their replacement**, so a stale `core.yaml` fails loudly instead of silently serving in the clear.

| Removed | Replacement | Kind |
| --- | --- | --- |
| `fsc.p2p.opts.websocket.tls.serverRootCAs.files` | `…tls.rootCAs.files` | rename |
| `fsc.metrics.prometheus.tls` | `fsc.metrics.clientAuthRequired` | rename |
| metrics served on the `fsc.web` listener | `fsc.metrics.address` | **behaviour** |
| `fabric.<net>.tls.clientAuthRequired` | `…tls.clientAuthEnabled` | rename |
| `fabric.<net>.tls.serverhostoverride` | `…tls.serverNameOverride` | rename |
| `fabric.<net>.tls.rootCert.file`, `.rootCertFile` | `…tls.rootCAs.files` | rename |
| `fabric.<net>.tls.cert`, `.key`, `.clientRootCAs` | removed — nothing read them | dead |
| `ordering.tlsEnabled`, `ordering.tlsClientAuthRequired` | `fabric.<net>.tls.{enabled,clientAuthEnabled}` | rename |
| `{orderers,peers}[].tlsEnabled` / `.tlsDisabled` | `…[].tls.enabled` (`false` for the latter) | rename |
| `{orderers,peers}[].tlsClientSideAuth` | `…[].tls.clientAuthEnabled` | rename |
| `{orderers,peers}[].tlsRootCertFile` | `…[].tls.rootCAs.files` | rename |
| `{orderers,peers}[].serverNameOverride` | `…[].tls.serverNameOverride` | rename |
| `{orderers,peers}[].tlsRootCertBytes` | none — YAML never set it | dead |
| Fabric-x `tls.clientKey`, `.clientCert` *(flat strings)* | `tls.clientKey.file`, `.clientCert.file` | rename |
| Fabric-x `tls.rootCerts` *(flat list)* | `tls.rootCAs.files` | rename |

Two changes worth calling out separately:

- **Metrics get their own listener.** `fsc.metrics.prometheus.tls` never meant transport TLS — it gated whether scraping `/metrics` required a client certificate. Metrics now have their own `address`, `enabled` and `tls`, decoupled from `fsc.web`.
- **`ordering.tls*` shadowing is gone.** That pair overrode the network block for orderer connections alone, so the same two settings had two homes and the narrower silently won. The network block now applies to every connection, orderers included. File-configured CAs *augment* the config-block CAs rather than replacing them.

Cipher suites are unified on the web list (ECDHE + ChaCha20), which **drops the static-RSA suites** gRPC previously accepted. A hardening change, not a silent collapse — flagged here for the release notes.

`docs/configuration.md` is updated for the new shape.

## Incidental fixes

Three pre-existing bugs surfaced while wiring this up:

- `CheckRemovedKeys` matched only intermediate paths, never leaf keys — so nearly every entry in the table above would have passed silently.
- `ConnectionConfig` could not round-trip through JSON: `SecureOptions.VerifyCertificate` is a func field, which `encoding/json` rejects even when nil. The CLI persists this as JSON.
- The FSCCLI collected `--tlsCert` / `--tlsKey` and never read them, so client-auth flags were silently inert.

## Not in scope

Separating the websocket P2P transport keypair from `fsc.identity` is **blocked on #719**: the P2P peer ID is `base64(sha256(marshalPKIX(pubkey)))` of the transport key, so changing the keypair changes the node's identity. This PR delivers resolution, the `rootCAs` rename and strict decoding for that surface; the split needs #719 first.

## Testing

- `make checks` clean in all four modules (golangci-lint + `go fix`).
- `go test ./platform/... -race` — **113 packages, 0 failures**.
- Integration: `fabricx-*` **4/4** green; `fabric-*` **9 of 9** green; `fsc/pingpong` 22/22, `fsc/stoprestart` 4/4.

`make tidy` was needed: routing the OTLP exporter through `tlsconfig` gives the libp2p host module a transitive `otelgrpc` dependency, which its own `go.sum` did not carry. Worth noting for reviewers that a root-module build cannot see this class of breakage — only `make checks` visits all five modules.


platform/view/services/tlsconfig/validate.go
```diff
@@ -0,0 +1,81 @@
...
+// rather than translated, so a configuration that no longer means what it says fails at
+// startup instead of silently weakening transport security.
+//
+// Only keys under prefix are considered; pass "fsc" to check this node's listeners, or a
+// network prefix to check one Fabric network. It returns nil when no removed key is present.
+func CheckRemovedKeys(src Source, prefix string) error {
+	var errs []error
+	lower := strings.ToLower(prefix)
+	for gone, replacement := range removed {
+		if !strings.HasPrefix(gone, lower) {
```
adecaro, line 74: **Blocking: this prefix filter is inverted, so every Fabric-network-scoped removed key in `removed.go` can never be detected.**

`removed.go` documents phase-3 entries (`tls.clientauthrequired`, `ordering.tlsenabled`, `ordering.tlsclientauthrequired`, etc.) as relative to a network prefix — `CheckRemovedKeys` is meant to be called with `"fabric.<network>."` and match e.g. `"fabric.mynetwork.ordering.tlsenabled"`. But this loop checks `strings.HasPrefix(gone, lower)`, i.e. whether the *removed key itself* starts with the prefix, then calls `src.IsSet(gone)` unprefixed. `strings.HasPrefix("ordering.tlsenabled", "fabric.mynetwork.")` is always `false`, so this entry (and every other phase-3 entry) is skipped unconditionally for any real call from `platform/fabric/core/generic/config/service.go:87`.

Concrete failure: a config with `fabric.mynetwork.ordering.tlsEnabled: false` gets no removed-key error. Nothing else reads `ordering.tlsEnabled` anymore (only `ordering.numRetries`/`ordering.retryInterval` are read, per `service.go`), so the setting silently vanishes instead of failing at startup — the opposite of this function's own stated purpose ("Removed keys are rejected rather than translated... fails at startup instead of silently weakening transport security", lines 61-63).

`TestCheckRemovedKeysRespectsPrefix` doesn't catch this: it only checks that a `fabric`-scoped call ignores an `fsc`-prefixed key, never that it actually catches a phase-3 key under a real network prefix.

Fix: check `src.IsSet(prefix + gone)` (prepending, not filtering by prefix-of), or store phase-3 entries fully-qualified like phase 1/2. Either way, add a test with a fake `Source` that sets `fabric.mynetwork.ordering.tlsenabled` and asserts `CheckRemovedKeys(src, "fabric.mynetwork.")` returns the removed-key error.

adecaro (commented):
## Review: does this address #1112?

Checked out at `c7b1e43b432f4c9e1a9c7fea11f50b8e8fd69d6d`.

```
go build ./... && go test -count=1 ./platform/view/services/tlsconfig/...
ok  	github.com/hyperledger-labs/fabric-smart-client/platform/view/services/tlsconfig	1.026s
```

```
golangci-lint run ./platform/view/services/tlsconfig/... ./platform/view/services/grpc/... ./platform/view/sdk/dig/...
0 issues.
```

CI is fully green across build/vet/lint/unit/postgres/integration jobs, and DCO/GPG/issue-link bot checks all pass.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| One resolution path for TLS across gRPC/Web/websocket-P2P/Fabric/Fabric-x | Read `tlsconfig.Resolve*` call sites in `sdk.go`, `servers.go`, `websocket/config.go`, `platform/fabric/core/generic/config/service.go`; all decode through the same `ServerTLS`/`ClientTLS`/`TLS` structs |
| Multi-CA support (pasquale95's ask in #1112) | `RootCAs`/`ClientRootCAs` are `*Files` (list), confirmed in `types.go` |
| Fail-closed on legacy/removed config keys | Partially true — see Blocking below |
| `docs/configuration.md` matches the new model | Spot-checked `fsc.tls`, `fsc.grpc.tls`, `fsc.metrics`, `fabric.<network>.tls`/orderers/peers sections against the code |

### Blocking

`CheckRemovedKeys` (`platform/view/services/tlsconfig/validate.go:66-81`) filters its migration table by checking whether the *removed key itself* has the caller's prefix (`strings.HasPrefix(gone, lower)`), instead of prepending the prefix to the removed key before checking `IsSet`. The phase-3 table entries in `removed.go` (`tls.clientauthrequired`, `ordering.tlsenabled`, `ordering.tlsclientauthrequired`, etc.) are documented as *relative* to a network prefix like `fabric.mynetwork.`, but since `"ordering.tlsenabled"` never has `"fabric.mynetwork."` as a string prefix, the check for every Fabric-network-scoped removed key is a no-op regardless of what the config actually contains. `ordering.tlsEnabled`/`ordering.tlsClientAuthRequired` specifically are also not read anywhere else in the code anymore (only `ordering.numRetries`/`ordering.retryInterval` are), so an old-style `fabric.mynetwork.ordering.tlsEnabled: false` now silently has zero effect instead of failing at startup — the exact "silently weakening transport security" outcome this mechanism's own doc comment says it exists to prevent. Full detail and suggested fix in the inline comment.

### Non-blocking

- `platform/view/services/tlsconfig/types.go:11`'s package Godoc references `docs/superpowers/specs/2026-09-03-tls-configuration-refactor-design.md`, which doesn't exist in the repo — looks like a dangling pointer to a local design note that wasn't included in the PR.

### Recommendation

Needs another look after the `CheckRemovedKeys` prefix bug is fixed (with a test that actually exercises a Fabric-network-scoped removed key) — everything else in this refactor looks solid and matches the design the issue thread converged on.

platform/view/services/tlsconfig/validate.go
mbrandenburger: Thanks @adecaro good catch! 

The prefix was applied as a filter, but the network entries are relative, so no relative key could match `fabric.<net>.` and every one was skipped. There are now two calls — `CheckRemovedKeys` filters absolute node keys, `CheckRemovedNetworkKeys` prepends the network prefix — and the Fabric call site uses the latter. Neither suggested fix worked as written: prefix + gone double-prefixes the absolute `fsc.*` keys, and the network entries can't be fully qualified since the name is only known at runtime. Three tests cover it, each verified by reintroducing the bug it guards: your case for both prefix shapes, a call-site test (the defect was a wrong function, invisible to package-level tests), and one guarding the two tables against being merged. The table also shrank from eight entries to three — keys inside a `tls:` block are already rejected by the strict subtree decode — with the full mapping now in `docs/configuration.md`.

The godocs are cleaned up as well: the local design-document references are gone, along with two development-phase references the sweep turned up, and the explanation they were standing in for moved into the package doc itself.

adecaro (approved):
LGTM

Issue #1741, opened 2026-09-07
OpenSSF scorecards TODOs


mbrandenburger:
### Description

The project has never been assessed against the [OpenSSF Best Practices passing criteria](https://www.bestpractices.dev/en/criteria/0), and is not registered on bestpractices.dev. An audit shows we already satisfy most of the MUST criteria through things we have — Apache-2.0 with `LICENSE`, `CONTRIBUTING.md`, `MAINTAINERS.md`, issue templates, semver tags with generated release notes, CodeQL and golangci-lint on every PR, `-race` unit tests, Coveralls, SHA-pinned actions, a DCO gate, secret scanning and Dependabot security updates.

The gaps are narrow, and this issue tracks closing them so the badge can be claimed:

- **`test_policy` is the one outright MUST failure.** The criteria require a stated policy that new functionality arrives with tests. We follow it and CI reports coverage per PR, but nothing writes it down.
- **`vulnerability_report_process` passes only weakly.** `SECURITY.md` is generic Hyperledger boilerplate: retired wiki link, email only, and no mention that GitHub private vulnerability reporting is already enabled on this repository.
- **`documentation_interface` passes only weakly.** The Go API is our main external interface and is linked from nowhere.
- Two README badges are dead: Go Report Card is retired, and the tests badge links to a `workflows/go.yml` that does not exist.
- No `dependabot.yml`, so nothing bumps dependencies on a schedule and the SHA-pinned actions never move.
- No dependency-vulnerability scanning of our own code paths (`govulncheck`) and no security linter (`gosec`).

### Implementation Steps

Documentation and repository hygiene:

- [x] Rewrite `SECURITY.md`, document the test policy, link the Go API reference, and drop the dead badges #1742
- [x] Remove the committed `.env` and ignore the path #1743
- [x] Add `.github/dependabot.yml` covering all five modules, with grouped updates #1744
- [x] Add OpenSSF scorecard badge / practices badge #1788

CI and tooling:

- [x] Add `govulncheck` across the modules, on a schedule and on PRs #1754
- [x] Enable `gosec` in `.golangci.yml` (also serves the SUGGESTED "maximally strict warnings" criterion) #1757
- [x] Run the existing fuzz target on a schedule with `-fuzz`, rather than only its seed
      corpus (SUGGESTED `dynamic_analysis`) #1794 
- [x] Set workflow privileges to minimum #1792 

Repository settings, maintainer action:

[...]

PR #1743, opened 2026-09-07
chore: drop the committed .env and ignore it

mbrandenburger:
Removes the committed `.env` and ignores the path.

The file set `GINKGO_TEST_OPTS`, but nothing ever read it: the Makefile defines those options itself (`Makefile:24-25`) and includes only `checks.mk`, and no script, workflow, or test references `.env`. It was a local override that got committed and then went stale.

A committed `.env` is also exactly what secret scanners and the OpenSSF Best Practices review flag, even when — as here — it holds nothing sensitive.

Part of #1741

### How it was verified

Confirmed unreferenced by grepping the Makefile, `checks.mk`, `scripts/`, `ci/`, `.github/workflows/`, and the Go tree, so deleting it changes nothing at build or test time. `git check-ignore -v .env` confirms the new ignore rule takes effect.

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,0 @@
-GINKGO_TEST_OPTS ?=
-#GINKGO_TEST_OPTS += --keep-going
-GINKGO_TEST_OPTS += --vv
-#GINKGO_TEST_OPTS += --until-it-fails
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

PR #1744, opened 2026-09-07
build(deps): let dependabot update every module

mbrandenburger:
Adds `.github/dependabot.yml`. There was none, so nothing bumped dependencies on a schedule: version updates were done by hand, and the SHA-pinned actions never moved at all. Only Dependabot *security* updates were enabled, and those react to advisories rather than keeping the tree current.

- **One `gomod` entry covering all five modules**, via a `directories` list. The modules are not independent — `integration` replaces the root, `cc/query`, and `libp2p` — so a shared dependency has to move in several `go.sum` files at once, and separate per-module PRs would each land with `make tidy` dirty.
- **Minor and patch grouped** into one PR per ecosystem. Majors deliberately match no group, so each breaking bump arrives alone. Security updates stay ungrouped — those are the PRs worth merging immediately.
- **`github-actions` covered too**, since the SHA pins currently never move.

Part of #1741

### How it was verified

YAML parses, and the directory list was diffed against `git ls-files` to confirm it is exactly the tracked `go.mod` directories: `make list-go-modules` plus `tools/`, which that target skips only because golangci-lint has nothing to analyse there.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,62 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+---
+version: 2
+
+updates:
+  # FSC is a multi-module repository, and the modules are not independent: the
+  # integration module `replace`s the root, cc/query, and libp2p modules, so a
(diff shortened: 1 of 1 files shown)


adecaro (approved):
LGTM

PR #1746, opened 2026-09-07
fix(comm): stop reusing the JSON decode target in readIncoming

### LINKED ISSUE #1745: Websocket multiplexer corrupts in-flight messages when built with Go 1.27
### What happened?

Both multiplexed `readIncoming` loops in `platform/view/services/comm/host/websocket/ws/multiplexed_provider.go` reuse a single `MultiplexedMessage` as the JSON decode target for every frame, and then hand `mm.Msg` — a `[]byte` that aliases that struct's field — to a consumer goroutine over a channel:

| Site | Function |
|------|----------|
| `multiplexed_provider.go:264` | `(*multiplexedClientConn).readIncoming` |
| `multiplexed_provider.go:321` | `(*multiplexedServerConn).readIncoming` |

Go 1.26's `encoding/json` allocated a fresh slice for every `[]byte` field, so a payload a consumer was holding could never be written again. Go 1.27 routes `[]byte` through `base64.(*Encoding).AppendDecode`, which **reuses a non-nil destination's backing array**. The next `ReadJSON` therefore writes over bytes a consumer is still reading, whenever the incoming payload fits within the previous allocation's capacity.

- **Expected:** a payload delivered to a sub-connection stays intact until its reader is done with it.
- **Observed:** it is overwritten in place by the next frame on the same connection.

This is silent message corruption on the wire, not only the data race the detector reports. `subConn.deliver` returns as soon as the value is queued on `receiverChan`, so the window between hand-off and `(*stream).Read` consuming the bytes is wide.

### How can we reproduce it?

**1. The mechanism, independent of FSC.** `go run` this against each toolchain:

```go
package main

import (
	"encoding/json"
	"fmt"
)

type M struct {
	Msg []byte `json:"msg"`
}

func main() {
	var reused M
	_ = json.Unmarshal([]byte(`{"msg":"QUFBQUFBQUFBQQ=="}`), &reused) // "AAAAAAAAAA"
	held := reused.Msg
	_ = json.Unmarshal([]byte(`{"msg":"QkJCQkJCQkJCQg=="}`), &reused) // "BBBBBBBBBB"
	fmt.Printf("%q\n", held)
}
```

| Toolchain | `held` after the second decode | |
|-----------|-------------------------------|---|
| `go1.26.5` | `"AAAAAAAAAA"` | intact |
| `go1.27.1` | `"BBBBBBBBBB"` | **corrupted** |

**2. In the tree.** With a Go 1.27 toolchain on `PATH`, from the repository root:

```bash
go test -race -count=1 ./platform/view/services/comm/host/websocket/...
```

Three tests fail, each reporting a write from `base64.decodeQuantum` (reached via `ReadJSON` in `readIncoming`) racing a read from `(*stream).Read`:

- `TestP2PLayerTestRound` and `TestSessionsMultipleMessagesTestRound` in `.../websocket`
- `TestConnections/*` in `.../websocket/ws`

### Environment

- FSC version (e.g., latest main, v0.9.0): latest main, `03a68004`
- Go Version (e.g., go1.25.1 darwin/arm64): `go1.27.1 darwin/arm64` (compared against `go1.26.5 darwin/arm64`)
- Operating system (e.g., Linux, MacOS): MacOS (darwin 25.5.0, arm64)

### Additional Information

**The `go` directive does not gate this.** A Go 1.27 toolchain corrupts even while `go.mod` still says `go 1.26.5`, so `main` is affected today for anyone whose local toolchain is already 1.27. CI stays quiet only because every workflow resolves its toolchain from `go-version-file: go.mod`; it will start failing the moment that directive is bumped.

**`GODEBUG=jsonv1=1` is not a workaround.** The buffer reuse lives in the shared `[]byte` path rather than in the v1/v2 semantic flag, so the old behaviour cannot be restored through GODEBUG. The code has to stop reusing the decode target.

**Scope.** Only those two loops are affected. `simple_provider.go:95` reads its `StreamMeta` once rather than in a loop, and `large_message_test.go` already declares its `MultiplexedMessage` per iteration.

**Suggested fix.** Declare `mm` inside each loop so every decode starts from a nil slice and `encoding/json` allocates a payload the reader owns. That also stops a field absent from one frame from inheriting the previous frame's value, which the reuse allowed too.

mbrandenburger:
# PR description

Both multiplexed `readIncoming` loops declared a single `MultiplexedMessage` outside the loop and decoded every frame into it, then handed `mm.Msg` — a `[]byte` aliasing that struct's field — to a consumer goroutine over a channel. Go 1.26's `encoding/json` allocated a fresh slice per `[]byte` field, so that was safe by accident; Go 1.27 routes `[]byte` through `base64.(*Encoding).AppendDecode`, which reuses a non-nil destination's backing array, so the next `ReadJSON` overwrites a payload that is still being read. Declaring `mm` per iteration makes each decode start from a nil slice, so `encoding/json` allocates a payload the reader owns outright.

Two consequences beyond the data race:

- a payload handed to a consumer is no longer aliased by the next frame on that connection;
- a field absent from one frame no longer inherits the previous frame's value, which reuse also allowed.

Fixes #1745

## How it was verified

```bash
# fails on main with a Go 1.27 toolchain, passes here
go test -race -count=1 ./platform/view/services/comm/host/websocket/...
#   ok  .../comm/host/websocket      89.429s
#   ok  .../comm/host/websocket/ws   25.236s

make checks       # 0 issues in every module
make unit-tests   # 146 packages ok, no failures, no races
```

The three tests that fail without this change are `TestP2PLayerTestRound`, `TestSessionsMultipleMessagesTestRound` and `TestConnections/*`. They pass on `main` today only because CI resolves its toolchain from `go-version-file: go.mod`, which still selects Go 1.26 — the bug reproduces on unmodified `main` as soon as the toolchain is 1.27, so this is worth merging on its own rather than as part of the Go upgrade.

## Notes for reviewers

The diff is two moved declarations; the reasoning is in the linked issue, which carries a standalone repro that shows the corruption without any FSC code.

Only these two sites had the pattern — `simple_provider.go:95` reads its `StreamMeta` once rather than in a loop, and `large_message_test.go` already declares per iteration.

Out of scope: the Go 1.27.1 upgrade itself, which is a separate PR that depends on this one landing first.

### DIFF
--- a/platform/view/services/comm/host/websocket/ws/multiplexed_provider.go
+++ b/platform/view/services/comm/host/websocket/ws/multiplexed_provider.go
@@ -264,8 +264,11 @@ func (c *multiplexedClientConn) readIncoming() {
 		err := c.Kill()
 		logger.Debugf("Client connection closed: %v", err)
 	}()
-	var mm MultiplexedMessage
 	for {
+		// Declared per iteration on purpose: mm.Msg is handed to another goroutine
+		// below, and encoding/json reuses a non-nil []byte field's backing array,
+		// so reusing mm would let the next read overwrite a payload still in flight.
+		var mm MultiplexedMessage
 		err := c.conn.ReadJSON(&mm)
 		if err != nil {
 			logger.Debugf("Client connection errored: %v", err)
@@ -318,8 +321,9 @@ func (c *multiplexedServerConn) readIncoming(newStreamCallback func(pStream host
 		err := c.Kill()
 		logger.Debugf("Server connection closed: %v", err)
 	}()
-	var mm MultiplexedMessage
 	for {
+		// Declared per iteration on purpose: see the note in the client's readIncoming.
+		var mm MultiplexedMessage
 		_ = c.conn.SetReadDeadline(time.Now().Add(readTimeout))
 		err := c.conn.ReadJSON(&mm)
 		if err != nil {


adecaro (commented):
## Review: does this address #1745?

Checked out at `c800eff9`.

Reverted the fix locally and re-ran the tests named in #1745 under `GOTOOLCHAIN=go1.27.1+auto go test -race`:

```
(reverted) testing.go:1865: race detected during execution of test
            + cascading READY/FINAL_ACK timeouts from the corrupted stream
FAIL	github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/host/websocket	125.362s
```

With the fix restored, the same run under Go 1.27.1 passes cleanly:

```
ok	github.com/hyperledger-labs/fabric-smart-client/platform/view/services/comm/host/websocket	21.859s
```

That reproduces the exact bug described in #1745 and confirms the fix actually resolves it, not just quiets the race detector.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Reused `mm` decode target lets the next frame overwrite a payload a consumer goroutine is still reading, under Go 1.27's `[]byte` decode path | Read both `readIncoming` loops (`multiplexed_provider.go:264`, `:321`); confirmed `mm.Msg` is handed to a channel-consuming goroutine before the next loop iteration decodes into the same `mm` |
| Only these two loops are affected | Checked `simple_provider.go:95` (one-shot read, not in a loop) and `large_message_test.go` (already declares `mm` per iteration); grepped the tree for other loop + decode-target-handoff patterns — none found |
| Fix declares `mm` per-iteration so `encoding/json` always allocates a fresh backing array | Read the diff directly — matches the issue's own suggested fix |

### Blocking

None.

### Non-blocking

None — this is a minimal, correctly-scoped fix with accurate explanatory comments. Build, `gofmt`, `golangci-lint`, and the full `websocket`/`websocket/ws` test suites (under both go1.26.5 and go1.27.1) are clean, and CI is green.

### Recommendation

Merge as-is — the fix is minimal, correctly scoped, and independently verified to resolve the reported corruption.

adecaro (approved):
LGTM

PR #1751, opened 2026-09-07
chore(deps): upgrade every module to Go 1.27.1

### LINKED ISSUE #1750: Upgrade to Go 1.27.1
### Description

All five modules currently declare `go 1.26.5`. This task moves them to `go 1.27.1` and refreshes the `golang.org/x` libraries that track the release cycle, using `scripts/gomate.sh` so every module is updated and tidied together rather than one at a time.

The bump does not stand on its own. Go 1.27 changes enough that three pieces of tooling and a handful of source files have to move with it, and each of these is enforced by a CI job today:

- **golangci-lint must be bumped.** The pinned `v2.12.2` is built with Go 1.26 and refuses to analyse a module targeting 1.27 at all: `can't load config: the Go language version (go1.26) used to build golangci-lint is lower than the targeted Go version (1.27.1)`. `v2.13.2` is built with Go 1.27. The version is pinned in two places, `Makefile` and `.github/workflows/golangci-lint.yml`, and both have to agree.
- **`gofumpt`'s configuration key changed.** `v2.13` deprecates `gofumpt.extra-rules` in favour of `gofumpt.extra.group-params` in `.golangci.yml`.
- **`go fix` has new modernizations.** The `gofix` workflow fails while `make go-fix` reports any, so `make go-fix-apply` has to run. Go 1.27 permits promoted fields of an embedded struct in a composite literal, so most of the resulting churn is `go fix` collapsing nested literals; the rest is `sync/atomic` types replacing `atomic.LoadX(&x)` helpers and `slices.Backward` replacing a manual reverse loop. Note that this output does **not** compile under `go 1.26.5`, so the directive bump and `go fix` have to land together.

Two smaller consequences fall out of the newer linter rather than out of Go itself. The updated `staticcheck` reports `SA1019` on `credentials.ProtocolInfo.SecurityVersion` in `platform/view/services/grpc/creds.go`; two tests assert that value and already suppress the warning at the read sites, so it should be suppressed at the write site rather than dropped. And in `integration/nwo/fabric/network/update.go`, `gci` and `gofumpt` disagree about how to indent one multi-value return of composite literals — `gci` formats through the Go 1.27 printer and wants the dedented form, `gofumpt` bundles an older printer and wants the original — so no formatting of that function satisfies both linters and the construct has to be restructured.

CI needs no version change of its own: every workflow resolves its toolchain from `go-version-file: go.mod`.

This task had one prerequisite, #1745, now fixed by #1746 and merged. That bug is independent of the upgrade and was already latent on `main`, but bumping the directive is what makes CI select a Go 1.27 toolchain and start failing on it, so it had to land first for this work to go green.

### Implementation Steps

- [ ] Bump the directive in every module: `./scripts/gomate.sh update go@1.27.1`
- [ ] Refresh the `golang.org/x` libraries: `./scripts/gomate.sh update golang.org/x/<pkg>@latest` for each of `sync`, `text`, `crypto`, `mod`, `net`, `sys`, `tools`, `exp`, `telemetry`, `time`
- [ ] Tidy every module: `./scripts/gomate.sh tidy`
- [ ] Bump `golangci-lint` to `v2.13.2` in `Makefile` and `.github/workflows/golangci-lint.yml`, and reinstall locally with `make install-tools`
- [ ] Migrate `gofumpt.extra-rules` to `gofumpt.extra.group-params` in `.golangci.yml`
- [ ] Apply the modernizations with `make go-fix-apply`
- [ ] Suppress `SA1019` at the `SecurityVersion` write site in `platform/view/services/grpc/creds.go`
- [ ] Restructure the multi-value composite-literal return in `integration/nwo/fabric/network/update.go` so `gci` and `gofumpt` agree
- [ ] Verify with `make checks` and `make unit-tests`

### Additional Information

Was blocked by #1745, fixed by #1746 and merged, so this task is unblocked.

A stale `golangci-lint` cache produces confusing results while working on this, because the binary is replaced partway through: run `golangci-lint cache clean` after installing `v2.13.2`. See [`docs/dev/development.md`](../blob/main/docs/dev/development.md#lint-failures-in-files-you-never-touched).

Editors will show spurious `unknown field X in struct literal` errors on the `go fix` output until `gopls` is itself rebuilt against Go 1.27, since older versions do not know the promoted-field composite literal syntax. The compiler, `go vet` and `golangci-lint` all accept the code.

mbrandenburger:
# PR description

Moves all five modules from `go 1.26.5` to `go 1.27.1` and refreshes the `golang.org/x` libraries, through `scripts/gomate.sh` so every module is updated and tidied together:

```bash
./scripts/gomate.sh update go@1.27.1
./scripts/gomate.sh update golang.org/x/<pkg>@latest
./scripts/gomate.sh tidy
```

`x/crypto`, `x/exp` and `x/telemetry` moved; the rest were already current. No workflow needs a version change of its own, since each resolves its toolchain from `go-version-file: go.mod`.

The rest of the diff is what `make checks` needs in order to pass at 1.27.1, and none of it is separable from the bump:

- **`golangci-lint` `v2.12.2` → `v2.13.2`**, in `Makefile` and the lint workflow. `v2.12.2` is built with Go 1.26 and refuses to analyse a module targeting 1.27 at all: `the Go language version (go1.26) used to build golangci-lint is lower than the targeted Go version (1.27.1)`.
- **`.golangci.yml`**: `gofumpt`'s `extra-rules` is deprecated in `v2.13` in favour of `extra.group-params`.
- **`make go-fix-apply`**, which the `gofix` workflow enforces. Go 1.27 allows promoted fields of an embedded struct in a composite literal, so most of the churn is `go fix` collapsing nested literals; the rest is `sync/atomic` types replacing the `atomic.LoadX(&x)` helpers, and `slices.Backward` replacing a manual reverse loop. This output does not compile under `go 1.26.5`, which is why it cannot be split into its own PR ahead of the directive bump.
- **`platform/view/services/grpc/creds.go`**: the newer `staticcheck` reports `SA1019` on `ProtocolInfo.SecurityVersion`. Two tests assert that value and already suppress the warning at the read sites, so this suppresses it at the write site rather than dropping the field.
- **`integration/nwo/fabric/network/update.go`**: `gci` and `gofumpt` disagree about how to indent a multi-value return of composite literals — `gci` formats through the Go 1.27 printer and wants the dedented form, `gofumpt` bundles an older printer and wants the original — so no formatting of that function satisfies both. Assigning the named results and returning them sidesteps the construct.

Fixes #1750 

## How it was verified

Verified on this branch rebased onto `99799c3b`:

```bash
make checks       # 0 issues in every module, no go fix suggestions
go vet ./...      # clean in every module
make unit-tests   # 146 packages ok, no failures, no data races
./scripts/gomate.sh tidy   # no-op, every module already tidy
```

That covers the `golangci-lint`, `gofix` and `utest-root` / `utest-integration` / `utest-extensions` jobs. `utest-postgres` and the integration suites were not run locally; CI covers those.

The Go 1.26 → 1.27 behaviour change behind #1745 was confirmed against both toolchains directly, rather than inferred from the failing tests: `GOTOOLCHAIN=go1.26.5` preserves a handed-off `[]byte` payload across a second decode into a reused target, `go1.27.1` overwrites it. Before #1746 landed, this branch went red in exactly two packages with 369 race reports, all of them that one bug.

## Notes for reviewers

This bump had one prerequisite, now already on `main`: Go 1.27's `encoding/json` reuses a non-nil `[]byte` field's backing array, which surfaced a pre-existing bug in the websocket multiplexer where one decode target was reused across frames and its payload handed to another goroutine (#1745, fixed by #1746). That bug was latent on `main` for anyone building with a Go 1.27 toolchain; bumping the directive here is what makes CI select 1.27 and would have made it fail. It was deliberately fixed in its own PR so the two changes stayed reviewable apart, and this branch is rebased on top of it — no leftover dependency, nothing to merge in a particular order.

Worth a look first: the `go.mod`/`go.sum` diffs and the five tooling and lint changes listed above. The `go fix` churn across the remaining ~28 files is mechanical and was applied by the tool, not by hand — the one thing to know while reading it is that a composite literal setting the fields of an embedded struct directly is valid Go 1.27, not a mistake.

Editors may flag that output as `unknown field X in struct literal` until `gopls` is rebuilt against Go 1.27; the compiler, `go vet` and `golangci-lint` all accept it.


adecaro (commented):
## Review: does this address #1750?

Checked out at `3727e7c3b7e0fa59fb9e5abf2361f3945e64390b`.

```
$ golangci-lint run ./...   # locally-installed v2.12.2, pre-upgrade
Error: can't load config: the Go language version (go1.26) used to build golangci-lint is lower than the targeted Go version (1.27.1)
```
Reproduces exactly the blocker the issue describes.

```
$ make lint   # after installing the pinned v2.13.2
  -> .
0 issues.
  -> integration
0 issues.
  -> platform/fabric/services/state/cc/query
0 issues.
  -> platform/view/services/comm/host/libp2p
0 issues.
```

```
$ make go-fix   # read-only `go fix -diff`, all 4 modules
Running go fix...
  -> .
  -> integration
  -> platform/fabric/services/state/cc/query
  -> platform/view/services/comm/host/libp2p
```
No residual modernizations in any module — `make go-fix-apply` was run to completion.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| `go` directive → 1.27.1 in all five modules | read every `go.mod`; all five agree, `golang.org/x/crypto`/`exp`/`telemetry` bumped consistently wherever referenced |
| golangci-lint v2.12.2 → v2.13.2 (Makefile + workflow) pinned together | reproduced the old-version failure and the new-version clean run locally |
| `gofumpt.extra-rules` → `extra.group-params` in `.golangci.yml` | matches the v2.13 deprecation the issue cites; lint is clean under the new key |
| `go fix` modernizations applied (promoted-field literals, `atomic.Int32`/`Uint32`, `slices.Backward`) | spot-checked each category against its receiver/embedding type (e.g. `ReadWriteSet` embeds `ReadSet`/`WriteSet`/`MetaWriteSet`); `go vet` raised no `copylocks` issues on the atomic-type conversions; neither converted struct is ever copied by value |
| `SecurityVersion` SA1019 suppressed, not dropped | `creds.go` field and value unchanged, `//nolint:staticcheck` added at the write site |
| `update.go` restructured so gci/gofumpt agree | `computeGroupUpdate` already had named results (`readSet, writeSet *cb.ConfigGroup`); the diff just assigns and returns them, behaviorally identical to the old direct-return literals |

### Blocking

None.

### Non-blocking

None — one unit-test run of the root module produced a bare `FAIL` with no attributable package or panic in the log; an immediate re-run passed cleanly. Looks like a pre-existing flake somewhere in the `comm`/`websocket` suite under load rather than anything this diff introduces, and CI's own `utest-root` run is green, but flagging it here in case it recurs for reviewers.

### Recommendation

Scope-disciplined, every claim in the commit message independently verified, all CI green. Fine to merge as soon as a maintainer's review/approval requirement is satisfied (currently `BLOCKED` only on required reviews, not conflicts or failing checks).

adecaro (approved):
LGTM

Issue #1756, opened 2026-09-07
Unbounded allocation in `varintReader.ReadData` when `fsc.p2p.maxRecvMsgSize` is 0


mbrandenburger:
### What happened?

`varintReader.ReadData` (`platform/view/services/comm/io/reader.go:52-65`) reads a length varint off the wire and allocates `make([]byte, l)` before reading any payload. The length check above it is `if r.maxMessageSize > 0 && l > uint64(r.maxMessageSize)`, so `maxMessageSize == 0` means *no limit*, and a peer can make the node allocate up to 2^64-1 bytes with a 10-byte frame — panic or OOM, remotely triggerable by anyone who can open a P2P stream.

Zero is reachable from configuration. `NewConfig` (`platform/view/services/comm/config.go:57-62`) clamps `maxRecvMsgSize` only when it is `< 0`, so `fsc.p2p.maxRecvMsgSize: 0` in `core.yaml` passes through to `NewVarintProtoReader` (`platform/view/services/comm/p2p.go:312`) unchanged. The sibling buffer sizes in the same function clamp on `<= 0`; the message sizes are inconsistent with them, and no "unlimited" mode is documented ([`docs/configuration.md:115`](docs/configuration.md)).

Expected: a length that cannot be satisfied is rejected before allocating, and no configuration value disables the limit.

Secondary: even with a limit in place, the full claimed length is allocated up front, so a peer can force a `maxRecvMsgSize`-sized allocation per stream with a few bytes of frame.

### How can we reproduce it?

1. Set `fsc.p2p.maxRecvMsgSize: 0` under `fsc.p2p` in a node's `core.yaml`.
2. Open a P2P stream to that node and write a varint length of `0xFFFFFFFFFFFFFFFF` followed by no payload.
3. The node panics with `runtime: out of memory` / `makeslice: len out of range` in `varintReader.ReadData`.

A unit test reproduces it without a network: `newVarintReader(bytes.NewReader([]byte{0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0x01}), 4096, 0).ReadData()`.

### Environment

- FSC version (e.g., latest main, v0.9.0): latest main (a32de69e)
- Go Version (e.g., go1.25.1 darwin/arm64): go1.27.1
- Operating system (e.g., Linux, MacOS): any

### Additional Information

Suggested fix, two parts:

[...]

Issue #1755, opened 2026-09-07
Enhanced golangci checks


mbrandenburger:
### Description

This tasks is about adding more checks to golangci. A recommended list is given in the Implementation Steps Section. 

### Implementation Steps

Default Go correctness checks (enabled by default)
  - [x] errcheck
  - [x] govet
  - [x] ineffassign
  - [x] staticcheck
  - [x] unused

Bug / correctness
  - [x] bodyclose #1782
  - [x] contextcheck #1783
  - [x] durationcheck #1767 
  - [x] errorlint #1780
  - [x] makezero #1769
  - [x] nilerr #1776
  - [x] nilnesserr #1770 
  - [x] recvcheck #1774
  - [x] rowserrcheck #1775

Security
  - [x] gosec #1757

Go API / language usage
  - [x] goheader #1680
  - [x] predeclared #1779 
  - [x] unconvert #1786 
  - [x] usestdlibvars #1777

Logging / observability
  - [x] loggercheck #1768
  - [x] spancheck #1778

Code quality
  - [x] gocritic #1789
  - [x] revive #1793
  - [x] misspell #1247

Tests
  - [x] testifylint #1785 
  - [x] thelper #1231
  - [x] tparallel #1234
  - [x] paralleltest #1234
  - [ ] usetesting #1812

mbrandenburger:
To enable these linters we will create a dedicated PR per linter. Here is the proposal for a merge plan:

1. standard-set alone — prerequisite, normalizes the list. 
https://github.com/hyperledger-labs/fabric-smart-client/pull/1765

2. Four in parallel, zero content risk: durationcheck, loggercheck, makezero, nilnesserr. Config-only, nothing to review beyond one line each.
https://github.com/hyperledger-labs/fabric-smart-client/pull/1770
https://github.com/hyperledger-labs/fabric-smart-client/pull/1769
https://github.com/hyperledger-labs/fabric-smart-client/pull/1768
https://github.com/hyperledger-labs/fabric-smart-client/pull/1767

3. Seven in parallel — I verified these are mutually content-disjoint: recvcheck, rowserrcheck, errorlint, predeclared, usestdlibvars, nilerr, spancheck. Each collides only with gocritic/revive/unconvert/bodyclose/testifylint, all of which come later.
https://github.com/hyperledger-labs/fabric-smart-client/pull/1780
https://github.com/hyperledger-labs/fabric-smart-client/pull/1779
[...]

PR #1757, opened 2026-09-07
chore(lint): enable gosec in golangci-lint

mbrandenburger:
Enables `gosec` and scopes it to the code that ships. `make checks` is green. 

This is part of #1755 and #1741 

### Scoping

A bare enable reports **220 findings**: 95 in `_test.go`, 72 in the `integration/` nwo harness, 1 in `ci/`, 52 in library code. The test and harness findings describe what that code is for — 0644 fixture certs, 0755 scratch dirs, subprocess launches with generated paths, deliberately weak TLS in benchmarks — so they are excluded by path. Four rule classes are excluded globally, each with its reasoning in `.golangci.yml`:

| Rule | Why |
|---|---|
| G304, G703 | file paths come from operator-supplied `core.yaml` / MSP config, not untrusted input; `filepath.Clean` would silence without securing |
| G404 | `math/rand` is retry jitter and peer/orderer load-balancing, never a secret or nonce |
| G115 | flagged conversions are vault version encoding and proto `int32` fields; truncation only past 2^32 blocks |

### Fixes

- **`platform/view/services/grpc/client.go`** — `VerifyPeerCertificate` does not run on a resumed TLS session, so a custom certificate check could be bypassed. Disables client-side session tickets to force a full handshake (G123). The one genuine security finding in the set.
- **`platform/view/services/storage/driver/sql/sqlite/persistence.go`** — database directory was created `0777` (world-writable), now `0750`.
- **`node/start/profile/profile.go`** — profile output directory `0755` → `0750`.

### Annotated false positives

- `platform/fabricx/core/vault/vault.go` — `unsafe.Pointer` comparison is deterministic lock ordering (G103).
- `platform/fabric/core/generic/delivery/delivery.go` — the nil-ctx fallback is documented behaviour, not a dropped request context (G118).

### Verification

`make checks` exits 0 — 0 issues across all four linted modules, `go fix` clean. Unit tests pass for every touched package.

### Follow-up

Enabling gosec surfaced an unrelated remote-OOM in `comm/io/reader.go` (unbounded `make([]byte, l)` from a wire varint when `fsc.p2p.maxRecvMsgSize` is 0). Tracked separately in #1756 — not in this PR.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,6 +7,7 @@ version: "2"
 linters:
   enable:
     - goheader
+    - gosec
     - misspell
     - thelper
     - tparallel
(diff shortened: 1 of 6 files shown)


SaidAltury-ibm (approved):
Thank you @mbrandenburger, LGTM. just link an issue

PR #1762, opened 2026-09-08
ci(govulncheck): run on pull requests

mbrandenburger:
## What

Adds the `pull_request` trigger to `.github/workflows/govulncheck.yml`, the `concurrency` block that cancels superseded pull request runs, and the corresponding row in the *What CI Gates* table in `docs/dev/development.md`.

This is the follow-up promised in #1754, and part of #1741.

## Why

#1754 landed the scan on `push` to `main` and a weekly cron, and deliberately left the `pull_request` trigger out. The reason was a reachable finding with no upstream fix, combined with the fact that code scanning treats every finding as new until a branch baseline exists — so a `pull_request` trigger added there would have failed that PR's own check on something nobody could action.

Both preconditions are now met: the baseline exists across all four categories, and the findings from that first `main` run have been dismissed as *won't fix*.

| Category | Results in baseline |
|---|---|
| `govulncheck-.` | 1 |
| `govulncheck-integration` | 2 |
| `govulncheck-platform/fabric/services/state/cc/query` | 0 |
| `govulncheck-platform/view/services/comm/host/libp2p` | 2 |

A pull request's analysis is diffed against its base, so pre-existing findings are not attributed to the PR. Only a *new* finding — a dependency bump that introduces a reachable advisory, which is exactly what this is for — turns a check red.

## How

`pull_request` mirrors the `push` trigger's `paths-ignore` and `codeql-analysis.yml`'s shape exactly, including `branches: [main]`.

The `concurrency` block is a no-op for the triggers that exist today: `push` and `schedule` fall back to `github.run_id`, which is unique per run, so each gets its own group and is never cancelled by — or queued behind — another run. It only bites on pull requests, where a newer push makes the in-flight scan obsolete.

### The docs row, and why it brought a `make` target

The table is documented as what CI gates **on every pull request**, with a local-equivalent column, so the row is only accurate once this trigger exists — hence it lands here rather than in #1754.

Every other row in that table names a `make` target, so this one does too. `make govulncheck` reuses the existing `in_all_modules` helper in `checks.mk`, which means it covers exactly the four modules the CI matrix does and cannot drift from it — `tools/` is excluded in both, for the same build-tag reason. It is deliberately **not** part of `make checks`: it needs a tool installed and queries the Go vulnerability database over the network, neither of which belongs in the default read-only check pass. `govulncheck` is pinned in `tools/tools.go`, so `make install-tools` provides it and the documented command actually runs.

### One asymmetry worth flagging for review

`govulncheck` exits 0 in `-format sarif` mode even with findings, which is what keeps CI green and leaves the pass/fail decision to code scanning. In **text** mode it does not — it exits 3. The CLI also knows nothing about code scanning dismissals, so dismissing an alert does not make the local command pass.

`make govulncheck` can therefore fail while CI is green — the one row in that table where the local equivalent and CI legitimately disagree. Rather than paper over it with `|| true`, which would hide real findings, the table notes it in a sentence, phrased without naming individual advisories so it does not go stale as they are dismissed or fixed.

## Testing

- `make checks` — clean across all four modules.
- `make tidy` — no churn outside `tools/`.
- `make govulncheck` — run end to end against every module, confirming the target reaches the same four the CI matrix does.
- `govulncheck` installed from the new `tools/tools.go` pin via `go install`, confirming `make install-tools` provides it (`govulncheck@v1.7.0`, Go 1.27.1).
- Workflow YAML parsed to confirm all three triggers (`push`, `pull_request`, `schedule`) and the `concurrency` block are present and well-formed. No `actionlint` in this repo or on this machine, so that check was not repeated from #1754.
- Verified SARIF upload works for fork pull requests before relying on it: CodeQL's `Analyze` job — same `security-events: write` and `upload-sarif` shape — passes on cross-repo PRs such as #1757.

This PR is itself the first live exercise of the trigger: the previous push ran all four matrix legs plus the code scanning check, and all five passed.


SaidAltury-ibm (approved):

PR #1766, opened 2026-09-08
fix(comm)!: bound the varint reader's allocation

### LINKED ISSUE #1756: Unbounded allocation in `varintReader.ReadData` when `fsc.p2p.maxRecvMsgSize` is 0

SaidAltury-ibm:
Fixes #1756

## The bug

`varintReader.ReadData` allocated `make([]byte, l)` from a length read off the
wire before reading any payload, guarded only by:

```go
if r.maxMessageSize > 0 && l > uint64(r.maxMessageSize) {
```

so `maxMessageSize == 0` meant **no limit**. Zero was reachable from config:
`NewConfig` clamped `maxRecvMsgSize` only on `< 0`, while the buffer sizes
directly above it clamp on `<= 0`.

With `fsc.p2p.maxRecvMsgSize: 0`, a 10-byte frame claiming 2^64-1 bytes **panics
the node** — `makeslice: len out of range`, unrecovered, in a per-stream
goroutine, so the process dies. Remotely triggerable by anyone who can open a
P2P stream.

Reproduced as a unit test before fixing, using the exact snippet from the issue:

```
panic: runtime error: makeslice: len out of range
  .../comm/io.(*varintReader).ReadData(...)
      .../comm/io/reader.go:57
```

That same snippet now returns `max message size [0] must be positive`.

## Amplification, measured

Even with a limit in place, the claimed length was trusted for the allocation.
Instrumenting the allocation directly:

```
wire bytes sent: 4   claimed len: 10485760   allocated: 10.02 MiB
wire bytes sent: 4   claimed len: 10485761   allocated:  0.00 MiB  (rejected)
```

**4 bytes on the wire → 10.02 MiB allocated**, per stream, before any payload
byte arrives. One reader per stream (`p2p.go:311`), each in its own goroutine.

Scope note in fairness: `libp2p/host.go:75` configures a `connmgr` (default
high-water 100 connections) and no explicit `ResourceManager`, so libp2p's
default stream caps bound the worst case. The panic path is the severe half; the
amplification is real but bounded.

## The fix

- **`config.go`** — clamp `maxRecvMsgSize`/`maxSendMsgSize` on `<= 0`, matching
  the buffer sizes above them.
- **`reader.go`** — treat a non-positive `maxMessageSize` as an error rather than
  as unlimited, so the guard holds for any future caller; read via
  `readFullGrowing`, which starts at 64 KiB and grows as data arrives, so a peer
  must actually send the bytes it claims.
- **`writer.go`** — `protoWriter.WriteMsg` carried the same `> 0` pattern, so a
  non-positive `maxSendMsgSize` disabled the send limit. The issue's expectation
  is that *no* configuration value disables the limit, which covers the send side;
  the writer now has the same guard as the reader.

`readFullGrowing` also bounds consecutive `(0, nil)` reads, which `io.Reader`
explicitly permits. My first version lacked that bound and spun **~859M
iterations in 2s** — a worse failure than the bug being fixed. It now returns
`io.ErrNoProgress`.

The `<= 0` guard additionally makes the adjacent `uint64(r.maxMessageSize)`
conversion provably safe, closing the G115 finding the issue mentions.

## ⚠️ Breaking change — please review this deliberately

`docs/configuration.md` documented zero as an escape hatch:

```
# Set to 0 to disable the limit (allow arbitrarily large messages - not recommended).
```

So this was a **documented, intentional option**, and there is a pre-existing
test named *"zero maxRecvMsgSize and maxSendMsgSize are allowed"* asserting it
passes through. This PR removes that capability: zero and negative values now
fall back to `DefaultMaxMessageSize` (10 MiB). The doc comments and that test are
updated accordingly.

Rationale: a disabled limit is an unauthenticated remote panic/OOM, so it isn't a
configuration the node should offer — "not recommended" undersells it. But this
is a real behaviour change to a documented setting, and if you would rather keep
the escape hatch (e.g. clamp in the reader only, and log a warning in config),
say so and I will rework it.

## Tests

- `reader_security_test.go` — non-positive limits rejected for both `0` and `-1`,
  on the read *and* send paths; `readFullGrowing` exactness and short-payload
  errors across the initial-buffer and growth boundaries
  (`initialReadBufferCap` ±1, 2×, 3×+7); allocation stays bounded for a huge
  claimed length; no spin on zero-byte reads.
- `config_test.go` — the zero case now asserts the default.

The `readFullGrowing` tests were written after the implementation, so I verified
they actually bite by mutation testing — all four mutants caught: up-front
allocation, missing spin guard, short payload accepted as success, and
non-positive limit treated as unlimited.

## Verification

- `go test -race` across `platform/view/services/comm/...` — all pass, including
  the websocket host and P2P session suites
- libp2p host module (separate `go.mod`) — builds and passes
- `golangci-lint run platform/view/services/comm/...` — 0 issues, `gosec`
  included (this bug was found while enabling `gosec`)
- `gofmt`, `go vet`, `go build ./...` clean

### DIFF
--- a/platform/view/services/comm/io/reader.go
+++ b/platform/view/services/comm/io/reader.go
@@ -50,14 +50,67 @@ func (r *varintReader) ReadData() ([]byte, error) {
 		return nil, err
 	}
 
-	if r.maxMessageSize > 0 && l > uint64(r.maxMessageSize) {
+	// A non-positive limit is a misconfiguration, not "unlimited": treating it as
+	// unlimited would let a peer claim any length and have it allocated below.
+	if r.maxMessageSize <= 0 {
+		return nil, errors.Errorf("max message size [%d] must be positive", r.maxMessageSize)
+	}
+	if l > uint64(r.maxMessageSize) {
 		return nil, errors.Errorf("message length [%d] exceeds max message size [%d]", l, r.maxMessageSize)
 	}
 
-	buffer := make([]byte, l) // We can re-use the buffer to avoid allocations
-
-	if n, err := io.ReadFull(r.r, buffer); err != nil || n != int(l) {
+	// The claimed length is bounded but still attacker-controlled, so it is not
+	// trusted for the allocation: grow the buffer while reading so a peer must
+	// actually send the bytes it claims.
+	buffer, err := readFullGrowing(r.r, int(l))
+	if err != nil {
 		return nil, errors.Wrapf(err, "error reading message of length [%d]", l)
+	}
+	return buffer, nil
+}
+
+// initialReadBufferCap bounds the allocation made before any payload byte has
+// been read. A message larger than this grows geometrically as data arrives.
+const initialReadBufferCap = 64 * 1024
+
+// maxConsecutiveEmptyReads bounds how many (0, nil) reads are tolerated before
+// giving up. io.Reader permits them, so without this a peer holding a stream
+// open without sending data would spin its per-stream goroutine forever.
+const maxConsecutiveEmptyReads = 100
+
+// readFullGrowing reads exactly n bytes, growing the buffer as data arrives
+// rather than allocating n up front. It returns an error if the reader ends
+// before n bytes have been read.
+func readFullGrowing(r io.Reader, n int) ([]byte, error) {
+	buffer := make([]byte, 0, min(n, initialReadBufferCap))
+	empty := 0
+	for len(buffer) < n {
+		if len(buffer) == cap(buffer) {
+			// Grow geometrically, never past the message length.
+			buffer = append(buffer, 0)[:len(buffer)]
+		}
+		read, err := r.Read(buffer[len(buffer):min(cap(buffer), n)])
+		buffer = buffer[:len(buffer)+read]
+		if err != nil {
+			// Any error is forgiven once the whole message has arrived, matching
+			// io.ReadAtLeast (which io.ReadFull wraps): bufio.Reader's large-read
+			// fast path returns the final bytes together with the underlying
+			// error, so a complete message must not be rejected because of it.
+			if len(buffer) == n {
+				break
+			}
+			if errors.Is(err, io.EOF) {
+				err = io.ErrUnexpectedEOF
+			}
+			return nil, err
+		}
+		if read == 0 {
+			if empty++; empty >= maxConsecutiveEmptyReads {
+				return nil, io.ErrNoProgress
+			}
+			continue
+		}
+		empty = 0
 	}
 	return buffer, nil
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,10 +138,10 @@ fsc:
     streamReaderBufferSize: 4096
     # Maximum allowed size (in bytes) for incoming P2P messages. Default: 10485760 (10 MiB)
     # This protects the node from memory exhaustion by rejecting oversized payloads before deserialization.
-    # Set to 0 to disable the limit (allow arbitrarily large messages - not recommended).
+    # The limit cannot be disabled: zero or negative values fall back to the default.
     maxRecvMsgSize: 10485760
     # Maximum allowed size (in bytes) for outgoing P2P messages. Default: 10485760 (10 MiB)
-    # Set to 0 to disable the limit (allow arbitrarily large messages - not recommended).
+    # The limit cannot be disabled: zero or negative values fall back to the default.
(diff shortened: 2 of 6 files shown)


platform/view/services/comm/io/reader.go
```diff
@@ -50,18 +50,67 @@ func (r *varintReader) ReadData() ([]byte, error) {
...
+		buffer = buffer[:len(buffer)+read]
+		if err != nil {
+			if errors.Is(err, io.EOF) && len(buffer) == n {
+				break
+			}
+			if errors.Is(err, io.EOF) {
+				err = io.ErrUnexpectedEOF
+			}
+			return nil, err
+		}
```
mbrandenburger, line 106: `readFullGrowing` only forgives `io.EOF` on the completing read — any other error on that call is returned even when the message is already complete (`len(buffer) == n`). `io.ReadAtLeast` (what this replaced) discards any error once enough bytes have arrived, not just EOF. 

I suggest to matching that behavior. WDYT?


For example:
```suggestion
		if err != nil {
			if len(buffer) == n {
				break
			}
			if errors.Is(err, io.EOF) {
				err = io.ErrUnexpectedEOF
			}
			return nil, err
		}
```

This keeps the io.ErrUnexpectedEOF conversion for a genuinely short read, but stops rejecting a complete message just because the last Read call happened to report a non-EOF error alongside the final bytes (which bufio.Reader's fast-path passthrough can do).

mbrandenburger (commented):
Thank you @SaidAltury-ibm for working on this. I have one comment below.

platform/view/services/comm/io/reader.go
SaidAltury-ibm: Good catch you're right, and this is a regression my code introduced.

mbrandenburger (approved):

PR #1774, opened 2026-09-08
chore(lint): enable recvcheck in golangci-lint

mbrandenburger:
`recvcheck` flags a type whose methods mix pointer and value receivers.

Part of #1755.

### Scoping

`recvcheck` reported one finding: `ApplicationProvider` (`platform/fabric/core/generic/membership/channelconfig/capabilities/application.go`) used a pointer receiver on every method except `CollectionUpgrade`, which used a value receiver. `ApplicationProvider` is only ever constructed and used through `*ApplicationProvider` — `NewApplicationProvider` returns a pointer, assigned to the `ApplicationCapabilities` interface in `channelconfig/application.go` — and is never copied or stored by value, so the value receiver was an inconsistency rather than a deliberate value-type design. Changed `CollectionUpgrade` to a pointer receiver to match the rest of the type.

### Verification

### Note for reviewers

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,7 @@ linters:
     - misspell
     - nilnesserr
     - paralleltest
+    - recvcheck
     - staticcheck
     - thelper
     - tparallel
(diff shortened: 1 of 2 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1775, opened 2026-09-08
chore(lint): enable rowserrcheck in golangci-lint

mbrandenburger:
`rowserrcheck` flags a `*sql.Rows` iteration loop that never checks `Rows.Err()` after `Next()` returns false — a mid-iteration failure (dropped connection, cancelled context, decode error) ends the loop the same way a clean finish does, so the caller can't tell the two apart.

Part of #1755.

### Scoping

Two findings, both in `platform/view/services/storage/driver/sql/common`: `BindingStore.HaveSameBinding` and `SignerInfoStore.FilterExistingSigners` iterated `*sql.Rows` without checking `rows.Err()` afterward. Added the missing check after each loop, wrapped with `pkg/utils/errors` to match the surrounding error style.

### Verification

### Note for reviewers

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,7 @@ linters:
     - misspell
     - nilnesserr
     - paralleltest
+    - rowserrcheck
     - staticcheck
     - thelper
     - tparallel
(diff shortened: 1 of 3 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1776, opened 2026-09-08
chore(lint): enable nilerr in golangci-lint

mbrandenburger:
`nilerr` flags `if err != nil { return nil }` — checking an error but returning `nil` (or another nil-valued error) instead of propagating it.

Part of #1755.

### Scoping

Five findings: three real bugs fixed, two deliberate swallows (each now logged, one also carrying `//nolint:nilerr`).

- `platform/fabric/sdk/dig/sdk.go:175` — real bug: a config-lookup error for the default network's driver was conflated with a driver-name mismatch and silently ignored, which would have nil-pointer-dereferenced on `c.Driver` had the error path ever fired. Split the error check out and return it wrapped with `errors.Wrapf`.
- `platform/fabricx/core/vault/vault.go:436` — real bug: a genuine `queryService.GetState` RPC/network error was conflated with "no committed value" and silently swallowed. Split it out and return it wrapped with `errors.Wrapf`.
- `platform/fabricx/core/vault/vault.go:441` — deliberate: `mds.GetFieldMapping` is documented to return an error on both a genuine failure and an ordinary miss (see its comment in `platform/fabric/core/generic/transaction/services.go`), so callers must treat any error as "no mapping". Documented that reasoning in a comment at the call site and added a debug log line that wasn't there before — logging `err` in the block is enough for `nilerr` to stop reporting, so no `//nolint` is needed here.
- `platform/view/services/config/config_util.go:58` — real bug: `byteSizeDecodeHook` already returns a hard "overflows uint32" error a few lines below for a related overflow case; the `ParseUint` failure a few lines above (uint64 overflow before the shift) was silently deferred to the default decoder instead of being reported the same way. Now returns the error wrapped with `errors.Wrapf`.
- `integration/fabric/events/views/events.go:49` — deliberate: this test callback captures `event.Err` into the outer `eventError` variable for the test's own assertions; the callback's own error return only triggers a log line in the listener goroutine (`RegisterChaincodeEvents`) and has no bearing on the test outcome. Added a debug log line for the error and `//nolint:nilerr` with that reasoning.

### Verification

### Note for reviewers

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,7 @@ linters:
     - makezero
     - loggercheck
     - misspell
+    - nilerr
     - nilnesserr
     - paralleltest
     - rowserrcheck
--- a/integration/fabric/events/views/events.go
+++ b/integration/fabric/events/views/events.go
@@ -44,9 +44,10 @@ func (c *EventsView) Call(viewCtx view.Context) (any, error) {
 	callBack := func(event *chaincode.Event) (bool, error) {
 		logger.Debugf("Chaincode Event Received in callback %s", event.EventName)
 		if event.Err != nil {
+			logger.Debugf("chaincode event callback received error [%s]", event.Err)
 			eventError = event.Err
 			wg.Done()
-			return true, nil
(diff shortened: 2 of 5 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1784, opened 2026-09-09
fix(deps): drop invalid applies-to from dependabot ignore rules

mbrandenburger:
PR #1781 introduced a misconfiguration in dependabot without noticing. This PR fixes it.


SaidAltury-ibm (approved):
LGTM

PR #1786, opened 2026-09-09
chore(lint): enable unconvert in golangci-lint

mbrandenburger:
`unconvert` flags a type conversion where the value already has the target type — e.g. `int(x)` where `x` is already `int`.

Part of #1755.

### Scoping

56 findings: 55 in the root module and 1 in `integration` (the other two modules had none). Every finding was a genuinely redundant conversion with no behavior change, for one of two reasons. Most of the codebase's domain identifiers (`TxNum`, `BlockNum`, `Namespace`, `PKey`, `TxID`, `RawValue`, `PeerID`, `PeerIPAddress`, `LabelName`) are declared as true Go type aliases (`type X = Y`), so converting across them is identical to converting a type to itself. The remainder — notably `FieldName` (`type FieldName string`) and the `driver.TransactionType` underlying the `TransactionType` alias — are genuinely defined types, not aliases; those deletions are safe for a different reason, namely that the conversion was a self-conversion of a value already carrying that exact named type. None of the four watch-for cases (pinning an untyped constant into an interface, crossing a type alias/named type, generics, or a deliberate width assertion) applied, so no `//nolint` was needed — all 56 conversions were deleted outright.

Removing two of the conversions turned a manual element-by-element slice copy in `Vault.Statuses` (fabricx) into two slices of the same underlying type; replaced the loop with `copy()` to satisfy staticcheck's S1001, which the change surfaced.

### Verification

### Note for reviewers

This branch is independent — rooted directly on current `main`, not stacked on any other PR in this series.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,6 +33,7 @@ linters:
     - testifylint
     - thelper
     - tparallel
+    - unconvert
     - unused
     - usestdlibvars
   settings:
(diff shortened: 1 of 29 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1789, opened 2026-09-09
chore(lint): enable gocritic in golangci-lint

mbrandenburger:
Enables `gocritic` at its default checks (style + diagnostics tags), with no added tags or curated check list, per this incremental linter-enablement series.

Part of #1755.

### Scoping

Baseline scan (before any fix): 86 findings — 57 in the root module, 29 in `integration`, 0 in the libp2p host and cc/query modules. Per-check breakdown:

- **captLocal (35)** — renamed capitalized local params/receivers to lowercase. Pure rename, no behavior change. Several sites had a param name colliding with a type/field name (e.g. `KVS`, `ID`, `NetworkID`, `PlatformRegistry`, `ExecutablePath`, `Registry`, `Default`); all such collisions were resolved by picking a distinct lowercase name and updating every use of the old identifier in scope, verified with `go build`.
- **ifElseChain (10)** — converted to a tagless `switch{}` where every branch was a simple, independent condition; suppressed with `//nolint:gocritic` (specific reason each time) where a branch had its own nested if/else-if with an independent error return, or mixed lock/unlock plus early-return control flow that a switch would flatten and obscure. Go's tagless switch evaluates cases in written order with first-match-wins — semantically identical to an if/else-if chain — so every rewrite preserves branch order and the no-match (default/else) path exactly as before.
- **elseif (3)** — two folded into the `ifElseChain` rewrites above; the third, in `platform/fabric/core/generic/msp/idemix/config.go`, is a standalone `else { if ... }` → `else if` collapse in its own hunk.
- **assignOp (9)** — mechanical `x = x op y` → `x op= y` (or `x++`) rewrites. No behavior change.
- **unlambda (9)** — replaced closures that only forward their arguments to a named function with the named function value directly (`strings.ToLower`, `strings.ReplaceAll` in three `template.FuncMap` literals and one `fsc.go` call site; `membership.NewService` in a DI provider). Every signature was checked to match exactly before the substitution.
- **singleCaseSwitch (6)** — converted `switch x { case v: ... }` (no default) to `if x == v { ... }`. Each rewrite preserves the original body verbatim; none of these switches had a default case or additional cases.
- **appendAssign (4)** — changed `s = append(s, ...)` / map-value append patterns that reused the source's backing array into `append(append([]T{}, src...), extra...)`, i.e. building into a fresh defensive copy. Removes the shared-backing-array aliasing risk `gocritic` flags; behavior-identical in every case.
- **badCall (4)** — three sites simplified a pointless single-argument `filepath.Join(x)` to bare `x`, each verified already clean; one site suppressed with `//nolint:gocritic` because the single-argument `Join` was deliberately used for `Clean`-like path normalization where behavior can differ on an empty input.
- **unslice (3)** — `x[:]` → `x`. No behavior change.
- **commentFormatting (3)** — `//text` → `// text`. No behavior change.

No default `gocritic` check produced pure-churn findings that dominated the count, and no non-default check was identified as worth adding as a follow-up.

### Verification

`go build` passes in both the root and `integration` modules. `go test` passes for every touched package — a pre-existing timing-sensitive test, `TestIdentityCacheClose` in `platform/fabric/core/generic/msp/idemix`, flakes under load independent of this change (confirmed clean under `-race` and in isolation; the touched files in that package are cosmetic renames only). `make checks` exits 0.

### Note for reviewers

This branch is independent — rooted directly on current `main`, not stacked on any other PR in this series.


SaidAltury-ibm (approved):

PR #1792, opened 2026-09-09
chore(ci): harden workflow token permissions

mbrandenburger:
## Summary
Addresses the OpenSSF Scorecard Token-Permissions findings (currently scoring 0):
- `codeql-analysis.yml`, `git.yml`, and `tests.yml` had no top-level `permissions` block. Scorecard flags this because a job added later without explicit permissions would silently inherit the repo/org default token scope instead of least privilege. Added `permissions: contents: read` at the top level of each (job-level overrides in `codeql-analysis.yml` are unchanged).
- `on-pr.yaml`, `on-pr-update.yaml`, and `on-pr-review.yaml` granted top-level `checks: write`. None of the bot scripts these workflows run (`bot-on-pr-open.js`, `bot-on-pr-update.js`, `bot-on-pr-review.js`) call the GitHub Checks API — `helpers/checks.js` is local DCO/GPG/merge-conflict validation, unrelated to that token scope. Removed the unused permission.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -26,6 +26,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
   analyze:
     name: Analyze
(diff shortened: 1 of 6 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1764, opened 2026-09-08
test(vault): cover interceptor and vault error and lifecycle paths

sid200727:
Fixes #1758

Completes the package coverage started in #1759, taking `platform/common/core/generic/vault` past the Stage 3 target.

## Context

The conformance suite in `vault_test_utils.go` drives the happy paths through a working store, so the interceptor's closed-instance guards, query-executor failures, version mismatches and comparison branches went untested, as did the vault's status and lifecycle wrappers. This covers those.

## interceptor.go - now at 100%

Adds three query executors alongside the existing `fake.QE`, so branches the passing fake cannot reach become testable:

- a failing executor, for the error paths in `GetState`, `GetStateMetadata`, `GetDirectState`, `GetReadAt` and `IsValid`
- a stale-version executor, for the version-mismatch rejections in `GetState`, `GetStateMetadata` and `IsValid`
- a nil-state executor, for a key absent from storage

Also covers the closed-instance rejections shared by `Clear`, `GetReadKeyAt`, `GetReadAt` and `SetStateMetadatas`; the option-validation branches (`len(opts) > 1` and an unrecognised option) on both getters; the write-only case where no executor is attached; `Reopen` in both the closed and already-open states; `Done`'s idempotence and its logged-but-not-propagated close failure; and all six comparison branches of `Equals` across the `*Interceptor` and `*Inspector` cases, plus the unsupported-type default.

## vault.go

Covers the six functions that were at 0%: `SetDiscarded`, `NewRWSetWithIsolationLevel`, `Close`, `RWSExists`, `Statuses` and `SetStatus`.

`UnmapInterceptor`'s contract is pinned as well, it refuses to unmap an interceptor whose `Done` has not been called, and reports unknown transactions rather than silently succeeding. Both were previously untested.

## A note on parallelism

The vault tests open an in-memory store, so they are serialized with `//nolint:paralleltest` and named from `t.Name()`, following the fix in #1470. Running them in parallel triggers the same `database schema is locked` failure that change addressed reproducible here with `-count=3` before the fix, clean after.

## Remaining partials

Five functions stay short of 100%, all in `vault.go`'s commit path — `commitTXs` (88.0%), `commitRWs` (81.0%), `newRWSet` (88.9%), `Match` (72.0%) and `Statuses` (77.8%). Their uncovered branches are store-failure paths that would need a fault-injecting `VaultStore` implementation, which felt like disproportionate scaffolding for the statements involved. Happy to add it if you'd rather have them closed.

## Testing

[...]

PR #1793, opened 2026-09-09
chore(lint): enable revive in golangci-lint

mbrandenburger:
`revive` is a configurable style linter; this enables it with a curated 21-rule list (`linters.settings.revive`, replacing revive's defaults) rather than its full default set.

Part of #1755. This is the last linter in the series.

### Scoping

1119 findings across all 4 modules (909 root, 187 `integration`, 21 libp2p comm-host, 2 cc/query), triaged per rule:

- **Mechanical, applied directly**: `unused-parameter` (374) and `unused-receiver` (482), renamed to `_` (parameter/receiver names aren't part of a function's type, so this is safe even on exported functions); also most `var-naming`, `error-naming`, `error-strings`, `errorf`, `indent-error-flow`, `superfluous-else`, `early-return`, `unnecessary-stmt`, `bool-literal-in-expr` and `receiver-naming` findings on unexported identifiers.
- **errorf**: fixed with this project's `errors.Errorf`, not revive's suggested `fmt.Errorf`, which project convention forbids.
- **confusing-naming**: fixed by renaming the unexported side wherever nothing outside the package required that name. Deferred with `//nolint:revive` only where both colliding names are exported, typically because each independently satisfies the same exported interface.
- **var-naming on exported identifiers**: deferred with `//nolint:revive` per instance, since renaming an exported field, type, func, const or method is an API break. Test-only exported identifiers used solely inside their own test package were renamed instead.
- **unexported-return**: deferred via an `issues.exclusions.rules` entry scoped to revive and to this rule's text, over an explicit path list. A constructor returning an unexported concrete type that satisfies an exported interface is a deliberate Go idiom for hiding an implementation, not a defect — changing it is a public-API decision, not a lint fix. 121 findings across 68 root-module files, 9 across 7 integration files, 2 in the libp2p comm-host module.

`exported` and `package-comments` are deliberately left out of the rule list — golangci-lint only runs the rules named here, so omitting a rule is how it's disabled, rather than enabling it and suppressing every finding. Both are public-API documentation work, not a lint-commit fix: `exported` measured 3672 findings (one per exported symbol missing a doc comment) and `package-comments` measured 268 (one per package missing a package doc comment); 273 packages need one or the other. That documentation effort is tracked as a follow-up rather than done here.

One naming cascade worth calling out: `confusing-naming` flagged the vault package's test helper `newInterceptor` against the exported `NewInterceptor`. Renaming it collided with an unrelated `newTestInterceptor(qe VersionedQueryExecutor)` helper added to the same package by #1764 after this commit was originally written, so the renamed helper is `newVaultInterceptor` instead; #1764 also added three small fake query-executor types whose methods this rule flags for unused receivers, fixed the same way as everywhere else in this change.

### Verification

`go build`, `go vet` and `go test` all pass in the root, `integration` and libp2p modules. `make checks` exits 0 across all four modules.

### Note for reviewers

This branch is independent — rooted directly on current `main`, not stacked on any other PR in this series. It touches nearly every `.go` file in the repository (naming/style rules), but every hunk is either a mechanical rename or a `//nolint`/exclusion with a stated reason.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,7 @@ linters:
     - nilnesserr
     - paralleltest
     - predeclared
+    - revive
     - spancheck
     - rowserrcheck
     - recvcheck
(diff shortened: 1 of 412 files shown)


SaidAltury-ibm (approved):
LGTM

PR #1795, opened 2026-09-09
chore(ci): add nightly fuzz-testing workflow

### LINKED ISSUE #1794: Add CI job that runs the existing fuzz tests
The tasks is to wire up a nightly CI job that runs the existing four targets with go test -fuzz=Fuzz... -fuztime=<budget>

We want a time budget of 2hrs; the script should use corpus caching, and upload the failing inputs on a crash as build artifact.

mbrandenburger:
### Description

Runs the existing `Fuzz*` targets in `platform/fabric/core/generic/transaction` (added in #1594) as real mutation-based fuzzing instead of just their seed corpus, which is all `go test` exercises today.

- Sunday-night schedule (`workflow_dispatch` also available for manual runs), one matrix job per `Fuzz` target so each gets its own full 2h `-fuzztime` budget.
- Caches the corpus the fuzzer discovers (under `GOCACHE/fuzz`) between runs, keyed per target, so each week resumes from the last one instead of restarting cold.
- Uploads any crashing input found as a build artifact so it can be pulled into the checked-in seed corpus via a normal PR.

Closes #1794.

### Test plan

- [x] `ratchet lint` clean on the new workflow.
- [x] Smoke-tested the exact `go test` invocation locally with a 3s `-fuzztime` budget — runs and passes.
- [ ] First scheduled/manual run on CI to confirm the cache step behaves as expected end to end.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -2,9 +2,9 @@ name: Fuzz
 
 on:
   schedule:
-    # Sunday night UTC. Off the round hour/minute so this doesn't land in the
-    # same instant as every other repo's midnight cron -- see govulncheck.yml.
-    - cron: '11 23 * * 0'
+    # Nightly at 3am UTC. Off the round minute so this doesn't land in the
+    # same instant as every other repo's 3am cron.
+    - cron: '13 3 * * *'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,73 @@
...
+        # only fuzzes a single target per invocation, so each gets its own job
+        # and its own full 2h budget. Add new targets here as they land (see #1787).
+        target:
+          - FuzzUnpackSignedProposal
+          - FuzzUnpackEnvelopeFromBytes
+          - FuzzTransactionSetFromBytes
+          - FuzzTransactionSetFromEnvelopeBytes
+    steps:
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # ratchet:actions/checkout@v5
+      - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # ratchet:actions/setup-go@v6
```
SaidAltury-ibm: these are two major versions behind every other workflow, i think we should line it up with the others

.github/workflows/fuzz.yml
```diff
@@ -0,0 +1,73 @@
+name: Fuzz
+
+on:
+  schedule:
+    # Sunday night UTC. Off the round hour/minute so this doesn't land in the
+    # same instant as every other repo's midnight cron -- see govulncheck.yml.
+    - cron: '11 23 * * 0'
```
SaidAltury-ibm: The schedule is weekly, but the PR says "nightly"

SaidAltury-ibm (changes requested):
thank you @mbrandenburger, i have left 2 comments please tell me WYT?

.github/workflows/fuzz.yml
mbrandenburger: Good catch

.github/workflows/fuzz.yml
mbrandenburger: Good catch 🙈

SaidAltury-ibm (approved):

PR #1797, opened 2026-09-10
fix(fabric): check the validation code on non-filtered delivery

### LINKED ISSUE #1737: Non-filtered `DeliverReceive` reports invalid transactions as committed
### What happened?

`DeliverReceive` has two branches. The `FilteredBlock` branch checks each matched transaction's `TxValidationCode` before declaring it committed. The full-`Block` branch does not — it matches the transaction ID and unconditionally sets `Committed = true`, never consulting the block's `TRANSACTIONS_FILTER` metadata. So on a non-filtered stream, `FabricFinality.IsFinal()` returns `nil` — "this transaction is final and valid" — for a transaction Fabric invalidated.

Expected: both branches agree, and an invalidated transaction is reported as invalid regardless of stream type.

The filtered branch is correct (`platform/fabric/core/generic/delivery/deliverclient.go:201-217`):

```go
case *pb.DeliverResponse_FilteredBlock:
	filteredTransactions := r.FilteredBlock.FilteredTransactions
	for i, tx := range filteredTransactions {
		if tx.Txid == txid {
			if tx.TxValidationCode == pb.TxValidationCode_VALID {
				event.Committed = true
				...
			} else {
				event.Err = errors.Errorf("transaction [%s] status is not valid: %s", tx.Txid, tx.TxValidationCode)
			}
			break read
		}
	}
```

The full-block branch is not (`platform/fabric/core/generic/delivery/deliverclient.go:218-233`):

```go
case *pb.DeliverResponse_Block:
	if r.Block == nil || r.Block.Data == nil || r.Block.Header == nil {
		event.Err = errors.Errorf("received malformed block from peer %s", address)
		break read
	}
	for i, tx := range r.Block.Data.Data {
		_, _, chdr, err := fabricutils.UnmarshalTx(tx)
		if err != nil {
			event.Err = errors.Wrapf(err, "error parsing transaction [%d,%d]", r.Block.Header.Number, i)
			break read
		} else if chdr.TxId == txid {
			event.Committed = true      // ← no validation-code check
			event.Block = r.Block.Header.Number
			event.IndexInBlock = i
			break read
		}
	}
```

Which branch runs is chosen by `useFilteredDelivery` (`platform/fabric/core/generic/finality/fabric.go:83-88`):

```go
if d.useFiltered {
	deliverStream, err = deliverClient.NewDeliverFiltered(ctx)
} else {
	deliverStream, err = deliverClient.NewDeliver(ctx)
}
```

Why it is latent today. The fabric SDK wires `useFilteredDelivery: true` — the last positional argument to `generic.NewChannelProvider` (`platform/fabric/sdk/dig/generic/providers.go:211`, signature at `platform/fabric/core/generic/channelprovider.go:116`). And `platform/fabricx`, which threads the same flag through `fabricx/core/channel/provider.go:37`, constructs no `FabricFinality` at all — its channel provider wires a `noopDeliveryService` and a stub committer (`fabricx/core/channel/provider.go:100`, `:124`), so this code path is unreachable there. Nothing in the tree reaches the bug today; it is one boolean away.

Impact if reached: `ch.Finality().IsFinal(ctx, txID)` returns success for a transaction Fabric rejected — an MVCC conflict, an endorsement-policy failure, a phantom read. The application proceeds as though its transaction committed. Silent, and on the wrong side: a false *positive*, unlike the `SeekNewest` false negative that the other findings on this path concern.

### How can we reproduce it?

The bug is not reachable through configuration on `main`, since `useFilteredDelivery` is hardcoded `true` at `providers.go:211`. Two ways to observe it:

By code inspection:

1. Compare `deliverclient.go:201-217` (filtered — checks `TxValidationCode`) with `:218-233` (full block — does not).
2. Confirm nothing else revalidates: `FabricFinality.IsFinal` returns `nil` whenever `DeliverWaitForResponse` reports `committed == true` (`finality/fabric.go:115-121`).

As a unit test — `platform/fabric/core/generic/delivery/deliverclient_test.go` already drives `DeliverReceive` against a fake stream:

1. Construct a full `DeliverResponse_Block` containing the target transaction.
2. Set its `TRANSACTIONS_FILTER` metadata entry to mark that transaction e.g. `TxValidationCode_MVCC_READ_CONFLICT`.
3. Call `DeliverReceive` with the matching txID.
4. Assert `event.Committed == false` and `event.Err != nil`. It currently reports committed with no error.

### Environment

- Operating system (e.g., Linux, MacOS): MacOS (platform-independent — missing validation check, not environment-specific)

### Additional Information

Found by code inspection while verifying a third-party analysis of the finality feature against the implementation. Not reachable at runtime on `main`, so filed as a latent defect rather than an observed failure — but it sits directly behind a boolean that exists precisely so it can be flipped.

Suggested fix — in the `DeliverResponse_Block` branch, read the validation code on a txID match before deciding:

```go
} else if chdr.TxId == txid {
	code, err := validationCodeAt(r.Block, i)
	if err != nil {
		event.Err = err
		break read
	}
	if pb.TxValidationCode(code) != pb.TxValidationCode_VALID {
		event.Err = errors.Errorf("transaction [%s] status is not valid: %s", txid, pb.TxValidationCode(code))
		break read
	}
	event.Committed = true
	event.Block = r.Block.Header.Number
	event.IndexInBlock = i
	break read
}
```

The helper already exists a file away: `validationCodeAt(block, i)` in `platform/fabric/core/generic/delivery/service.go:35-44`, complete with the bounds checks a malformed or malicious block needs, and in the same package so it is directly callable. Note that the branch's existing nil check does not cover `r.Block.Metadata`, which `validationCodeAt` handles itself.

Related:

- #1732 and #1735 — other defects on the `FabricFinality.IsFinal` path.

SaidAltury-ibm:
### Description

`DeliverReceive` has one branch per stream shape. The `FilteredBlock` branch reads each matched transaction's `TxValidationCode` before declaring it committed; the full-`Block` branch matched the transaction ID and set `Committed = true` without ever consulting the block's `TRANSACTIONS_FILTER` metadata. Matching a transaction ID only proves the transaction was *included* in the block — Fabric includes rejected transactions too — so the branch conflated "present in a block" with "committed successfully". Nothing downstream revalidates, so `ch.Finality().IsFinal()` reported "final and valid" for a transaction Fabric rejected on an MVCC conflict or a failed endorsement policy: a silent false positive.

- Reads the verdict via `validationCodeAt` before reporting the transaction committed, matching the filtered branch. That helper already lives in this package (`delivery/service.go:35`) with the bounds checks a malformed or malicious block needs, and already has two production callers in `Service.Scan` and `Service.ScanFromBlock` — both of which fail closed on a metadata error, so this follows the package's existing convention rather than introducing one.
- Behaviour change: a block whose `TRANSACTIONS_FILTER` is absent, or too short to cover the matched transaction, previously reported committed and now reports an error. Failing closed is the point — a block carrying no validity information for a transaction cannot be said to have committed it.
- Fixes a pre-existing test that asserted the buggy behaviour: `"DeliverResponse_Block VALID"` built a block with no `Metadata` at all, so it required a block with *no validity information* to be treated as valid. It now builds a block with a `VALID` filter entry, via a `blockWithTx` helper the new cases share.
- Latent on `main`: the only caller of `generic.NewChannelProvider` passes `useFilteredDelivery: true` (`sdk/dig/generic/providers.go:209`), and `fabricx` constructs no `FabricFinality` at all. It sits behind a boolean that exists to be flipped, and whoever flips it will be reaching for a performance knob with no reason to suspect it also turns off validity checking.

Closes #1737.

### Test plan

- [x] Four cases added, each failing before the fix: invalid transaction rejected; missing validation metadata fails closed; truncated filter fails closed; and both branches report an invalid transaction with byte-identical errors (the issue's stated expectation that the two branches agree).
- [x] `go test -race` across `delivery`, `finality` and `committer` — all pass.
- [x] `golangci-lint run platform/fabric/core/generic/delivery/...` — 0 issues, including the newly-enabled `revive`.
- [x] Full `platform/fabric/...` passes except the two Postgres tests needing a local Docker daemon, which fail identically on an unmodified tree.

### DIFF
--- a/platform/fabric/core/generic/delivery/deliverclient.go
+++ b/platform/fabric/core/generic/delivery/deliverclient.go
@@ -226,6 +226,21 @@ read:
 					event.Err = errors.Wrapf(err, "error parsing transaction [%d,%d]", r.Block.Header.Number, i)
 					break read
 				} else if chdr.TxId == txid {
+					// A full block carries its verdicts in TRANSACTIONS_FILTER metadata
+					// rather than inline, so matching the transaction ID only proves the
+					// transaction was included in the block - Fabric includes rejected
+					// transactions too. Read the verdict before reporting it committed,
+					// matching the FilteredBlock branch above.
+					code, err := validationCodeAt(r.Block, i)
+					if err != nil {
+						event.Err = err
+						break read
+					}
+					if pb.TxValidationCode(code) != pb.TxValidationCode_VALID {
+						logger.Debugf("transaction [%s] in block [%d] is not valid [%s]", txid, r.Block.Header.Number, pb.TxValidationCode(code))
+						event.Err = errors.Errorf("transaction [%s] status is not valid: %s", txid, pb.TxValidationCode(code))
+						break read
+					}
 					event.Committed = true
 					event.Block = r.Block.Header.Number
 					event.IndexInBlock = i
--- a/platform/fabric/core/generic/delivery/deliverclient_test.go
+++ b/platform/fabric/core/generic/delivery/deliverclient_test.go
@@ -384,22 +384,10 @@ func TestDeliverReceive(t *testing.T) {
 
 	t.Run("DeliverResponse_Block VALID", func(t *testing.T) {
 		t.Parallel()
-		chdr := &cb.ChannelHeader{TxId: "tx1"}
-		chdrBytes, _ := proto.Marshal(chdr)
-		payload := &cb.Payload{Header: &cb.Header{ChannelHeader: chdrBytes}}
-		payloadBytes, _ := proto.Marshal(payload)
-		env := &cb.Envelope{Payload: payloadBytes}
-		envBytes, _ := proto.Marshal(env)
-
 		df := &mockDeliverStream{
 			recvResp: &pb.DeliverResponse{
 				Type: &pb.DeliverResponse_Block{
-					Block: &cb.Block{
-						Header: &cb.BlockHeader{Number: 11},
-						Data: &cb.BlockData{
-							Data: [][]byte{envBytes},
-						},
-					},
+					Block: blockWithTx(11, []string{"tx1"}, pb.TxValidationCode_VALID),
 				},
 			},
 		}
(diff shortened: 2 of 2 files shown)


mbrandenburger (approved):
Thank you @SaidAltury-ibm

PR #1798, opened 2026-09-10
fix(fabric): guard the nil PickPeer returns

### LINKED ISSUE #1732: `PickPeer` can return nil and both finality callers dereference it, panicking the node
### What happened?

`ConfigService.PickPeer` returns `nil` when neither the requested peer-function bucket nor the `PeerForAnything` fallback bucket has any entry. Two callers on the finality path dereference the result without checking, so a reachable `core.yaml` shape panics the node on the first finality check for an unknown transaction.

Expected: a configuration with no peer usable for finality is either rejected at startup or reported as an error from `IsFinal()` — not a nil-pointer panic under load.

`PickPeer` signals the condition by returning nil (`platform/fabric/core/generic/config/service.go:331-340`):

```go
func (s *Service) PickPeer(ft driver.PeerFunctionType) *ConnectionConfig {
	source, ok := s.peerMapping[ft]
	if !ok {
		source = s.peerMapping[driver.PeerForAnything]
	}
	if len(source) == 0 {
		return nil
	}
	return source[rand.Intn(len(source))]
}
```

Caller 1 — field access on the result (`platform/fabric/core/generic/committer/committer.go:399`):

```go
peerForFinality := c.ConfigService.PickPeer(driver.PeerForFinality).Address
```

Caller 2 — pointer dereference on the result (`platform/fabric/core/generic/finality/fabric.go:70`):

```go
client, err := d.Services.NewPeerClient(*d.ConfigService.PickPeer(driver.PeerForFinality))
```

`peerMapping` only gains a key for a function type that at least one peer declared (`config/service.go:362-375`), and only a peer with an empty `usage` field lands in `PeerForAnything` (`funcTypeMap`). So the panic is reached whenever every configured peer carries an explicit `usage:` that is not `finality`.

The misconfiguration is not caught at startup: `createPeerMap` only logs a warning for an *unrecognised* usage string (`service.go:373`); a recognised-but-wrong usage passes silently.

Other `PickPeer` callers have the same exposure and should be audited in the same pass — `delivery.connect` at `delivery/delivery.go:372` uses `PickPeer(PeerForDelivery)` and dereferences the result too (`*peerConnConf`), which panics for a config that declares only `usage: finality` peers.

### How can we reproduce it?

1. Configure a channel whose peers all carry an explicit `usage:` that is not `finality`:

   ```yaml
   peers:
     - address: peer0.example.com:7051
       connectionTimeout: 10s
       tlsEnabled: true
       usage: delivery      # → PeerForDelivery bucket only; PeerForAnything stays empty
   ```

2. Start the node. Startup succeeds — no validation error, no warning.
3. Submit a transaction, or call `ch.Finality().IsFinal(ctx, txID)` for a transaction ID the local vault does not know (`Status()` returns `Unknown`).
4. `Committer.IsFinal` reaches the `driver.Unknown` branch (`committer.go:391`) and evaluates `PickPeer(driver.PeerForFinality).Address` at `committer.go:399`: `PickPeer` finds no `finality` key, falls back to `PeerForAnything`, finds that empty too, returns `nil`, and the field access panics.

As a unit test, `platform/fabric/core/generic/config/service_test.go` already builds a `Service` from a config fixture: add a case whose peers all declare `usage: delivery` and assert `PickPeer(driver.PeerForFinality)` returns nil; then a committer/finality test asserting the caller returns an error rather than panicking.

### Environment

- Operating system (e.g., Linux, MacOS): MacOS (platform-independent — nil dereference, not environment-specific)

### Additional Information

Found by code inspection while verifying a third-party analysis of the finality feature against the implementation, not by observing the panic at runtime. The reproduction steps above are derived, not executed.

Suggested fix — guard at the call sites rather than defending inside `PickPeer`. A nil return is a meaningful "no peer configured for this role" answer, and the caller is the only one able to describe it:

```go
// committer.go
cc := c.ConfigService.PickPeer(driver.PeerForFinality)
if cc == nil {
    return errors.Errorf("no peer configured for finality on channel [%s]", c.ChannelConfig.ID())
}
```

Optionally also warn at startup when a channel ends up with both an empty `PeerForFinality` and an empty `PeerForAnything` bucket, so the misconfiguration surfaces before traffic does.

A second, cosmetic defect on the same lines, worth fixing in the same change: `FabricFinality.IsFinal(txID, address string)` ignores its `address` argument when connecting — `fabric.go:70` calls `PickPeer` again — while `address` is used only for error messages and `TxEvent.CommitPeer`. So the peer the committer logs (`committer.go:399`, `:401`) may not be the peer actually probed whenever more than one `usage: finality` peer exists. Harmless to correctness, actively misleading when debugging a finality timeout. Either pass the picked config through, or drop the parameter.

Related:

- #1731 — same call path.
- #1735 — same function, and touches the same two lines.

SaidAltury-ibm:
### Description

`ConfigService.PickPeer(role)` returns `nil` when neither the requested peer-function bucket nor the `PeerForAnything` fallback has an entry (`config/service.go:342`) — a deliberate "no peer for this role" signal. Three callers dereferenced it without checking, so a node whose peers all declare some *other* role panics on first use. It is reachable from a plain `core.yaml`: `peerMapping` only gains a key for a role some peer explicitly declared, and a peer lands in `PeerForAnything` only if its `usage:` field is empty — so a single peer with `usage: delivery` is enough to make `PickPeer(PeerForFinality)` return nil. Startup does not catch it: `createPeerMap` warns only for an *unrecognised* usage string, so a recognised-but-wrong one passes silently and the crash arrives later under traffic.

- Guards the three dereference sites: `finality/fabric.go:70` (peer client for the finality probe), `delivery/delivery.go:372` (`connect`), and `chaincode/discovery.go:218` (`query`).
- Follows the shape already in the tree for the sibling `PickOrderer` at `ordering/cft.go:124` — guarded at the call site rather than inside `PickPeer`, since only the caller knows which role it asked for and what to say when there isn't one.
- `ledger.go:108` also calls `PickPeer` - we use the same pattern.
- Drops the unused `address` parameter from `FabricFinality.IsFinal`. It was never used to connect — the implementation called `PickPeer` again — so with more than one finality peer the address the committer logged need not be the peer actually probed, which misleads in exactly the logs read when debugging a finality timeout. The implementation now reports the peer it picked, and the committer's debug lines no longer name a peer they cannot vouch for.
- Skipped the optional startup warning the issue mentions: the guards already turn a crash into a clear error, and a startup check is a second place to keep in sync. Worth adding if operators actually hit this.

Closes #1732.

### Test plan

- [x] `finality/fabric_test.go`: `PickPeer` returns nil, asserts an error rather than a panic. Before the fix this test **panicked** with `nil pointer dereference` — the bug itself.
- [x] `config/service_test.go`: one line asserting `PickPeer(PeerForFinality) == nil` for the `usage: query` / `usage: delivery` fixture already in that test — the reachability proof, without which the other test looks hypothetical.
- [x] Mutation-checked: flipping the finality guard to `if false` brings the panic straight back, so the test is load-bearing.
- [x] `go test -race` across `committer`, `finality`, `delivery`, `chaincode`, `config` — all pass.
- [x] `golangci-lint run platform/fabric/...` — 0 issues.

### DIFF
--- a/platform/fabric/core/generic/committer/committer.go
+++ b/platform/fabric/core/generic/committer/committer.go
@@ -56,7 +56,7 @@ type (
 )
 
 type FabricFinality interface {
-	IsFinal(txID, address string) error
+	IsFinal(txID string) error
 }
 
 type CommitTx struct {
@@ -395,15 +395,14 @@ func (c *Committer) IsFinal(ctx context.Context, txID string) error {
 			if logger.IsEnabledFor(zapcore.DebugLevel) {
 				c.logger.Debugf("Tx [%s] is unknown with no deps, remote check [%d][%s]", txID, iter, debug.Stack())
 			}
-			peerForFinality := c.ConfigService.PickPeer(driver.PeerForFinality).Address
-			err := c.FabricFinality.IsFinal(txID, peerForFinality)
+			err := c.FabricFinality.IsFinal(txID)
 			if err == nil {
-				c.logger.Debugf("Tx [%s] is final, remote check on [%s]", txID, peerForFinality)
+				c.logger.Debugf("Tx [%s] is final, remote check succeeded", txID)
 				return nil
 			}
 
 			if vd, _, err2 := c.Status(ctx, txID); err2 == nil && vd == driver.Unknown {
-				c.logger.Debugf("Tx [%s] is not final for remote [%s], return [%s], [%d][%s]", txID, peerForFinality, err, vd, err2)
+				c.logger.Debugf("Tx [%s] is not final remotely, return [%s], [%d][%s]", txID, err, vd, err2)
 				return err
 			}
 		default:
--- a/platform/fabric/core/generic/finality/fabric.go
+++ b/platform/fabric/core/generic/finality/fabric.go
@@ -61,13 +61,22 @@ func NewFabricFinality(
 	return d, nil
 }
 
-func (d *FabricFinality) IsFinal(txID, address string) error {
+// IsFinal probes a peer configured for finality for the given transaction. The
+// peer is chosen here rather than supplied by the caller, so that the address
+// reported in errors and events is always the peer actually probed.
+func (d *FabricFinality) IsFinal(txID string) error {
 	d.Logger.Debugf("remote checking if transaction [%s] is final in channel [%s]", txID, d.Channel)
 	var eventCh chan delivery.TxEvent
 	var ctx context.Context
 	var cancelFunc context.CancelFunc
 
-	client, err := d.Services.NewPeerClient(*d.ConfigService.PickPeer(driver.PeerForFinality))
+	peerConnConf := d.ConfigService.PickPeer(driver.PeerForFinality)
+	if peerConnConf == nil {
+		return errors.New("no peer configured for finality")
+	}
+	address := peerConnConf.Address
+
+	client, err := d.Services.NewPeerClient(*peerConnConf)
 	if err != nil {
 		return errors.WithMessagef(err, "failed creating peer client for address [%s]", address)
 	}
--- a/platform/fabric/core/generic/finality/fabric.go
+++ b/platform/fabric/core/generic/finality/fabric.go
@@ -61,9 +61,8 @@ func NewFabricFinality(
 	return d, nil
 }
 
-// IsFinal probes a peer configured for finality for the given transaction. The
-// peer is chosen here rather than supplied by the caller, so that the address
-// reported in errors and events is always the peer actually probed.
+// IsFinal probes a peer configured for finality for the given transaction,
+// returning an error if no peer is configured for that role.
 func (d *FabricFinality) IsFinal(txID string) error {
 	d.Logger.Debugf("remote checking if transaction [%s] is final in channel [%s]", txID, d.Channel)
 	var eventCh chan delivery.TxEvent
--- a/platform/fabric/core/generic/chaincode/discovery.go
+++ b/platform/fabric/core/generic/chaincode/discovery.go
@@ -215,7 +215,11 @@ func (d *Discovery) query(req *discovery.Request) (discovery.Response, error) {
 			pCli.Close()
 		}
 	}()
-	pc, err := d.chaincode.Services.NewPeerClient(*d.chaincode.ConfigService.PickPeer(driver.PeerForDiscovery))
+	peerConnConf := d.chaincode.ConfigService.PickPeer(driver.PeerForDiscovery)
+	if peerConnConf == nil {
+		return nil, errors.New("no peer configured for discovery")
+	}
+	pc, err := d.chaincode.Services.NewPeerClient(*peerConnConf)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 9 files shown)


platform/fabric/core/generic/finality/fabric.go
```diff
@@ -61,13 +61,22 @@ func NewFabricFinality(
 	return d, nil
 }
 
-func (d *FabricFinality) IsFinal(txID, address string) error {
+// IsFinal probes a peer configured for finality for the given transaction. The
+// peer is chosen here rather than supplied by the caller, so that the address
+// reported in errors and events is always the peer actually probed.
```
mbrandenburger: I think this godoc reads nice if you know about the API change; in the future I believe the following comment is just confusing:

> rather than supplied by the caller, so that the address reported in errors and events is always the peer actually probed

platform/fabric/core/generic/committer/committer.go
```diff
@@ -56,7 +56,7 @@ type (
 )
 
 type FabricFinality interface {
-	IsFinal(txID, address string) error
+	IsFinal(txID string) error
```
mbrandenburger, line 59: I am wondering who is a consumer of this interface? Do we break Panurus with it? If so - we should clearly mark this as breaking change.

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for creating this PR.

> `ledger.go:108` also calls PickPeer but passes the pointer on without dereferencing it, so it cannot panic and is left alone.

I was wondering if this is really the case, nevertheless, I think it would be cleaner to follow the same error handling scheme you introduced here and do something like in `ledger.go`:
```go
if peerConnConf == nil {
    return nil, errors.New("no peer configured for query")
}
return c.ChaincodeManager.Chaincode(QuerySystemChaincode).
    NewInvocation(function, params...).
    WithSignerIdentity(c.LocalMembership.DefaultIdentity()).
    WithEndorsersByConnConfig(peerConnConf).
    Query()
```

WDYT?


Please also see my other comments below.

platform/fabric/core/generic/committer/committer.go
SaidAltury-ibm: Not breaking for Panurus, and the public API is untouched.

`committer.FabricFinality` is exported but consumed only in-tree: channelprovider, the committer itself, the dig provider at `providers.go:149`, and a test fake. The public `driver.Finality.IsFinal(ctx, txID)` is unchanged by this PR.

I checked Panurus against its current `main` (`191d1dd8`, all eight modules on FSC v0.18.0) rather than assuming. It does import `core/generic/committer`, but only for `DependencyResolver`, `MapValidationCode`, `NewParallelDependencyResolver`, `ProcessNamespace` and `ValidationFlags` — no reference to `FabricFinality` anywhere, and its finality path goes through `ch.Finality()`.

platform/fabric/core/generic/finality/fabric.go
SaidAltury-ibm: good catch, thank you!

mbrandenburger (approved):

PR #1807, opened 2026-09-10
fix(fabric): do not wrap a nil err as if it were the failure cause

### LINKED ISSUE #1806: Nil pointer dereference in UnpackEnvelopePayload on malformed envelope
## Summary
`FuzzTransactionSetFromEnvelopeBytes` found an input that panics with a nil-pointer SIGSEGV in `transaction.SetFromEnvelopeBytes`, called from `UnpackEnvelopePayload` (`platform/fabric/core/generic/transaction/envelope.go`). This code parses attacker-controlled bytes before signature verification, so a panic here is a remotely triggerable DoS.

## Root cause
`UnpackEnvelopePayload`'s `pRespPayload.Extension == nil` branch returns `errors.Wrap(err, "nil pRespPayload.Extension")`, but `err` is already `nil` at that point: it holds the result of the preceding, successful unmarshal call. `errors.Wrap` returns `nil` when given a `nil` error, so the function returns `(nil, nil)` instead of an error. `SetFromEnvelopeBytes` only checks `err != nil` before dereferencing the result, so it dereferences the nil `*UnpackedEnvelope` and panics at `transaction.go:242`.

## Reproduction
```
go test -run=FuzzTransactionSetFromEnvelopeBytes ./platform/fabric/core/generic/transaction -v
```
panics on the crasher input before the fix.

## Fix
Use `errors.New("nil pRespPayload.Extension")` instead of wrapping the stale `err`. This turned up the same pattern, `errors.Wrap`/`Wrapf` reusing an `err` that's nil because the enclosing branch checks something else entirely, at two
more call sites:
- `state.GetCertification` reported a missing certification as success, with a nil payload.
- `ordering.CFTBroadcaster.Broadcast` returned nil (success) when `BroadcastNumRetries()` is configured as `0`.

The crasher input becomes a permanent regression entry at `platform/fabric/core/generic/transaction/testdata/fuzz/FuzzTransactionSetFromEnvelopeBytes/`.

mbrandenburger:
## Summary
- A fuzz-found nil-pointer panic in `transaction.UnpackEnvelopePayload`. A nil-check branch wrapped an already-nil `err` with `errors.Wrap`, which returns `nil` on a `nil` error, so the function returned `(nil, nil)` and the caller dereferenced a nil `*UnpackedEnvelope`.
- Fixes the same pattern at one more call site, found by checking every `errors.Wrap`/`Wrapf(err, ...)` call in the codebase for a guard condition that doesn't mention `err`: `state.GetCertification` reported a missing certification as success.
- Adds the fuzzer's crasher input as a permanent regression entry under `testdata/fuzz/FuzzTransactionSetFromEnvelopeBytes/`, along with two related seed corpus entries (nil and empty input).
- Updates a `certification_test.go` case that had asserted on the bug's downstream symptom (a later, unrelated "failed getting channel" error) rather than the actual missing-certification error.

Fixes #1806

## Test plan
- [x] `go build ./platform/fabric/...`
- [x] `go test -run=FuzzTransactionSetFromEnvelopeBytes ./platform/fabric/core/generic/transaction -v` — all seeds and the crasher entry pass
- [x] `go test ./platform/fabric/services/state/...`
- [x] `go test ./platform/fabric/core/generic/transaction/...`

Note that a third instance of the same pattern, in `ordering.CFTBroadcaster.Broadcast`, is fixed separately in #1808


SaidAltury-ibm (approved):

PR #1808, opened 2026-09-10
refactor(fabric): simplify CFTBroadcaster retry loop, fix connection-pool leak

mbrandenburger:
## Summary
- `Broadcast` tracked retry state through a `forceConnect` flag and a `connection` variable reused across loop iterations, so a single logical step (get a connection, send, receive) was split across four different branches. Extracts the send/receive pair into a `sendAndRecv` helper and has each retry acquire its own connection, so the loop body reads as one attempt per iteration instead of a state machine.
- Fixes connection-pool resource leaks in `getConnection`: all four error returns after a successful `connSem.Acquire` (no orderer configured, client creation, broadcast-client creation, stream creation) never released the acquired semaphore slot, and the broadcast-client-creation failure additionally never closed the already-created client, unlike its sibling stream-creation failure a few lines below. Each lost slot permanently shrinks the pool, and enough connection failures would eventually deadlock every future `Broadcast` call waiting on the semaphore.
- Adds a regression test (`TestCFTBroadcaster_GetConnection_SemaphoreNotLeakedOnFailure`) that drives `poolSize` consecutive failures of each of the four `getConnection` failure kinds and confirms the pool still yields `poolSize` semaphore slots afterward. Verified it fails against the pre-fix `getConnection`.
- Adds a regression test (`TestCFTBroadcaster_Broadcast_ZeroRetries`) for the zero-retries case: the old code left its error variable uninitialized when the retry loop never ran, and `errors.Wrap(nil, ...)` returns `nil`, so `Broadcast` silently reported success without ever attempting to send.

## Test plan
- [x] `go build ./platform/fabric/...`
- [x] `go vet ./platform/fabric/core/generic/ordering/...`
- [x] `go test ./platform/fabric/core/generic/ordering/... -race -count=1`

### DIFF
--- a/platform/fabric/core/generic/ordering/cft_test.go
+++ b/platform/fabric/core/generic/ordering/cft_test.go
@@ -34,10 +34,12 @@ func newTestMetrics() *metrics.Metrics {
 
 // fakeCFTServices provides orderer clients for testing
 type fakeCFTServices struct {
-	shouldFailClient bool
-	shouldFailStream bool
-	clientError      error
-	streamError      error
+	shouldFailClient    bool
+	shouldFailStream    bool
+	shouldFailBroadcast bool
+	clientError         error
+	streamError         error
+	broadcastError      error
 }
 
 func (f *fakeCFTServices) NewOrdererClient(grpc.ConnectionConfig) (Client, error) {
@@ -48,16 +50,20 @@ func (f *fakeCFTServices) NewOrdererClient(grpc.ConnectionConfig) (Client, error
 		return nil, errors.New("client creation failed")
 	}
 	return &fakeCFTClient{
-		shouldFailStream: f.shouldFailStream,
-		streamError:      f.streamError,
+		shouldFailStream:    f.shouldFailStream,
+		shouldFailBroadcast: f.shouldFailBroadcast,
+		streamError:         f.streamError,
+		broadcastError:      f.broadcastError,
 	}, nil
 }
 
 type fakeCFTClient struct {
 	Client
-	shouldFailStream bool
-	streamError      error
-	closed           bool
+	shouldFailStream    bool
+	shouldFailBroadcast bool
+	streamError         error
+	broadcastError      error
+	closed              bool
 }
 
 func (f *fakeCFTClient) OrdererClient() (ab.AtomicBroadcastClient, error) {
@@ -67,16 +73,28 @@ func (f *fakeCFTClient) OrdererClient() (ab.AtomicBroadcastClient, error) {
 		}
 		return nil, errors.New("stream creation failed")
 	}
-	return &fakeCFTAB{}, nil
+	return &fakeCFTAB{
+		shouldFailBroadcast: f.shouldFailBroadcast,
+		broadcastError:      f.broadcastError,
+	}, nil
 }
 
 func (f *fakeCFTClient) Close() {
 	f.closed = true
 }
 
-type fakeCFTAB struct{}
+type fakeCFTAB struct {
+	shouldFailBroadcast bool
+	broadcastError      error
+}
 
-func (fakeCFTAB) Broadcast(context.Context, ...ggrpc.CallOption) (ggrpc.BidiStreamingClient[common.Envelope, ab.BroadcastResponse], error) {
+func (f fakeCFTAB) Broadcast(context.Context, ...ggrpc.CallOption) (ggrpc.BidiStreamingClient[common.Envelope, ab.BroadcastResponse], error) {
+	if f.shouldFailBroadcast {
+		if f.broadcastError != nil {
+			return nil, f.broadcastError
+		}
+		return nil, errors.New("broadcast stream creation failed")
+	}
 	return &fakeCFTStream{status: common.Status_SUCCESS}, nil
 }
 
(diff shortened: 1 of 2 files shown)


SaidAltury-ibm (commented):
thank you @mbrandenburger for this PR,
One request: **please add a regression test.** I stripped all four `o.connSem.Release(1)`
calls from your branch and **the entire existing suite still passed** — so the fix is
currently unguarded.

platform/fabric/core/generic/ordering/cft_test.go
```diff
@@ -451,6 +451,72 @@ func TestCFTBroadcaster_GetConnection(t *testing.T) {
...
+			hasOrderer: false,
+		},
+		{
+			name:       "client creation fails",
+			services:   &fakeCFTServices{shouldFailClient: true, clientError: errors.New("connection refused")},
+			hasOrderer: true,
+		},
+		{
+			name:       "broadcast client creation fails",
+			services:   &fakeCFTServices{shouldFailStream: true, streamError: errors.New("stream error")},
```
adecaro, line 533: **Should-fix: this test table doesn't cover the fourth semaphore-leak branch — the `oClient.Broadcast()` (stream-creation) failure at `cft.go:145-150`.**

`fakeCFTAB.Broadcast` always returns success (`cft_test.go:79-81`), so nothing in this file ever drives that branch. I confirmed the gap is real: removing just the `o.connSem.Release(1)` on that branch and re-running the full package suite (`go test ./platform/fabric/core/generic/ordering/... -race`) still passes, and `go tool cover -func` shows lines 147-150 at 0 hits.

Failure scenario: a future refactor of `getConnection` drops or misplaces that specific `Release(1)` and CI stays green — silently reintroducing the leak for this one path.

Fix: add a fourth case here (e.g. a `shouldFailBroadcast`/`broadcastError` field threaded through `fakeCFTClient`/`fakeCFTAB` so `Broadcast(...)` can return an error) alongside `no orderer configured`, `client creation fails`, and `broadcast client creation fails`.

adecaro (commented):
## Review: does this address the reported issue?

(No issue is linked to this PR — see Non-blocking below.)

Check out at `f30d6f348fd95b529768f9d6a3e98cf50f3b9a3e`.

```
go test ./platform/fabric/core/generic/ordering/... -race -count=1
ok  	github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/ordering	2.6s
```

```
# Reverted cft.go to pre-fix (842d31b8), kept the new test:
go test -run TestCFTBroadcaster_GetConnection_SemaphoreNotLeakedOnFailure -v ./platform/fabric/core/generic/ordering/...
--- FAIL (all 3 subtests): context deadline exceeded acquiring the semaphore
# Confirms the leak was real and the new test catches it.
```

```
# Removed only the Release(1) in the *fourth* leak branch (stream-creation failure) and reran the full suite:
go test ./platform/fabric/core/generic/ordering/... -race
ok  (all pass) -- this branch's fix is unguarded by any test; go tool cover confirms 0 hits on cft.go:147-150
```

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Simplifies `Broadcast`'s retry loop (state-machine flags → one attempt per iteration) | Read `cft.go:45-89`; behaviorally equivalent to pre-fix logic (old `forceConnect`/`i>0` condition was always true on every retry anyway) |
| Fixes 4 semaphore-leak paths in `getConnection` | Diff shows `Release(1)` added to all 4 `return nil, err` paths after `Acquire`; reproduced the pre-fix leak by reverting `cft.go` and re-running the new test |
| Adds a regression test that fails against the pre-fix code | Confirmed by revert-test (see above) — but only 3 of the 4 leak paths are exercised; see Non-blocking below |

### Non-blocking

- **Should-fix**: the semaphore-release fix on the `oClient.Broadcast()` stream-creation-failure branch (`cft.go:145-150`) has zero test coverage. I verified this by deleting just that `Release(1)` line and re-running the full package test suite — everything still passed, and `go tool cover -func` shows those lines at 0 hits. A future edit that drops this specific release call would go unnoticed by CI. Suggest adding a 4th case to `TestCFTBroadcaster_GetConnection_SemaphoreNotLeakedOnFailure` where `fakeCFTAB.Broadcast(...)` returns an error (posted inline).
- No issue is linked to this PR (bot's own `Issue Link` check already flags this as failing) — worth adding a `Fixes #...` reference or an explicit issue link per `docs/dev/workflow.md` before merge.

### Recommendation

The refactor and the core fix are solid and verified. Recommend closing the test-coverage gap on the fourth leak branch (and linking an issue) before merge; otherwise this looks ready.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

adecaro (approved):
LGTM

SaidAltury-ibm (approved):
LGTM

platform/fabric/core/generic/ordering/cft_test.go
mbrandenburger: Added a `shouldFailBroadcast`/`broadcastError` field threaded through `fakeCFTClient`/`fakeCFTAB` and a fourth "broadcast stream creation fails" case, in `547923ae`. Verified it fails when that specific `Release(1)` is reintroduced.

PR #1809, opened 2026-09-10
test: use require in tests where a failure matters

mbrandenburger:
Converts `testify/assert` to `testify/require` wherever a failed check would make the rest of the test meaningless: a later line indexes, dereferences, or type-asserts the value just checked, or a later step only makes sense once this one succeeded. Also folds in the handful of manual `if err != nil { t.Fatalf(...) }` patterns that are really the same check spelled out by hand, converting those to `require.NoError` too.

### Scoping

72 files in the repo import `testify/assert`. 12 changed; the rest already split correctly, or their `assert.X` calls exist for a specific reason:
- **Off the test's main goroutine.** `require`'s `t.FailNow()` calls `runtime.Goexit()`, which is unsafe from anywhere but the test's own goroutine. A call inside a spawned `go func()`, `wg.Go()`, or an `httptest` handler closure has to stay `assert`.
- **Inside an `EventuallyWithT` callback.** Its signature is fixed as `func(c *assert.CollectT)` — testify never defined a `require.CollectT` — so those calls stay `assert.X` no matter which outer function, `assert.EventuallyWithT` or `require.EventuallyWithT`, was used.
- **Genuinely independent checks**, mostly table-driven tests verifying several unrelated fields, where seeing every failure in one run beats stopping at the first.

Wrote the resulting rule down in a new `docs/dev/testing.md` (a testing best-practices doc), linked from `docs/agents/testing.md` and from `AGENTS.md`'s "Where to look next" table, so it outlives this PR.

### Verification

`go build`, `go vet`, and a targeted `go test` across every touched package pass clean in both the root and `integration` modules.

### Note for reviewers

This PR carries only the `testify` changes and the new testing-best-practices doc. It's cut from a larger internal cleanup pass that also retired a couple of unrelated helper functions; those aren't included here since they're unrelated to `assert`/`require`.

### DIFF
--- a/platform/common/utils/cache/timeout_test.go
+++ b/platform/common/utils/cache/timeout_test.go
@@ -46,25 +46,25 @@ func TestTimeoutSimple(t *testing.T) {
 	}
 
 	mu.RLock()
-	assert.Empty(t, allEvicted)
-	assert.Equal(t, 5, c.Len())
+	require.Empty(t, allEvicted)
+	require.Equal(t, 5, c.Len())
 	for k, expected := range input {
 		actual, _ := c.Get(k)
-		assert.Equal(t, expected, actual)
+		require.Equal(t, expected, actual)
 	}
 	mu.RUnlock()
 
-	assert.EventuallyWithT(t, func(a *assert.CollectT) {
+	require.EventuallyWithT(t, func(a *assert.CollectT) {
 		// eventually our cache is empty again due to eviction
 		assert.Equal(a, 0, c.Len())
 	}, timeout, tick)
 
 	mu.RLock()
-	assert.Equal(t, input, allEvicted)
+	require.Equal(t, input, allEvicted)
 	mu.RUnlock()
 
 	_, ok := c.Get(1)
-	assert.False(t, ok)
+	require.False(t, ok)
 }
 
 func TestTimeoutParallel(t *testing.T) {
@@ -89,18 +89,18 @@ func TestTimeoutParallel(t *testing.T) {
 	initialLen := c.Len()
 	initialEvicted := int(evictedCount.Load())
 
-	assert.GreaterOrEqual(t, initialLen, 0)
-	assert.LessOrEqual(t, initialLen, numItem)
-	assert.GreaterOrEqual(t, initialEvicted, 0)
-	assert.LessOrEqual(t, initialEvicted, numItem)
+	require.GreaterOrEqual(t, initialLen, 0)
+	require.LessOrEqual(t, initialLen, numItem)
+	require.GreaterOrEqual(t, initialEvicted, 0)
+	require.LessOrEqual(t, initialEvicted, numItem)
 
-	assert.EventuallyWithT(t, func(a *assert.CollectT) {
+	require.EventuallyWithT(t, func(a *assert.CollectT) {
 		// eventually our cache is empty again due to eviction
 		assert.Equal(a, 0, c.Len())
 	}, timeout, tick)
 
 	// once empty evictedCount should match the number of items we cached before
-	assert.Equal(t, numItem, int(evictedCount.Load()))
+	require.Equal(t, numItem, int(evictedCount.Load()))
 }
 
 // TestTimeoutCache_CleanupGoroutineStopsOnCancel asserts the background eviction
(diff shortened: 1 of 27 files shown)


platform/common/utils/cache/timeout_test.go
```diff
@@ -46,25 +46,25 @@ func TestTimeoutSimple(t *testing.T) {
...
-		assert.Equal(t, expected, actual)
+		require.Equal(t, expected, actual)
 	}
 	mu.RUnlock()
 
-	assert.EventuallyWithT(t, func(a *assert.CollectT) {
+	require.EventuallyWithT(t, func(a *assert.CollectT) {
 		// eventually our cache is empty again due to eviction
-		assert.Equal(a, 0, c.Len())
+		require.Equal(a, 0, c.Len())
```
SaidAltury-ibm: Converting the outer assert.EventuallyWithT → require.EventuallyWithT is right.
The inner require is what docs/dev/testing.md in this same PR says not to do:

`The call sits inside an EventuallyWithT callback. Its signature is fixed as func(c *assert.CollectT) — testify has no require.CollectT — so those calls stay assert.X no matter which outer function you used.`

platform/common/utils/cache/timeout_test.go
```diff
@@ -89,18 +89,18 @@ func TestTimeoutParallel(t *testing.T) {
...
+	require.GreaterOrEqual(t, initialLen, 0)
+	require.LessOrEqual(t, initialLen, numItem)
+	require.GreaterOrEqual(t, initialEvicted, 0)
+	require.LessOrEqual(t, initialEvicted, numItem)
 
-	assert.EventuallyWithT(t, func(a *assert.CollectT) {
+	require.EventuallyWithT(t, func(a *assert.CollectT) {
 		// eventually our cache is empty again due to eviction
-		assert.Equal(a, 0, c.Len())
+		require.Equal(a, 0, c.Len())
```
SaidAltury-ibm: Converting the outer assert.EventuallyWithT → require.EventuallyWithT is right.
The inner require is what docs/dev/testing.md in this same PR says not to do:

`The call sits inside an EventuallyWithT callback. Its signature is fixed as func(c *assert.CollectT) — testify has no require.CollectT — so those calls stay assert.X no matter which outer function you used.`

SaidAltury-ibm (changes requested):
Thank you @mbrandenburger for this PR, Nice work!
I have left you 2 comments please take a look

SaidAltury-ibm (approved):

adecaro (commented):
## Review: does this address the PR's stated intent (no issue linked)?

Checked out at `1fbe950bf09af094902de2f664e548ad9d24a307`.

```
$ go build ./... && go vet ./...   # root + integration module
clean

$ go test ./platform/... (touched packages) ./integration/nwo/... (touched packages)
ok  x24 packages, no failures

$ git merge --no-commit --no-ff origin/main
CONFLICT (content): platform/fabric/sdk/dig/sdk_test.go   # trivial, same file also touched by #1796 on main
```

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Converts `assert`→`require` only where a later statement dereferences/indexes/type-asserts the checked value, or the check gates a dependent call | Read every hunk (27 files); spot-checked `ds_test.go` (`b.SW.Hash` deref), `provider_test.go` (`.( bool)` assertion), `service_extended_test.go` (`ArgsForCall(0)` indexing) — all load-bearing |
| Folds hand-written `if err != nil { t.Fatal(...) }` into `require.NoError` | `kvstore_test_utils.go`, `grpc/server_test.go` |
| Leaves `assert` for independent checks and `EventuallyWithT`/`CollectT` callbacks | Confirmed in `timeout_test.go` — both `EventuallyWithT` outer calls are now `require`, inner `assert.Equal(a, ...)` callbacks are unchanged (correct — `CollectT` has no `require` counterpart) |
| New `docs/dev/testing.md` rule matches the code changes | Read section in full, no gaps found |

Note for the record: @SaidAltury-ibm's earlier comments about an inner `require` inside `EventuallyWithT` in `timeout_test.go` no longer apply to the current head — both call sites already keep the inner check as `assert.Equal(a, ...)`.

### Non-blocking

- No issue is linked (`issueLink` bot check: `FAIL — no_issue_linked`).
- `mergeStateStatus` is `DIRTY` — `main`'s #1796 touched the same file (`platform/fabric/sdk/dig/sdk_test.go`), producing a trivial one-file conflict. Needs a rebase before this can merge.
- Nit: `grpc/config_test.go`'s three converted `Equal` checks aren't strictly followed by a dependent dereference; a defensible read either way, not worth reworking.

### Recommendation

The conversions themselves are correct and every touched package builds/vets/tests clean. Needs a rebase (trivial conflict) and, ideally, a linked issue before merge; no design or correctness concerns in the diff itself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

adecaro (approved):
LGTM

PR #1810, opened 2026-09-10
fix(fabric): bound the remote finality probe by the caller's context

### LINKED ISSUE #1735: `IsFinal` ignores the caller's context during the remote finality probe
### What happened?

`Committer.IsFinal(ctx, txID)` documents that the caller's context sets the deadline for the whole wait. Stage 2 of its cascade — the remote peer probe — breaks that: it builds its own context from `context.Background()`. A caller that asks for 30s can be held for the full `FinalityWaitTimeout` (20s default, commonly raised to 120s in deployments), and cancelling the caller's context does not stop the probe.

The documented contract (`platform/fabric/core/generic/committer/committer.go:368-371`):

```go
// IsFinal takes in input a transaction id and waits for its confirmation
// with the respect to the passed context that can be used to set a deadline
// for the waiting time.
func (c *Committer) IsFinal(ctx context.Context, txID string) error {
```

Stage 1 (`c.Status(ctx, …)`, `committer.go:375`) and stage 3 (`listenTo`, which selects on `ctx.Done()`, `committer.go:494-497`) both honour it. Stage 2 does not — `FabricFinality.IsFinal` takes no context at all (`platform/fabric/core/generic/finality/fabric.go:64`):

```go
func (d *FabricFinality) IsFinal(txID, address string) error {
```

…and manufactures one (`platform/fabric/core/generic/finality/fabric.go:81`):

```go
ctx, cancelFunc = context.WithTimeout(context.Background(), d.WaitForEventTimeout)
```

That context bounds the deliver stream and `DeliverWaitForResponse` (`platform/fabric/core/generic/delivery/deliverclient.go:255-266`), so the probe runs its own full budget regardless of what the caller asked for. The interface the committer holds has no room for a context either (`committer.go:58-60`):

```go
type FabricFinality interface {
	IsFinal(txID, address string) error
}
```

Impact:

- **Caller deadlines are not respected.** A caller with a 60s context, on a channel configured `finality.waitForEventTimeout: 120s`, blocks for ~120s and then reports a deadline error that looks like a slow orderer. This is what makes a caller-side timeout shorter than `finality.waitForEventTimeout` behave counter-intuitively rather than simply failing early.
- **Timeout budgets add rather than `max()`.** Worst normal case for one `IsFinal` call is one full probe **plus** one full `listenTo`: `FinalityWaitTimeout + CommitterWaitForEventTimeout`. At FSC defaults that is 20s + 300s = **320s**, already above the 5-minute default budget the Panurus/token-SDK `finalityView` allows for everything beneath it. Any sizing that treats the three timeouts as alternatives rather than a sum will under-provision. (Three probes are possible but not the normal path: `committer.go:406-409` returns after the first failed probe whenever the vault is still `Unknown`; a second probe needs the vault status to have moved, or the vault read itself to error.)
- No goroutine or connection leak — `defer client.Close()` and `defer cancelFunc()` (`fabric.go:74`, `:82`) unwind the `DeliverReceive` goroutine spawned at `fabric.go:112`. The problem is latency and contract, not resources.

### How can we reproduce it?

1. Configure a channel with `finality.waitForEventTimeout: 120s`.
2. Call `ch.Finality().IsFinal(ctx, txID)` with `ctx` carrying a 5s deadline, for a transaction ID the local vault does not know so that stage 1 returns `Unknown` and stage 2 is entered (`committer.go:391`).
3. Ensure the transaction never appears on the peer's deliver stream — use an unknown/never-submitted transaction ID.
4. Measure the call. It returns after ~120s, not after ~5s. Cancelling `ctx` early does not shorten it.

As a unit test, `platform/fabric/core/generic/finality/fabric_test.go` and the committer's `fake` package already support this: call `Committer.IsFinal` with a context cancelled after 100ms against a fake peer that never delivers the transaction, and assert the call returns promptly rather than after `WaitForEventTimeout`.

### Environment

- Operating system (e.g., Linux, MacOS): MacOS (platform-independent — context plumbing, not environment-specific)

### Additional Information

Found by code inspection while verifying a third-party analysis of the finality feature against the implementation, not by observing it at runtime. The reproduction steps above are derived, not executed. The analysis had independently flagged a downstream caller that hardcodes a 1-minute context and noted it expires "before the committer's own event-loop timeout"; the underlying cause is this defect, and it is worse than described there, since the probe does not merely outlast the caller's deadline — it never observes it.

Suggested fix — thread the context through. Mechanical, contained to one interface and its implementations:

1. Change the interface to `FabricFinality.IsFinal(ctx context.Context, txID, address string) error` (`committer/committer.go:58`), plus the mocks and fakes in `committer/fake` and `finality/fake` (`make generate-mocks`).
2. In `finality/fabric.go:81`, derive from the caller instead of `context.Background()`: `ctx, cancelFunc = context.WithTimeout(ctx, d.WaitForEventTimeout)` — keeping the probe's own ceiling while letting an earlier caller deadline win.
3. Pass `ctx` at the call site (`committer/committer.go:401`).

Worth doing in the same change as #1732, which touches the same two lines.

Separately, the sizing consequence deserves a note in `docs/configuration.md` beside the three timeout keys: they compose additively, so a caller's context must be at least `finality.waitForEventTimeout + committer.waitForEventTimeout` for a slow-but-succeeding transaction to be reported as final rather than as a deadline error.

Related:

- #1732 — same function and call site.
- #1733 — also inflates time spent in the stage-2 path.
- `platform/fabricx` has no equivalent: `fabricx/core/channel/wrappers.go:66-120` is a single `select` on the listener channel versus `ctx.Done()`, with no retry loop and no manufactured context.

SaidAltury-ibm:
### Description

`Committer.IsFinal(ctx, txID)` documents the caller's context as bounding the whole wait, and its local status check and its `listenTo` wait both honour it. The remote peer probe did not: `FabricFinality.IsFinal` built its own context from `context.Background()` (`finality/fabric.go:89`), so the probe ran its full `finality.waitForEventTimeout` regardless of the caller's deadline, and cancelling the caller's context did not reach it. A caller with a 5s deadline on a channel configured `waitForEventTimeout: 120s` blocked for ~120s and then returned a deadline error indistinguishable from a slow orderer.

- Threads the context through `committer.FabricFinality` (`committer.go:59`), its call site (`committer.go:398`), the implementation, and the test fake, and derives the probe's context from it: `context.WithTimeout(ctx, d.WaitForEventTimeout)`. `WaitForEventTimeout` stays a ceiling that an earlier caller deadline wins against.
- The derived context reaches both the Deliver stream (`NewDeliver`/`NewDeliverFiltered`, which pass it to the gRPC call) and `DeliverWaitForResponse`, so a cancelled caller tears the stream down instead of leaving it running to its own budget.
- `docs/configuration.md`: records that the probe and the local event wait run in sequence, so `finality.waitForEventTimeout` and `committer.waitForEventTimeout` add up rather than being alternatives — 320s at the defaults. Sizing a caller's context as if they were alternatives under-provisions it.

Compatibility: `committer.FabricFinality` is exported but consumed only in-tree — `channelprovider.go`, the committer itself, the dig provider at `providers.go:149`, and a test fake. The public `driver.Finality.IsFinal(ctx, txID)` already carries a context and is untouched, so this brings the internal seam into line with it. Panurus does not reference `FabricFinality`; it reaches finality through `ch.Finality()`.

Closes #1735.

### Test plan

- [x] `finality/fabric_test.go`: a peer that accepts the seek and then goes quiet, `WaitForEventTimeout` at 30s, caller context at 100ms — asserts the call returns in under 5s.
- [x] Mutation-checked: restoring `context.Background()` on that one line makes the same test take **30.00s** while ignoring the 100ms deadline, so the assertion is load-bearing.
- [x] `-count=5` on the new test — stable; the 5s bound against a 100ms deadline leaves margin for loaded CI.
- [x] `go test -race` across `committer`, `finality`, `delivery`, and `platform/fabricx/...` (which has its own `IsFinal` implementations) — all pass.
- [x] `integration/` module builds (separate `go.mod`).
- [x] `golangci-lint run platform/fabric/...` — 0 issues.
- [x] Full `platform/fabric/...` passes except the two Postgres tests needing a local Docker daemon, which fail identically on an unmodified tree.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -583,6 +583,10 @@ fabric:
         numRetries: 3 # number of retries on a chaincode operation failure
         retrySleep: 1s # waiting time before retry again a failed chaincode operation
         # section about the finality service
+        # The remote peer probe and the local event wait run in sequence, so
+        # finality.waitForEventTimeout and committer.waitForEventTimeout add up
+        # rather than being alternatives: size a caller's context for the sum.
+        # Both are ceilings under the caller's context, which wins when shorter.
         finality:
           waitForEventTimeout: 20s
           forPartiesWaitTimeout: 1m
--- a/platform/fabric/core/generic/committer/committer.go
+++ b/platform/fabric/core/generic/committer/committer.go
@@ -56,7 +56,7 @@ type (
 )
 
 type FabricFinality interface {
-	IsFinal(txID string) error
+	IsFinal(ctx context.Context, txID string) error
 }
 
 type CommitTx struct {
@@ -395,7 +395,7 @@ func (c *Committer) IsFinal(ctx context.Context, txID string) error {
 			if logger.IsEnabledFor(zapcore.DebugLevel) {
 				c.logger.Debugf("Tx [%s] is unknown with no deps, remote check [%d][%s]", txID, iter, debug.Stack())
 			}
-			err := c.FabricFinality.IsFinal(txID)
+			err := c.FabricFinality.IsFinal(ctx, txID)
 			if err == nil {
 				c.logger.Debugf("Tx [%s] is final, remote check succeeded", txID)
 				return nil
--- a/platform/fabric/core/generic/committer/fake/fakes.go
+++ b/platform/fabric/core/generic/committer/fake/fakes.go
@@ -310,7 +310,7 @@ type FabricFinality struct {
 	Err error
 }
 
-func (f *FabricFinality) IsFinal(string) error {
+func (f *FabricFinality) IsFinal(context.Context, string) error {
 	return f.Err
 }
 
--- a/platform/fabric/core/generic/finality/fabric.go
+++ b/platform/fabric/core/generic/finality/fabric.go
@@ -63,10 +63,12 @@ func NewFabricFinality(
 
 // IsFinal probes a peer configured for finality for the given transaction,
 // returning an error if no peer is configured for that role.
-func (d *FabricFinality) IsFinal(txID string) error {
+//
+// ctx bounds the whole probe: WaitForEventTimeout applies as a ceiling on top of
+// it, so an earlier caller deadline wins.
+func (d *FabricFinality) IsFinal(ctx context.Context, txID string) error {
 	d.Logger.Debugf("remote checking if transaction [%s] is final in channel [%s]", txID, d.Channel)
 	var eventCh chan delivery.TxEvent
-	var ctx context.Context
 	var cancelFunc context.CancelFunc
 
 	peerConnConf := d.ConfigService.PickPeer(driver.PeerForFinality)
@@ -86,7 +88,7 @@ func (d *FabricFinality) IsFinal(txID string) error {
 		return errors.WithMessagef(err, "failed creating deliver client for address [%s]", address)
 	}
 
-	ctx, cancelFunc = context.WithTimeout(context.Background(), d.WaitForEventTimeout)
+	ctx, cancelFunc = context.WithTimeout(ctx, d.WaitForEventTimeout)
 	defer cancelFunc()
 	var deliverStream delivery.DeliverFiltered
 	if d.useFiltered {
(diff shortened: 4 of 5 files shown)


mbrandenburger:
cc @mffrench

mbrandenburger (approved):
Thank you @SaidAltury-ibm for creating this PR!

PR #1813, opened 2026-09-11
refactor(grpc)!: unify gRPC/TLS config

### LINKED ISSUE #1799: fabricx: committer query-service gRPC client has no way to configure max message size (hits gRPC's 4MB default)
When querying states through the fabric-x committer's query service, large result sets can exceed gRPC's default 4 MiB receive limit:

```
failed getting state: get states: query get rows: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5154640 vs. 4194304)
```

The client connection for `QueryServiceClient`/`NotificationServiceClient` is built in `platform/fabricx/core/committer/grpc/grpc.go` (`ClientConn`), which sets TLS credentials and a connection timeout but never sets `grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(...))` / `MaxCallSendMsgSize`. The endpoint config in `platform/fabricx/core/committer/config/config.go` (`Config`/`Endpoint`) exposes `Address`, `ConnectionTimeout`, and `TLS` only — there's no yaml key or field to override the message size.

For contrast, the general-purpose FSC gRPC client (`platform/view/services/grpc`) already defaults to 100 MB (`config.go:24-25`) and exposes `SetMaxRecvMsgSize`/`SetMaxSendMsgSize` (`client.go:151-159`, applied at `client.go:199-202`).

**Request**: add a configurable max receive/send message size to the fabricx committer's query-service `Endpoint` config (mirroring the `platform/view/services/grpc` pattern), or at minimum raise the hardcoded default so it doesn't silently fall back to gRPC's 4 MiB limit.

**Workaround for now**: none via config — requires a code change to `ClientConn()` in `platform/fabricx/core/committer/grpc/grpc.go`.

mbrandenburger:
Fabric-x's committer query-service client had no way to raise gRPC's 4 MiB default receive limit, so large result sets failed with `ResourceExhausted` (#1799). Fixing that surfaced a second, hand-rolled gRPC client in `platform/fabricx/core/committer/grpc` duplicating most of `platform/view/services/grpc`, with fewer knobs and its own TLS credentials builder that existed only to pin TLS 1.3.

This unifies both:
- Moves `platform/view/services/grpc` to `platform/common/services/grpc` and `platform/view/services/tlsconfig` to `platform/common/services/tlsconfig`, since fabric, fabricx, and view all consume them. No compatibility alias remains at the old paths.
- Adds configurable `MaxRecvMsgSize`/`MaxSendMsgSize` to `ServerConfig`, `ClientConfig`, and `ConnectionConfig`. Zero still falls back to the existing 100 MiB package default.
- Adds a configurable TLS version range (`MinVersion`/`MaxVersion`) to `SecureOptions` and the `ServerTLS`/`ClientTLS` wire types, with the same nil-inherits rule as every other TLS field.
- Deletes `platform/fabricx/core/committer/grpc`. Its endpoints become `commongrpc.ConnectionConfig`, and `ClientProvider` moves to `platform/fabricx/core/committer/config` on `lazy.Provider`, replacing its own two `sync.Map`s and hand-rolled locking. `NewClientProvider` and `ServiceConfigProvider` now live in that package.
- Fabric-x's TLS 1.3 requirement, previously a forked `TransportCredentials`, is now `minVersion: 772` in its own `tls:` block.

New configuration keys:
- `fsc.grpc.maxRecvMsgSize` / `maxSendMsgSize`
- Per-endpoint `maxRecvMsgSize` / `maxSendMsgSize` (Fabric peers/orderers, fabricx query/notification services)
- `minVersion` / `maxVersion` in any `tls:` block (771 = TLS 1.2, 772 = TLS 1.3)

No existing default changes: an unset client still resolves to TLS 1.2-1.3 and 100 MiB message limits; an unset listener still negotiates TLS 1.2 exactly, as it always has.

## Test plan
- `make checks` clean across all five modules
- Unit tests cover the message-size fallback/override on all three config structs and the TLS version-range resolution and validation
- Integration suite needs Fabric binaries and Docker; left to CI, not run locally for this pass

Closes #1799


adecaro (commented):
## Review: does this address #1799?

Checked out at `a75d7f18cced71302998f881bd206df138e6e393` (up to date with `main`). Build/vet/lint/tests all clean; no stale references to the old `platform/view/services/grpc` import path anywhere in the repo.

### What it fixes / adds

| Claim | Verified how |
|---|---|
| Fabricx query/notification client had no way to configure max gRPC message size (#1799) | `platform/fabricx/core/committer/config/config.go` now embeds the shared `ConnectionConfig` (100 MiB default, `maxRecvMsgSize`/`maxSendMsgSize` yaml knobs). RPC-level tests in `client_test.go` actually trigger `ResourceExhausted` at the configured limit. |
| `platform/view/services/grpc` → `platform/common/services/grpc` move is complete | repo-wide grep, zero stale hits; DI wiring repoints correctly. |

### Non-blocking

- **Should-fix:** unifying onto the shared client silently changes fabricx's TLS floor. The deleted fabricx client hardcoded TLS 1.3 minimum; the shared client defaults to negotiable 1.2–1.3, and nothing re-asserts 1.3 for fabricx in code — `minVersion: 772` is shown in docs only as an example, not a required migration step. An upgrading deployment that doesn't add it silently loses the TLS 1.3 guarantee. Suggest defaulting fabricx's endpoint config to TLS 1.3 explicitly, or marking it required in the migration docs.
- **Nit:** the docs migration note uses changelog phrasing ("used to pin unconditionally") that CLAUDE.md's doc conventions ask to avoid.

### Recommendation

Solid refactor, genuinely closes #1799. Worth resolving the TLS-default point before this leaves draft, since it's a silent security-relevant change for existing fabricx deployments.

platform/common/services/grpc/server_test.go
```diff
@@ -687,6 +688,106 @@ func TestNewSecureGRPCServer(t *testing.T) {
...
+		require.NoError(t, conn.Handshake())
+		require.Equal(t, uint16(tls.VersionTLS12), conn.ConnectionState().Version)
+	})
+}
+
+// TestNewGRPCServerVersionRange pins the two rules server.go applies before building a
+// listener's TLS config: MaxVersion falls back through MinVersion (not straight to the
+// package default) so an explicit floor alone cannot invert the range, and the combined
+// range is validated at startup rather than left to fail every handshake later.
+func TestNewGRPCServerVersionRange(t *testing.T) {
```
SaidAltury-ibm, line 714: i think we should add a case where an unconfigured range is TLS 1.2 exactly and make sure that every deployment that configures no TLS version reaches this fallback

SaidAltury-ibm (commented):
Thank you @mbrandenburger for this great work, i have one comment regarding testing the new behaviour, please tell me WYT?

mbrandenburger:
Thank you @adecaro and @SaidAltury-ibm for the quick review. I've restored the previous defaults for fabric-x and added additional tests. 

I believe we should come back to this TLS default discussion in a follow up as I strongly believe we should not enforce a minTLS on "client-side" components for external services. On the other hand, als FSC server-side components, we could enforce 1.3. Should we go with this PR as is and then do a follow up on sane TLS defaults? WDYT? 

Update: Note that we had previous attempts related to TLS min/max versions https://github.com/hyperledger-labs/fabric-smart-client/issues/1297

SaidAltury-ibm (approved):
LGTM

PR #1814, opened 2026-09-11
test(grpc/logging): cover the interceptors, options and stream wrapper, and guard stream payload logging

### LINKED ISSUE #1800: test: increase coverage of platform/view/services/grpc/logging
### Description

Coverage for `platform/view/services/grpc/logging` sits at 5.8%. `fields.go` is fully covered by `fields_test.go`, but `server.go` and `context.go` have no tests at all, 12 of the package's 15 functions are at 0%, including both gRPC interceptors.

**Current state**

| File | Coverage |
|---|---|
| `server.go` | 0% (11 functions) |
| `context.go` | 0% (1 function) |
| `fields.go` | 100% |

### Implementation Steps

-> [ ] `server.go`: cover `UnaryServerInterceptor` and `StreamServerInterceptor`, the main entry points, including the logging emitted at each level and error propagation from the wrapped handler.
-> [ ] `server.go`: cover the option plumbing, `WithLeveler`, `WithPayloadLeveler`, `applyOptions` and the defaults it applies, plus `LevelerFunc`'s `Level` and `PayloadLevel`.
-> [ ] `server.go`: cover `getFields`, and the `serverStream` wrapper's `Context`, `SendMsg` and `RecvMsg`, including payload logging when enabled.
-> [ ] `context.go`: cover `WithFields` and its retrieval path.
-> [ ] Use plain testify rather than Ginkgo, in line with the direction for new unit tests in this repo.

### Additional Information

-> **Target package:** `platform/view/services/grpc/logging`
-> **Target:** >90% statements covered
-> **Parent:** #1566

sid200727:
Fixes #1800

The package sat at 5.8%: fields.go was covered by fields_test.go, but server.go and context.go had no tests at all, leaving 12 of its 15 functions at zero, including both gRPC interceptors.

Covers the unary and stream interceptors through a fake handler and an observed logger: the response and error paths, the status codes recorded on completion, the payload entries emitted either side of the call, and the fields threaded onto the handler's context. Adds the option plumbing (WithLeveler, WithPayloadLeveler, applyOptions and its defaults), getFields across its method-parsing, deadline and peer branches including a TLS peer certificate, WithFields, and the serverStream wrapper's Context, SendMsg and RecvMsg.

Plain testify throughout, no new Ginkgo specs. The package is now at 100%.


platform/view/services/grpc/logging/server_test.go
```diff
@@ -0,0 +1,483 @@
...
+			fields := fieldMap(getFields(t.Context(), tc.method))
+
+			if tc.expect == nil {
+				require.NotContains(t, fields, "grpc.service")
+				require.NotContains(t, fields, "grpc.method")
+
+				return
+			}
+			for k, v := range tc.expect {
+				require.Equal(t, v, fields[k])
```
mbrandenburger: I think this is a good example where to use `assert.Equal`. See https://github.com/hyperledger-labs/fabric-smart-client/blob/main/docs/dev/testing.md#choosing-assert-over-require

platform/common/services/grpc/logging/server_test.go
```diff
@@ -0,0 +1,483 @@
...
+	"google.golang.org/grpc/status"
+)
+
+// observedLogger returns a logger that records everything written to it,
+// including payload entries, which sit below debug.
+func observedLogger() (*zap.Logger, *observer.ObservedLogs) {
+	core, logs := observer.New(DefaultPayloadLevel)
+
+	return zap.New(core), logs
+}
```
mbrandenburger, line 38: Other tests in this package instantiate the observer inline. See https://github.com/sid200727/fabric-smart-client/blob/c25d3afcd98b9626fea81e8c80c0076d3cba9f2b/platform/view/services/grpc/logging/fields_test.go#L37

However, I believe using this helper method is the right way to do it. I suggest to this his helper in `fields_test.go` as well, so we have a consistent way to create the logger in these tests. WDYT?

platform/view/services/grpc/logging/server_test.go
```diff
@@ -0,0 +1,483 @@
...
+		WithLeveler(LevelerFunc(func(context.Context, string) zapcore.Level { return zapcore.ErrorLevel })),
+	)
+	require.Equal(t, zapcore.ErrorLevel, o.Level(t.Context(), "/svc/Method"))
+	require.Equal(t, DefaultPayloadLevel, o.PayloadLevel(t.Context(), "/svc/Method"), "payload leveler keeps its default")
+
+	o = applyOptions(
+		WithPayloadLeveler(LevelerFunc(func(context.Context, string) zapcore.Level { return zapcore.InfoLevel })),
+	)
+	require.Equal(t, zapcore.DebugLevel, o.Level(t.Context(), "/svc/Method"), "leveler keeps its default")
+	require.Equal(t, zapcore.InfoLevel, o.PayloadLevel(t.Context(), "/svc/Method"))
```
mbrandenburger: We could use sub tests here as we covering two test cases here.

platform/view/services/grpc/logging/server_test.go
```diff
@@ -0,0 +1,483 @@
...
+
+	return f.sendErr
+}
+
+func (f *fakeServerStream) RecvMsg(m any) error {
+	if f.recvFill != nil {
+		f.recvFill(m)
+	}
+
+	return f.recvErr
```
mbrandenburger: I am wondering if this is correct behavior, if we set the an error of our fake, I  believe m should not be filled. Might this be a bug from the production code of the server stream?

mbrandenburger (changes requested):
Thank you @sid200727 for working on this. This PR is in a good shape! We few suggestions below.

platform/view/services/grpc/logging/server_test.go
sid200727: Good catch, switching to assert.Equal and thanks for the pointer to the doc, I hadn't seen that section.

platform/common/services/grpc/logging/server_test.go
sid200727: Agreed, that's better than the two files constructing loggers differently. Moving it somewhere both can use and updating fields_test.go to match.

platform/view/services/grpc/logging/server_test.go
sid200727: Will do, splitting those into t.Run blocks so a failure names which case broke.

platform/view/services/grpc/logging/server_test.go
sid200727: You're right, and I think it points at a real inconsistency rather than just an unrealistic fake.

serverStream.RecvMsg logs the message regardless of the error:

```go
err := ss.ServerStream.RecvMsg(msg)
if ce := ss.payloadLogger.Check(ss.payloadLevel, "received stream message"); ce != nil {
    ce.Write(ProtoMessage("message", msg))
}
return err
```

So on a failed receive it logs whatever is in msg a zero value, or stale data from a previous call on a reused message. The unary path guards against exactly this on the way out: if ce := ...; ce != nil && err == nil. The stream path doesn't, on either SendMsg or RecvMsg.
My fake filling the message and returning an error is unrealistic, and that's what obscured it, a real stream wouldn't populate msg on a failed receive. I'll make the fake behave properly (fill or fail, not both).

On the production side, do you want the err == nil guard added here to match the unary path? 
Happy to include it in this PR, or open a separate issue if you'd rather keep this one to tests.

mbrandenburger: Thank you @sid200727 - yes please add the `err == nil` guard in this PR.

mbrandenburger:
Thank you @sid200727 - this looks almost ready to merge. Can you please rebase and resolve the conflicts.

mbrandenburger (approved):

PR #1815, opened 2026-09-11
test(metrics/operations): replace the dead spec suite with real tests

### LINKED ISSUE #1801: test: restore the commented-out test suite in platform/view/services/metrics/operations
### Description

`platform/view/services/metrics/operations` reports 0% coverage. `system_test.go` is 331 lines with its entire suite commented out, the Ginkgo `Describe` block is still live and empty, so the suite passes in about a second while exercising nothing, and CI stays green.

This happened in #1017, when statsd support was dropped: the statsd-specific parts were deleted and the remainder was commented rather than removed. The commented specs can't be restored, though they target an API that no longer exists. `operations.NewSystem(options)` is now `NewOperationSystem(server, logger, metricsProvider, opts)`, and `Addr`, `RegisterHandler`, `RegisterChecker`, `Log` and the exported `Provider` field are all gone, as is the `fabhttp.Options` the test options embedded. They also depend on `fabhttp`, `healthz` and `fakes` types that were removed, `fakes/` is down to `handler.go`.

So this is fresh tests against the current API, with the dead specs deleted. Plain testify, in line with the direction for new unit tests in this repo.

**Note on scope**: the commented-out specs can't be restored or converted, since the API they exercise is gone. They should be deleted and replaced with fresh tests written in plain testify against the current API, in line with the direction for new unit tests in this repo.

### Implementation Steps

-> [ ] Delete the commented-out specs in `system_test.go` and the Ginkgo suite runner in `operations_suite_test.go`.
-> [ ] Cover `NewOperationSystem`, including the error path when the metrics provider fails to initialise.
-> [ ] Cover `System.Start` and `System.Stop`, including that `Stop` is safe when `Start` was never called.
-> [ ] Cover `initializeMetricsProvider` across its provider branches - `prometheus` (which registers the `/metrics` handler), the empty string (metrics disabled), and an unknown provider (disabled with a warning) - and the `RequireClientCert` flag passed through to handler registration.
-> [ ] Cover `initializeLoggingHandler`.
-> [ ] Cover `metrics.go` - `NewMetricsProvider`, `versionGauge`, `NewDisabledHistogram` and its `NewHistogram`.
-> [ ] Cover `logger.go` - `NewOperationsLogger`.

### Additional Information

- **Target package:** `platform/view/services/metrics/operations`
- **Parent:** #1566
- **Origin of the commented-out specs:** #1017 (drop statsd support)

sid200727:
Fixes #1801

The package reported 0% coverage. system_test.go held 331 lines of commented-out Ginkgo specs from #1017, where statsd support was dropped: the statsd parts were deleted and the rest was commented rather than removed. The Describe block stayed live and empty, so the suite passed in about a second while exercising nothing and CI stayed green.

Those specs could not be restored. They call operations.NewSystem(opts), Addr, RegisterHandler, RegisterChecker, Log and an exported Provider field, none of which exist now, and they depend on fabhttp, healthz and fakes types that were removed with them. They are deleted here, along with the Ginkgo suite runner, and replaced with plain testify tests against the current API.

Covers NewOperationSystem and the logspec handler it registers, the RequireClientCert flag threaded through handler registration, initializeMetricsProvider across its prometheus, disabled and unknown branches, Start, and Stop before Start, after it and twice over. Also covers metrics.go and logger.go, including the nil-logger fallback.

The package is at 97.1%. The remainder is NewOperationSystem's error branch, which wraps a failure from initializeMetricsProvider -- that function returns nil on every path today, so the branch is unreachable.

Plain testify throughout, no Ginkgo. `golangci-lint` (v2.13.2, the pinned version), `go vet` and `gofmt -l` all clean. No production code changed.


mbrandenburger (approved):
Thank you @sid200727 for working on the tests. Well done!

PR #1820, opened 2026-09-14
chore(lint): enable sqlclosecheck in golangci-lint

mbrandenburger:
## Summary
- Enable the `sqlclosecheck` linter, which flags `*sql.Rows`/`*sql.Stmt` that aren't closed via a directly-visible `defer ...Close()`.
- Fix the two flagged sites (`BindingStore.HaveSameBinding`, `SignerInfoStore.FilterExistingSigners`): both deferred `rows.Close()` through `utils.IgnoreErrorFunc`, an indirection the linter can't see through. Defer a closure that calls `rows.Close()` directly instead, which both `sqlclosecheck` and `errcheck` recognize, and drop the now-unneeded import.

This is part of https://github.com/hyperledger-labs/fabric-smart-client/issues/1755

## Test plan
- [x] `go build ./...`
- [x] `golangci-lint run --enable-only sqlclosecheck ./...` — 0 issues
- [x] `golangci-lint run --enable-only errcheck ./platform/view/services/storage/driver/sql/common/...` — 0 issues


SaidAltury-ibm (approved):

PR #1821, opened 2026-09-14
chore(lint): enable forcetypeassert in golangci-lint

mbrandenburger:
## Summary
- Enable the `forcetypeassert` linter, which flags unchecked type assertions (`x.(T)` without the `, ok` form).
- Fix every flagged production-code site with a checked assertion: `errors.Errorf` for error-returning functions, `panic(errors.Errorf(...))` for non-error-returning functions, restructured predicates for `if v, ok := x.(T); ok && ...` conditions, and collapsed a few redundant double-assertions into one.
- Add a `pkg/utils/proto.CloneOf` helper and use it in `platform/fabric/core/generic/discovery/client.go` in place of `proto.Clone(x).(*T)` + assertion.
- Exclude `forcetypeassert` for `integration/`, `_test.go`, `_test_utils.go`, and `fake/` — hand-written test doubles and the integration harness aren't attacker-reachable, so an unchecked assertion there is a test bug at worst.

This is part of https://github.com/hyperledger-labs/fabric-smart-client/issues/1755

## Test plan
- [x] `go build ./...`
- [x] `golangci-lint run --enable-only forcetypeassert ./...` — 0 issues
- [x] `make checks` across all 5 modules


SaidAltury-ibm (approved):
thank you @mbrandenburger , LGTM

PR #1829, opened 2026-09-14
refactor!: retire closer.go, replace with standard Go error discards

mbrandenburger:
## Summary
- `closer.go` existed only to hide an error return behind a name (`IgnoreError`, `IgnoreErrorWithOneArg`, `IgnoreErrorFunc`, `Closer`, `CloseMute`); `defer utils.IgnoreError(f())` also evaluated `f()` immediately when the `defer` ran rather than when the surrounding function returned, so cleanup fired at the wrong time
- Every call site is rewritten with the standard Go idiom for discarding an error: an inline `_ = expr()`, `defer func() { _ = expr() }()`, or `go func() { _ = expr() }()`, and `closer.go` is deleted along with its five now-unused helpers
- `csp_test.go` and `validators_test.go` used `IgnoreErrorWithOneArg`/`IgnoreError` only to clean up a manually created temp dir; removing the call let cleanup move to `t.TempDir()`/`t.Cleanup()` instead of a bare defer

## Test plan
- [x] `make checks` passes with 0 issues across all 4 modules
- [x] `go build ./...` passes in both the root and `integration` modules


SaidAltury-ibm (approved):
LGTM

PR #1837, opened 2026-09-15
test(logging): cover the context-aware methods, lazy types and replacer registry

### LINKED ISSUE #1833: test: increase coverage of platform/common/services/logging
### Description

Coverage for `platform/common/services/logging` sits at 57.9% once `mock.go` and `logging_test_utils.go` are filtered out of the denominator by `scripts/filter-coverage.sh`. `types.go` and the context-fields registry are well covered. Two areas are not: the logger construction path, and the context-aware logging methods in `otel.go`.

**Logger construction and the replacer registry**

|---|---|---|
| `RegisterReplacer` | `config.go` | 0% |
| `Replacers` | `config.go` | 0% |
| `ContextLogFields` | `config.go` | 66.7% |
| `MustGetLogger`, `GetLogger`, `GetLoggerWithReplacements`, `GetPackageName`, `NewSpecHandler`, `loggerName` | `logger.go` | 0% |

**Context-aware logging**

Each `*Context` method branches on whether its level is enabled: when it isn't, it delegates to the plain logger and returns; when it is, it emits a span event and attaches the context fields. The tests exercise one side of that branch, and three of the methods aren't called at all.

| Function | Coverage |
|---|---|
| `WarnfContext`, `WarnwContext`, `ErrorfContext` | 0% |
| `DebugfContext`, `DebugwContext`, `InfofContext`, `InfowContext`, `ErrorwContext` | 60% |
| `contextLogFieldArgs` | 35.7% |
| `withContextFields` | 66.7% |
| `appendContextLogFields` | 94.1% |

**Note on the replacer registry:** `replacers` is package-level mutable state and `RegisterReplacer` panics on a duplicate key, so tests that register have to restore it afterwards and can't run in parallel with anything calling `Replacers()`.

### Implementation Steps

- [ ] `config.go`: cover `RegisterReplacer`, including the panic on a duplicate key, and `Replacers`. Restore the registry between tests.
- [ ] `config.go`: close the remaining branch in `ContextLogFields`.
- [ ] `logger.go`: cover `loggerName` - package-path splitting, appended params, replacement application table-driven.
- [ ] `logger.go`: cover `GetLoggerWithReplacements`, `GetLogger` and `MustGetLogger`, asserting the resulting logger name reflects the replacements in force.
- [ ] `logger.go`: cover `GetPackageName` through the public entry points rather than directly, it reads `runtime.Caller(4)`, so a direct call sits at a different stack depth. Its two error branches aren't reachable from a normal call.
- [ ] `logger.go`: cover `NewSpecHandler`.
- [ ] `otel.go`: cover both branches of all eight `*Context` methods level enabled and level disabled table-driven over the level/method pairs.
- [ ] `otel.go`: cover `contextLogFieldArgs` across its exits: nil context, no registered specs, a registered spec absent from the context, and one present.
- [ ] `otel.go`: close the remaining branches in `withContextFields` and `appendContextLogFields`.

### Additional Information

- **Target package:** `platform/common/services/logging`
- **Parent:** #1566
- **Note:** 57.9% is the filtered figure. An unfiltered `go test -cover` reports 49.8%, because `mock.go` and `logging_test_utils.go` land in the denominator.

sid200727:
Fixes #1833

## Background

The package sat at 57.9% with `scripts/filter-coverage.sh` applied. `types.go`'s filtering helpers and the context-field registry were covered; the context-aware logging methods, the lazy `fmt.Stringer` helpers and the replacer registry were not.

## What this covers

**`otel.go`** - both branches of all eight level-guarded `*Context` methods: the enabled path that emits a span event and attaches the registered context fields, and the disabled path that delegates to the plain logger and returns. Plus `PanicfContext` and `PanicwContext`, the remaining exits in `contextLogFieldArgs`, and `appendContextLogFields` including its promise that the caller's `keysAndValues` slice is never mutated.

**`types.go`** - the lazy helpers: `Keys`, `Identifier` across its nil, pointer and value shapes, `Eval` (checking the function isn't called until rendered), `Since`, `SHA256Base64` and `Base64`.

**`config.go` / `logger.go`** - `RegisterReplacer` including the panic on a duplicate key, `Replacers`, `loggerName` table-driven over its path/params/replacement handling, and `NewSpecHandler`.

The package is now at **91.4%**.

## Not covered, and why

`GetLogger`, `MustGetLogger`, `GetLoggerWithReplacements` and `GetPackageName` are left uncovered.

`GetPackageName` reads `runtime.Caller(4)`, which assumes the production call depth (`MustGetLogger → GetLogger → GetLoggerWithReplacements → GetPackageName`). From a test at that depth, frame 4 is `testing.tRunner`:

```go
dotAfterSlash := strings.Index(fullFuncName[lastSlash:], ".")
```

`testing.tRunner` has no `/`, so `lastSlash` is `-1` and `fullFuncName[-1:]` panics. The function already returns an error for its other two failure modes (`!ok`, `fn == nil`), so the unguarded parse looks like an oversight rather than a decision. It only bites a caller at an unexpected stack depth, so nothing in production reaches it today.

## Testing

Plain testify, no Ginkgo. Tests that touch the global registries are marked `//nolint:paralleltest` in line with the existing files. `golangci-lint` (v2.13.2), `go vet` and `gofmt -l` all clean. No production code changed.


sid200727:
@mbrandenburger, one thing worth your call before review. GetPackageName panics rather than returning an error when it's called at an unexpected stack depth (details in the description). 
Happy to guard the parse in this PR, which would make GetLogger, MustGetLogger and GetLoggerWithReplacements testable too. 
Or I'll leave it and open a separate issue, whichever you prefer.

SaidAltury-ibm (approved):
thank you @sid200727 LGTM

PR #1840, opened 2026-09-15
fix(fabric/delivery): classify commit failures instead of stopping the channel on any error

### LINKED ISSUE #1624: committer: define failure-handling strategy for MembershipService.Update errors in CommitConfig
## Summary

PR #1611 replaced a `panic(err)` in `Committer.CommitConfig` (`platform/fabric/core/generic/committer/config.go`) with `return errors.Wrapf(err, ...)` when `MembershipService.Update` fails after a config block has already been durably committed to the vault. This was the right call defensively — a malformed/rejected config transaction should no longer crash the whole process — but per [review feedback on #1611](https://github.com/hyperledger-labs/fabric-smart-client/pull/1611#discussion_r3665107275), returning an error just moves the node into a different bad state, and this issue is the tracked follow-up to design a real strategy for it.

## Current behavior after #1611

`CommitConfig` is invoked as the `common.HeaderType_CONFIG` handler from `Committer.commitTxs`, which itself is called from `Committer.Commit`, which is the block-commit callback passed to the channel's `Delivery` service (wired up in `platform/fabric/core/generic/channelprovider.go`).

When `MembershipService.Update` fails now:

1. `CommitConfig` returns a wrapped error instead of panicking.
2. `commitTxs` propagates it up through `errgroup.Group.Wait()`, so `Committer.Commit` returns a non-nil error.
3. The delivery callback (`channelprovider.go`, the `func(ctx, block) (bool, error) { return false, committerService.Commit(ctx, block) }` closure) returns that same error.
4. `Delivery.readBlocks` (`platform/fabric/core/generic/delivery/delivery.go`) sees the error, logs it, and calls `d.Stop(err)` — which permanently terminates the delivery goroutine for that channel. There is no retry, backoff, or restart anywhere in this path (the doc comment on `driver.BlockCallback` says "the same block is re-processed after a delay" but that is not what the current implementation of `readBlocks` does on error — it stops, it doesn't retry).
5. Nothing else in the process observes that delivery has stopped: no health check, no metric, no supervisor that restarts it. The node keeps running, keeps serving whatever RPCs/views are already up, but silently stops ingesting new blocks on that channel.

Net effect: instead of a loud, obvious crash (the old `panic`), we now have a quiet failure mode — the FSC node stays up, looks alive, but its committer/membership state for that channel is permanently stuck as of the last successfully committed config block. Any channel config change after that point (org added/removed, orderer endpoints changed, MSP rotated, etc.) is never picked up, and — because delivery for the whole channel is stopped, not just config-transaction handling — no further blocks of *any* type are committed either, silently blocking all "progress" on that channel.

This is arguably worse to operate than the panic: a crashed process is unambiguous and typically restarted by the process supervisor (systemd/k8s/etc.), while a "successfully running but stuck" node can go unnoticed until it causes a harder-to-diagnose downstream symptom (stale membership, transactions timing out waiting for finality, etc.).

## Why `MembershipService.Update` can fail

`Update` (`platform/fabric/core/generic/membership/membership.go`) parses the config envelope's payload and builds a new `channelconfig.ChannelConfig`. Failure modes include at least:
- malformed/unparseable config payload (`protoutil.UnmarshalPayload`/`UnmarshalConfigEnvelope`)
- a config block that is structurally valid protobuf but semantically invalid per Fabric's channel-config rules (`channelconfig.NewChannelConfig`)

Since the block reaches this point only after already being sequenced/ordered and durably committed to the vault (`commitConfig` runs first and succeeds), this is not "attacker forges a config block" (ordering/endorsement already gate that) so much as "this node's parsing/validation logic rejects a config block that the rest of the network accepted" — e.g. a version skew bug, or a corner case in `channelconfig.NewChannelConfig` this code doesn't handle. That distinction matters for picking a strategy: it's a local processing failure on an already-agreed-upon block, not an untrusted/malicious input in the same sense as the rest of #1611's findings.

## Open questions raised in review (unresolved)

From [@mbrandenburger's review](https://github.com/hyperledger-labs/fabric-smart-client/pull/1611#discussion_r3665107275):

> what is our strategy here to update the membership service eventually? With the panic - we killed the FSC node - not good at all - but very obvious that something serious happened. If we not return an error, we need to handle this situation accordingly - retry? partially apply? do something? otherwise the FSC node continues to run with a stale membership service - which may also prevent any "progress".

Concretely, this needs a decision on:
1. **Retry?** Config parsing/validation is presumably deterministic given the same input — retrying the same call without change is unlikely to help. Would only make sense if `Update` failure could be transient (it doesn't look like it, today — no I/O in the failure path).
2. **Partially apply?** Not obviously well-defined for a channel config: `ChannelConfig` is replaced wholesale in `Update` (`c.channelResources = b`), there's no notion of a partial config today.
3. **Surface it loudly without killing the process** — e.g. a health/readiness signal the node exposes (metric, log at ERROR with alerting, a `Channel`-level "degraded" state queryable by the embedding application) so operators/monitoring can detect "delivery stalled on channel X" instead of it being silent.
4. **Keep delivery running for other transaction types, only stall config processing** — right now a `CONFIG` handler failure stops delivery for the *entire* channel (all tx types), which seems disproportionate: an endorser transaction after the failed config block can't be causally dependent on that config having applied within the same block, so continuing to deliver/commit non-config transactions may be safe, while quarantining just the config path.
5. **Fail fast/deliberately re-introduce a controlled process exit** for this specific case, since a stuck node with no external signal may be strictly worse than a crash a supervisor can restart — but if so, that should be an explicit, documented decision (and ideally distinguishable in logs/metrics from an actual crash), not a bare `panic`.

## Suggested next step

Design (in a follow-up PR, per @mbrandenburger's and @adecaro's agreement on the #1611 thread) an explicit failure-handling strategy for `MembershipService.Update` failures in `CommitConfig`, covering at minimum:
- Whether delivery for the channel should actually stop (today's incidental behavior) or continue for non-config transactions.
- How this failure is surfaced to an operator/embedding application (metric, health check, structured log at minimum) rather than being observable only as "this channel stopped progressing."
- Whether any retry/backoff is warranted, and if not, explicitly documenting why (deterministic failure on an already-final block).

## References

- PR: #1611
- Review thread: https://github.com/hyperledger-labs/fabric-smart-client/pull/1611#discussion_r3665107275
- Code: `platform/fabric/core/generic/committer/config.go` (`CommitConfig`), `platform/fabric/core/generic/membership/membership.go` (`Service.Update`), `platform/fabric/core/generic/delivery/delivery.go` (`Delivery.readBlocks`)

### LINKED ISSUE #1731: Delivery service stops permanently when the committer returns an error
### What happened?

A single error returned by `Committer.Commit()` — including a transient one, such as a vault/Postgres hiccup — permanently stops the channel's block delivery goroutine. Nothing restarts it. From that point on the node never sees another block on that channel: the vault stops being updated, `notifyFinality` never fires again, and every subsequent `IsFinal()` degrades to the one-shot `SeekNewest` peer probe followed by a `listenTo` timeout.

Expected: a transient commit failure is retried, or the delivery service is restarted, so block processing resumes. The comment at the callback site says a retry is what happens.

The block callback wired into `Delivery` propagates the commit error (`platform/fabric/core/generic/channelprovider.go:248-251`):

```go
func(ctx context.Context, block *common.Block) (bool, error) {
    // commit the block, if an error occurs then retry
    return false, committerService.Commit(ctx, block)
},
```

There is no retry — `readBlocks` stops the whole service (`platform/fabric/core/generic/delivery/delivery.go:203-208`):

```go
stop, err := d.callback(b.ctx, b.block)
if err != nil {
    logger.Errorf("callback errored for block [%d], stop delivery: [%v]", b.block.Header.Number, err)
    d.Stop(err)
    return
}
```

And `Delivery` is single-use by design (`platform/fabric/core/generic/delivery/delivery.go:75-77`):

```go
// callback for each one. A Delivery is single-use: once stopped it cannot be
// restarted, and Run must not be called more than once. Stop may be called
// concurrently with Run and from any number of goroutines.
```

`delivery.Service` calls `deliveryService.Start(ctx)` exactly once (`platform/fabric/core/generic/delivery/service.go:124-127`) and there is no supervisor above it.

Note the asymmetry: `runReceiver` reconnects **indefinitely** on gRPC stream failure (`delivery.go:230-290`, sleeping `DeliverySleepAfterFailure()`, 10s default). Stream-level faults self-heal; committer-level faults are fatal. Any description of the delivery service as "resilient, self-healing, recovery built-in and automatic" covers only the stream half.

Impact:

- Silent, permanent loss of the block pipeline for a channel after one transient commit error. Nothing surfaces to the application except `IsFinal()` timeouts that look like ordering or peer problems.
- It turns the known `SeekNewest` false negative in `FabricFinality.IsFinal` from a race into a steady state: with delivery dead the vault never learns about the transaction, so stage 1 always returns `Unknown`, stage 2's `SeekNewest` probe misses any already-committed block, and stage 3's `listenTo` can only time out.
- Above FSC, the Panurus/token-SDK `finalityView` never sees `Confirmed`/`Deleted` and expires on its own default (5 min) timeout.

### How can we reproduce it?

By code inspection:

1. Read `platform/fabric/core/generic/delivery/delivery.go:196-212` (`readBlocks`) — any non-nil error from the callback calls `d.Stop(err)` and returns.
2. Confirm `Stop` is one-shot and no path re-runs `Run`: `grep -rn "deliveryService.Start\|\.Run(" platform/fabric/core/generic/delivery/`.
3. Confirm the callback is the committer: `platform/fabric/core/generic/channelprovider.go:248-251`.

As a unit test (`platform/fabric/core/generic/delivery/delivery_test.go` already has the harness):

1. Build a `Delivery` whose callback returns an error on the second block.
2. Feed it three blocks.
3. Assert `Run` returns and the callback is never invoked for the third block.

End to end:

1. Start a node against a running Fabric network on any channel.
2. Make `Vault.CommitTX` fail exactly once (fault injection, or take the vault's Postgres down for one block).
3. Restore the vault and submit further transactions.
4. Observe that the node commits no further blocks for the remainder of the run, and that `IsFinal()` on the new transactions times out.

### Environment

- Operating system (e.g., Linux, MacOS): MacOS (platform-independent — this is a control-flow defect, not environment-specific)

### Additional Information

Found by code inspection while verifying a third-party analysis of the finality feature against the implementation, not by observing it at runtime. The reproduction steps above are derived, not executed.

Suggested fix — pick one of the two intended semantics; the comment says the first:

1. **Retry the block.** Bounded retry inside `readBlocks` before giving up, and only then `Stop`. Note that `Commit` is **not** idempotent per block as written: replaying a block whose transactions already landed makes `CommitTX` return `"[txid] is already valid"` / `"[txid] is invalid"` (`platform/fabric/core/generic/committer/committer.go:266-274`), so a naive retry fails a second time for a different reason. A retry would first have to treat those two returns as "already done, skip" rather than as errors — defensible on its own, since a block replay after a reconnect is a normal event, but real work rather than a loop around the existing call.
2. **Supervise and restart the Delivery.** In `delivery.Service`, when `Run` returns a non-nil error, construct a fresh `Delivery` and `Start` it again after a backoff. `GetStartPosition` (`delivery.go:416-455`) already resumes from `lastBlockReceived` → vault → last txID → genesis, so no block is skipped. Smaller change, and it respects `Delivery`'s documented single-use contract instead of bending it. It does replay the last block, so it depends on the same "already valid is not an error" question as option 1 — but only at the resume boundary, which the existing reconnect path already hits today.

Either way, fix or delete the misleading comment at `channelprovider.go:249`.

Related:

- #1732 — same call path.
- #1733 — the other reason a stalled block stream goes unnoticed.
- `platform/fabricx` has the same class of problem in `fabricx/core/finality/provider.go:170-183` (a single `listen()` with no restart, and `finalityServiceAdapter` holding the dead manager for the channel's lifetime). Worth its own issue if fabric-x stream recovery is in scope.

SaidAltury-ibm:
### Summary

`Committer.Commit` returning an error permanently stopped the channel's block
delivery. `Delivery` is single-use, nothing supervised it, and nothing reported the
stop — so one transient storage fault silently ended block ingestion for that
channel. Note the asymmetry: `runReceiver` already reconnects indefinitely on gRPC
failure, so stream faults self-healed while committer faults were fatal.

Classify the failure **in the committer** and act on the class:

- **retry** — non-deterministic and replay-safe (`DeadlockDetected`, `SqlBusy`, an
  expired context): commit the same block again, bounded, then escalate
- **degrade** — deterministic on an already-final block, such as a configuration
  this node cannot apply: return the error, stopping the channel, and count it
- **fatal** — an invariant is broken, so committing further blocks would build on
  state the node cannot vouch for

Classification lives in the committer because that is the component that can tell
the cases apart: it owns the vault, the membership service and the channel
configuration. `Commit` therefore returns only final errors, and `readBlocks` stops
on any error without inspecting one — which is what its own doc comment always
described.

Retrying needs no new idempotence work: `CommitEndorserTransaction` already skips
transactions marked valid or invalid, and `CommitConfig` skips a configuration
present in the vault.

A stopped channel emits no blocks and no further errors, so it was invisible in
every existing signal — `block_commit` simply stops receiving observations, which
looks identical to an idle channel. Two counters record it, labelled by `network`,
`channel` and failure `class`.

Also drops the comment at the callback site claiming a failed commit is retried,
and the same claim on `driver.BlockCallback`.

Fixes #1731
Fixes #1624


docs/configuration.md
```diff
@@ -642,6 +642,19 @@ fabric:
...
+          # treated as permanent and this channel's delivery stops. Replaying a
+          # block is safe: the committer skips transactions it has already
+          # committed.
+          #
+          # sleepAfterFailure is the wait between attempts, so the two together
+          # set how long a channel is given to recover on its own — with the
+          # defaults, about a minute. Raise it to ride out a longer storage
+          # failover; set it to 0 to commit once and never retry. A stopped
+          # channel commits no further blocks of any type and is reported as
+          # fsc_fabric_core_generic_delivery_commit_failures.
```
mbrandenburger: should this comment sit above sleepAfterFailure?

platform/fabric/core/generic/delivery/delivery.go
```diff
@@ -88,9 +88,20 @@ type Delivery struct {
...
 	client              services.PeerClient
 	tracer              trace.Tracer
+	metrics             *Metrics
 	lastBlockReceived   uint64
 	bufferSize          int
 
+	// commitRetries bounds how many times a block whose commit failed
+	// transiently is replayed before the failure is treated as permanent. Zero
+	// means a single attempt with no retry.
+	commitRetries int
```
mbrandenburger: I am wondering if this is the correct naming. Should this callbackRetries instead as the delivery service is actually independent from the committer? Should the committer contain the retry logic - as it understands better understands what are transient/permanent errors are? What would mean that if the delivery received an error - it's already permanent or fatal?

platform/fabric/core/generic/delivery/metrics.go
```diff
@@ -0,0 +1,58 @@
...
+	// labelled by failure class. Any increment means a channel has stopped
+	// making progress and will not resume without intervention.
+	CommitFailures metrics.Counter
+}
+
+// NewMetrics builds the delivery counters from p. A nil p yields counters that
+// discard their observations, so that a Delivery built without a metrics provider
+// still runs; the counters are a diagnostic, never a precondition for delivering
+// blocks.
+func NewMetrics(p metrics.Provider) *Metrics {
```
mbrandenburger: Should we add a channel label for these metrics?

platform/fabric/core/generic/delivery/failure.go
```diff
@@ -0,0 +1,173 @@
...
+
+// classify maps an error returned by a block callback to the action to take.
+//
+// Retry is granted only on evidence, never by default: an error must name a
+// cause known to be non-deterministic — storage contention, or an expired
+// context — before the same block is replayed. Everything else is a degrade,
+// because retrying a deterministic failure burns the budget and arrives at the
+// same place later. That is the conservative direction to be wrong in: a
+// misclassified degrade stalls the channel loudly, a misclassified retry spins.
+func classify(err error) failureClass {
```
mbrandenburger: Are these error classes already wired into the committer? Coming back to my earlier comments; should the delivery component care about these errors classes which are committer specific? Note that the committer has access to the vault and may understand what vault error is persistent or permanent, it has access to the membership service and can decide if there is a config tx updating failing that produces a fatal error. etc ... WDYT?

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for working on this. I strongly believe that the classify-instead-of-stop idea is right - but I am not sure if the delivery service is the right place. 

Please see my comments below.

docs/configuration.md
```diff
@@ -638,6 +638,23 @@ fabric:
...
+          # ride out a longer storage failover; set retries to 0 to commit once
+          # and never retry.
+          #
+          # When the budget is exhausted the error is returned, which stops this
+          # channel's block stream: no further blocks of any type are committed
+          # until the node is restarted. That is reported as
+          # fsc_fabric_core_generic_committer_commit_failures, labelled by
+          # network, channel and failure class.
+          retries: 5
+          retrySleep: 10s
```
mbrandenburger, line 657: Can we shorten this doc? I believe we should keep this here brief.

platform/fabric/core/generic/channelprovider.go
```diff
@@ -246,7 +246,11 @@ func (p *provider) NewChannel(nw driver.FabricNetworkService, channelName string
 		ledgerService,
 		&vaultDeliveryWrapper{vaultStore: vaultStore},
 		func(ctx context.Context, block *common.Block) (bool, error) {
-			// commit the block, if an error occurs then retry
+			// Commit classifies and retries internally, so anything it returns is
+			// final and stops this channel's delivery. Returned unannotated: the
+			// committer has already logged and counted the failure with the
+			// context to describe it, and wrapping here would only obscure the
+			// cause for a caller matching on it.
```
mbrandenburger, line 253:
```suggestion
```

platform/fabric/core/generic/committer/committer.go
```diff
@@ -289,19 +294,104 @@ func (c *Committer) RemoveFinalityListener(txID string, listener driver.Finality
...
+// retryBlock commits a block, retrying while the failure classifies as
+// transient. It returns nil once the block is committed, or the error that ended
+// the attempts.
+//
+// The budget is bounded on purpose: an unbounded retry against a fault that turns
+// out to be permanent is a silent stall, which is the failure mode this path
+// exists to avoid. Exhausting it wraps the last error in ErrRetriesExhausted so
+// that it escalates out of the retry class, keeping the original cause matchable
+// underneath.
+func (c *Committer) retryBlock(ctx context.Context, block *common.Block) error {
```
mbrandenburger, line 335: while this is a solid implementation I am wondering if we could re-usue things from `pkg/utils/retry.go` here.

platform/fabric/core/generic/committer/failure.go
```diff
@@ -0,0 +1,163 @@
...
+var ErrRetriesExhausted = errors.New("commit retries exhausted")
+
+// ErrFatalCommit marks an error whose cause leaves this node unable to trust its
+// own committed state, so that block processing must not continue over it.
+//
+// Nothing in the commit path returns it today: the failure modes reached here are
+// either transient or bounded to one block or one channel, and none has been
+// shown to corrupt state. It exists so that a caller which does detect such a
+// condition can say so distinguishably, rather than reaching for panic from
+// inside a commit.
```
mbrandenburger, line 100: Should we use this error class for failures related to certain membership service updates?

platform/fabric/core/generic/committer/committer.go
```diff
@@ -289,19 +294,104 @@ func (c *Committer) RemoveFinalityListener(txID string, listener driver.Finality
...
+			}
+		}
+
+		// Checked before every attempt, including the first: a caller that has
+		// already given up must not have this block committed on the way out, and
+		// a zero retry delay would otherwise let the select above fall straight
+		// through to another attempt.
+		if err := ctx.Err(); err != nil {
+			return err
+		}
```
mbrandenburger, line 357: I believe returning the error here is not the right thing - as it might hide the last actual error. What about the following ...

```go
      abandon := func(err error) error {
              return errors.Wrapf(errors.Join(ErrRetriesExhausted, err), "block [%d] abandoned: %v", blockNum, ctx.Err())
      }

      var err error
      for attempt := 0; attempt <= retries; attempt++ {
              if attempt > 0 {
                      c.logger.Warnf("retrying block [%d] after transient commit failure, attempt [%d/%d]: [%v]",
                              blockNum, attempt, retries, err)
                      select {
                      case <-ctx.Done():
                              return abandon(err)
                      case <-time.After(c.ChannelConfig.CommitRetrySleep()):
                      }
              }

              if ctx.Err() != nil {
                      return abandon(err)
              }

              err = c.commitOnce(ctx, block)
              if classify(err) != classRetry {
                      return err
              }
              c.metrics.CommitRetries.Add(1)
      }
```

platform/fabric/core/generic/committer/committer.go
```diff
@@ -289,19 +294,104 @@ func (c *Committer) RemoveFinalityListener(txID string, listener driver.Finality
...
+
+		err = c.commitOnce(ctx, block)
+		if classify(err) != classRetry {
+			return err
+		}
+		c.metrics.CommitRetries.Add(1)
```
mbrandenburger, line 363: Do we check that the metrics counter is correct? I am wondering if we are off by one here - can you please check.

platform/fabric/core/generic/committer/committer.go
mbrandenburger, line 335: How do we deal with an interrupt via stop. Is the context used here bound the the fsc node lifecycle? So, if a node is being shutdown - the retry should be aborted?!

platform/fabric/driver/config.go
```diff
@@ -53,6 +53,13 @@ type ChannelConfig interface {
 	CommitterWaitForEventTimeout() time.Duration
 	DeliveryBufferSize() int
 	DeliverySleepAfterFailure() time.Duration
+	// CommitRetries is how many times a block whose commit failed transiently is
+	// committed again before the failure is treated as permanent and the
+	// channel's block stream is stopped. Zero commits once without retrying.
+	CommitRetries() int
+	// CommitRetrySleep is how long to wait between attempts at committing the
+	// same block.
+	CommitRetrySleep() time.Duration
```
mbrandenburger, line 62: I am wondering if we need these config knobs as part of the driver or if they are just specific to the to committer implementation and can be a local interface extension in the `platform/fabric/core/generic/committer`. WDYT?

platform/fabric/core/generic/committer/committer.go
```diff
@@ -289,19 +294,104 @@ func (c *Committer) RemoveFinalityListener(txID string, listener driver.Finality
...
 
-// Commit commits the transactions in the block passed as argument
+// Commit commits the transactions in the block passed as argument.
+//
+// A transient failure — storage contention, an expired context — is retried here,
+// on this same block, because this is the component that can tell a fault worth
+// retrying from one that will never clear: it owns the vault, the membership
+// service and the channel configuration. Committing a block again is safe:
+// CommitEndorserTransaction skips transactions already marked valid or invalid,
+// and CommitConfig skips a configuration already in the vault.
```
mbrandenburger, line 304: > CommitConfig skips a configuration already in the vault

I believe the current PR still has a gap around config commits that could silently mask the exact failure #1624 describes.

In particular, the current `CommitConfig` first tries to commit the config to the vault, and the checks if the membership service can be updated; if that fails, we may retry again and then early return without an error as the config tx is already committed to the vault. In other words, we have a commit-before-verify problem for the config transactions. Does this make sense? Maybe we could write a test describing this scenario to double check.

Maybe a simple fix is just a reordering the steps in `CommitConfig`. For example something like that:

```go
func (c *Committer) CommitConfig(ctx context.Context, blockNumber driver.BlockNum, raw []byte, env *common.Envelope) error {
      ...
      switch vc {
      case driver.Valid:
              c.logger.Debugf("config block [%s] already committed, skip it.", txID)
              return nil
      case driver.Unknown:
              // this is okay
      default:
              return errors.Errorf("invalid configtx's [%s] status [%d]", txID, vc)
      }

      // Membership and orderer config are applied before the vault write, so that
      // a retry landing after the write finds vc == Valid only once both have
      // actually succeeded, not just the write itself.
      if err := c.MembershipService.Update(env); err != nil {
              return errors.Wrapf(err, "failed updating membership service for configtx [%s]", txID)
      }
      if err := c.applyConfigUpdates(); err != nil {
              return errors.Wrapf(err, "failed applying config updates for configtx [%s]", txID)
      }

      if err := c.applyConfigCommit(ctx, txID, blockNumber, sequence, raw); err != nil {
              return errors.Wrapf(err, "failed committing configtx to the vault")
      }
      return nil
}
```

mbrandenburger (changes requested):
Thanks @SaidAltury-ibm for the last updates on this PR. I left you another round of comments.

platform/fabric/core/generic/committer/committer.go
SaidAltury-ibm: Good question, I did start there but:
- It cannot express "retry only this class of error."
- The sleep is not interruptible. RunWithErrors does time.Sleep(f.getNextDelay()), so a retry cannot be aborted part-way. That conflicts directly with your question about aborting on shutdown
- The error it returns is the wrong shape. It returns errors.Join of every error seen, or ErrMaxRetriesExceeded when there were none. We need the last error wrapped in ErrRetriesExhausted so that classify escalates it out of the retry class

mbrandenburger: ok understood! Refactoring the existing retry code is probably out of scope.

SaidAltury-ibm: Yes it should, and it does but by the context, not by Stop.
The context is the node's: it comes from Delivery.Run(ctx), through the block callback, into Commit. Both exits from the retry loop check it, so cancelling it aborts a retry promptly, including mid-sleep.

Two things I would like to do about it:

- State the contract in Commit's Godoc — a caller that wants to interrupt a retry must cancel its context, not only stop its stream. That is a one-line change and I will include it.
- If you want a hard guarantee rather than a documented convention, the honest fix is for Delivery.Stop to cancel the context it handed to Run, so the two signals cannot diverge. That is a delivery-side change and slightly beyond this PR's scope — tell me if you would like it here and I will add it.

WYT?

mbrandenburger: ok understood. So the committer is canceled with node shutdown - which is independent from the delivery shutdown. In other words, delivery.Stop should not trigger the committer to shutdown as they are co-located and controlled by the channel. Ok, so no need to add additional info here.

platform/fabric/core/generic/committer/failure.go
SaidAltury-ibm: No i do not think so, a rejected config breaks one channel (membership is per-channel), nothing was written incorrectly, and fatal should mean the node's own state is untrustworthy; using it here would make the most severe class the most common one.

platform/fabric/core/generic/committer/committer.go
SaidAltury-ibm: good catch!

mbrandenburger (approved):

mbrandenburger:
cc @mffrench

PR #1844, opened 2026-09-15
fix(logging): report an unparseable caller name instead of panicking

### LINKED ISSUE #1841: fix(logging): GetPackageName panics instead of returning an error when the caller's name has no '/'
### What happened?

`GetPackageName` in `platform/common/services/logging/logger.go` parses the caller's function name without guarding the parse:

```go
func GetPackageName() (string, error) {
	pc, _, _, ok := runtime.Caller(4)
	if !ok {
		return "", errors.New("failed to get caller package name")
	}
	fn := runtime.FuncForPC(pc)
	if fn == nil {
		return "", errors.New("failed to get caller package name")
	}
	fullFuncName := fn.Name()
	dotAfterSlash := strings.Index(fullFuncName[lastSlash:], ".")
	return fullFuncName[:lastSlash+dotAfterSlash], nil
}
```

`strings.LastIndex` returns `-1` when the name contains no `/`, and `fullFuncName[-1:]` panics. Function names in a module path always contain one, but stdlib names don't - `testing.tRunner`, for example.

The `runtime.Caller(4)` depth assumes the production call chain `MustGetLogger → GetLogger → GetLoggerWithReplacements → GetPackageName`. A caller at any other depth resolves frame 4 to something else, and if that something is in the stdlib the function panics rather than returning the error its signature already provides for.

**How it shows up:** calling `GetLogger()` from a test in the `logging` package itself panics, because frame 4 is `testing.tRunner`. That is why `GetLogger`, `MustGetLogger`, `GetLoggerWithReplacements` and `GetPackageName` are the only functions left uncovered in that package after #1837.

**Not a production problem today** - every production caller goes through the expected chain. This is about the function being robust enough to test, and consistent with the error handling it already has.

### How can we reproduce it?

- [ ] Guard the parse in `GetPackageName`: return the existing "failed to get caller package name" error when `lastSlash` or `dotAfterSlash` is negative, rather than slicing.
- [ ] Add tests for the guarded paths.
- [ ] Cover `GetLogger`, `MustGetLogger` and `GetLoggerWithReplacements`, which become testable once the panic is gone.

### Environment

The bug is not environment-specific: the parse is unguarded regardless of platform or Go version. These are just the values from where it was found.

### Additional Information

Found while covering the package in #1837, where `GetLogger`, `MustGetLogger`, `GetLoggerWithReplacements` and `GetPackageName` had to be left uncovered for this reason. Reproduces by calling `GetLogger()` from any test in `platform/common/services/logging`.

sid200727:
Fixes #1841

## The problem

`GetPackageName` parses the caller's function name assuming it looks like `github.com/org/repo/pkg.Func`  everything up to the first dot after the last slash:

```go
dotAfterSlash := strings.Index(fullFuncName[lastSlash:], ".")
return fullFuncName[:lastSlash+dotAfterSlash], nil
```

`strings.LastIndex` returns `-1` when there is no slash, and `fullFuncName[-1:]` panics.

Names from a module path always carry a slash, so nothing in production reaches it: every caller arrives through the chain that the function's `runtime.Caller(4)` depth assumes. A caller at any other depth can resolve frame 4 to a standard-library function, whose name has no slash.

## The fix

Both cases now return the error the signature already provides for, matching the two failure modes it already handles (`!ok` and `fn == nil`).

## What that makes testable

With the panic gone, the three entry points can be called from a test:

- `MustGetLogger` adds a frame of its own, so from a test the name resolves  `fsc.platform.common.services.logging`, with the registered replacement applied  and params are appended as expected.
- `GetLogger` and `GetLoggerWithReplacements` called directly find no fourth frame and report it, rather than crashing the run.

## Coverage

The package goes from 91.4% to **98.2%** (filtered). `GetPackageName` sits at 85.7%: the two new branches aren't reachable from a test, since no test can place a slash-less name at frame 4. They're worth having anyway the point is that a caller who does hit them gets an error rather than a panic.

## Testing

`go test`, `go vet`, `gofmt -l` and `golangci-lint` (v2.13.2) all clean.


platform/common/services/logging/logger.go
```diff
@@ -96,9 +97,22 @@ func GetPackageName() (string, error) {
...
+	// whose final segment carries no dot, so report those rather than slicing
+	// with a negative index.
 	fullFuncName := fn.Name()
 	lastSlash := strings.LastIndex(fullFuncName, "/")
+	if lastSlash < 0 {
+		return "", fmt.Errorf("caller package name has no path separator: %s", fullFuncName)
+	}
 	dotAfterSlash := strings.Index(fullFuncName[lastSlash:], ".")
+	if dotAfterSlash < 0 {
+		return "", fmt.Errorf("caller package name has no function separator: %s", fullFuncName)
```
SaidAltury-ibm:
```suggestion
		return "", errors.Errorf("caller package name has no function separator: %s", fullFuncName)
```

platform/common/services/logging/logger.go
```diff
@@ -96,9 +97,22 @@ func GetPackageName() (string, error) {
...
+	// A function name from a module path looks like
+	// github.com/org/repo/pkg.Func, so the package name is everything up to the
+	// first dot after the last slash. Names without a slash (the standard
+	// library's, for instance) do not have that shape, and neither does one
+	// whose final segment carries no dot, so report those rather than slicing
+	// with a negative index.
 	fullFuncName := fn.Name()
 	lastSlash := strings.LastIndex(fullFuncName, "/")
+	if lastSlash < 0 {
+		return "", fmt.Errorf("caller package name has no path separator: %s", fullFuncName)
```
SaidAltury-ibm:
```suggestion
		return "", errors.Errorf("caller package name has no path separator: %s", fullFuncName)
```

platform/common/services/logging/logger.go
```diff
@@ -8,6 +8,7 @@ package logging
 
 import (
 	"errors"
+	"fmt"
```
SaidAltury-ibm:
```suggestion

```

platform/common/services/logging/logger.go
SaidAltury-ibm, line 18:
```suggestion

	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"

```

SaidAltury-ibm (dismissed):
thank you @sid200727 for this PR, i have added few comments below. please also resolve the conflict. thanks!

mbrandenburger (approved):
Thank you @sid200727 for working on this!

mbrandenburger:
@sid200727 can you please rebase this PR manually; the github UI rebase cannot do it due to a rebase conflict. Thank you

sid200727:
Hi @mbrandenburger !
Rebased manually onto current main as a single commit. Upstream had already inlined MustGetLogger in #1858, so that part dropped out; the change is now just the guard in GetPackageName plus the tests, which replace the TestMustGetLogger comment that described #1841 as unfixed.
Thank you 🙏

PR #1850, opened 2026-09-16
test(fabric): cover the identity provider in platform/fabric/core/generic/id

### LINKED ISSUE #1849: test(fabric): increase coverage of platform/fabric/core/generic/id
### Description

Coverage for `platform/fabric/core/generic/id` sits at 56.2% with `scripts/filter-coverage.sh` applied. `info_test.go` covers `Info` for both an idemix and an x509 identity, but `provider.go` has no tests at all.

|---|---|---|
| `NewProvider` | `provider.go` | 0% |
| `Identity` | `provider.go` | 0% |
| `Info` | `info.go` | 81.8% |

`provider` wraps a single-method `EndpointService`, so it needs only a small fake. `Info`'s uncovered path is the one where neither the x509 nor the idemix deserializer can parse the identity.

### Implementation Steps

- [ ] Cover `NewProvider`.
- [ ] Cover `Identity`: a label that resolves, and one whose lookup fails, checking the error names the label.
- [ ] Cover `Info` for an identity neither deserializer understands.

### Additional Information

- **Target package:** `platform/fabric/core/generic/id`
- **Parent:** #1566

sid200727:
Fixes #1849

## Background

The package sat at 56.2% with `scripts/filter-coverage.sh` applied. `info_test.go` already covered `Info` for both an idemix and an x509 identity, but `provider.go` had no tests at all.

## What this covers

`provider` wraps a single-method `EndpointService`, so a small stub drives both functions: `NewProvider`, and `Identity` for a label that resolves and one whose lookup fails, checking the error names the label that could not be resolved, since that is what makes it useful to a caller.

Also covers `Info` for an identity neither deserializer understands, which renders as the empty string rather than failing.

## Coverage

56.2% → **93.8%** (filtered). `provider.go` is at 100%.

The one remaining line is `Info`'s guard on `idemix.NewDeserializer(nil)`. With an empty issuer public key that constructor skips its key import entirely, so it can only fail if the crypto provider itself does — not reachable from a test.

## Testing

Plain testify, no Ginkgo. No production code changed. `go vet`, `gofmt -l` and `golangci-lint` (v2.13.2) all clean.


SaidAltury-ibm (approved):
LGTM

PR #1854, opened 2026-09-16
refactor(logging): inline MustGetLogger's panic instead of utils.MustGet

mbrandenburger:
## Summary
- `MustGetLogger` routed through the generic `utils.MustGet` helper to turn `GetLogger`'s error into a panic. Inlines that check directly so the panic follows the `regexp.MustCompile` convention on its own terms, without depending on the generic helper.
- Adds Godoc to `MustGetLogger`, `GetLogger`, and `GetPackageName` explaining why `GetLogger` can only fail on a static invariant (the caller's stack frame can't be resolved) rather than runtime input, which is why panicking on it here is appropriate.
- Adds `TestMustGetLogger` as a regression test for the inlined panic/return path.

## Test plan
- [x] `go test ./platform/common/services/logging/...`
- [x] `make checks`


SaidAltury-ibm (approved):
LGTM

PR #1855, opened 2026-09-16
refactor!: retire utils.IsNil

mbrandenburger:
`utils.IsNil` used reflection to check whether a generic value's pointer, slice, map, chan, func, or interface is nil, but `iterators.flatten.go` was its only caller outside its own test. Keeping a reflection-based nil check as public API for one internal use isn't worth the surface area.

`isNil` moves into the `iterators` package, unexported, next to `flatten.go`. Its test coverage moves with it into a new `flatten_internal_test.go`, and `utils.IsNil` and its test are removed.

BREAKING CHANGE: `utils.IsNil` is removed, with no replacement in package `utils`.

How it was verified: `go test ./platform/common/utils/...` (all green,
including the new flatten_internal_test.go cases).


SaidAltury-ibm (approved):
LGTM

PR #1856, opened 2026-09-16
refactor(utils): remove Zero, inline its zero-value construction at call sites

mbrandenburger:
## Summary
- `utils.Zero[A any]() A` just does `var a A; return a`. Removes it and inlines a local `var zero V` at each of its six call sites instead.
- All six sites are `return utils.Zero[V](), err` in the generic persistence constructors (`newPersistenceWithOpts`/`NewPersistenceWithOpts`) across the view-storage and fabric-db drivers (memory, sqlite, postgres), which already return `(V, error)`.
- Drops the now-unused `utils` import from each of those six driver files.

BREAKING CHANGE: utils.Zero is removed, with no replacement in package utils.

## Test plan
- [x] `go build ./...`
- [x] `go test ./platform/common/utils/... ./platform/view/services/storage/driver/... ./platform/fabric/services/db/driver/...`
- [x] `make checks`


SaidAltury-ibm (changes requested):
Thank you @mbrandenburger, great work but i think it should be marked as breaking change. WYT?

SaidAltury-ibm (approved):
LGTM

PR #1858, opened 2026-09-16
refactor!: retire utils.Must and utils.MustGet

mbrandenburger:
## Summary
- `utils.Must` and `utils.MustGet` panicked on error instead of returning it, so every call site was a hidden crash risk. This converts each caller to its surrounding idiom: testify `require`/`assert` or gomega `Expect(...).NotTo(HaveOccurred())` in `*_test.go` files, plain error returns in `integration/benchmark/grpc`'s `ServerOption`/`ClientOption`, and `kvs.CreateCompositeKeyOrPanic` (an existing, equivalent helper) for the one package-level var that can't take a `*testing.T`.
- With no callers left, `Must`, `MustGet`, and their tests are removed from `platform/common/utils`.

**BREAKING CHANGE**: `utils.Must` and `utils.MustGet` are removed from `platform/common/utils`.

## Test plan
- [x] `go build ./...` / `go vet ./...` (root and `integration` modules)
- [x] `make checks`
- [x] Targeted `go test` on every touched package


SaidAltury-ibm (approved):
thank you @mbrandenburger, LGTM

PR #1860, opened 2026-09-16
refactor: replace fmt.Errorf and stdlib errors.New with pkg/utils/errors

### LINKED ISSUE #1859: Production code drifted from the pkg/utils/errors convention
docs/agents/conventions.md requires all error construction and wrapping to go through pkg/utils/errors and forbids fmt.Errorf. Nothing enforces that today: .golangci.yml has no forbidigo rule (or equivalent) for it, so the convention only holds as long as everyone remembers it by hand.

Production code had drifted: 127 fmt.Errorf call sites across 44 files, plus 14 stdlib errors.New call sites across 9 files, none of them going through pkg/utils/errors.

Scope: migrate those call sites to pkg/utils/errors (New/Errorf/Wrap/Wrapf as appropriate). Test files are excluded on purpose: their errors.New("boom")-style sentinel errors for mocks are ordinary Go testing idiom, not a violation of the convention.

Out of scope for this issue but worth a follow-up: adding a lint rule so new fmt.Errorf/stdlib errors.New in production code gets caught automatically instead of drifting again.

mbrandenburger:
Fixes #1859

Migrates the 127 fmt.Errorf call sites (across 44 files) and 14 stdlib errors.New call sites (across 9 files) in production code to pkg/utils/errors, per the convention in docs/agents/conventions.md. Test files are untouched: their errors.New("boom")-style sentinel errors for mocks are ordinary Go testing idiom, not a convention violation.

fmt.Errorf(...%w..., err) becomes errors.Wrapf(err, ...) (or errors.Wrap when no other args remain); other fmt.Errorf calls become errors.Errorf or errors.New; stdlib errors.New becomes our errors.New. Two files needed non-mechanical import changes: platform/fabric/sdk/dig/sdk.go dropped its stdlib errors import and un-aliased pkg/utils/errors from e to errors, and platform/view/services/metrics/prometheus/provider.go keeps stdlib errors under a stderrors alias for its errors.As call so pkg/utils/errors can stay unaliased as errors.

No behavior change: New/Errorf/Wrap/Wrapf in pkg/utils/errors are thin wrappers over the same cockroachdb/errors calls.

## Test plan
- [x] `make checks` (lint across all 5 modules)
- [x] `make unit-tests`
- [x] `go build ./...` in the integration module
- [x] grep sweep confirming no remaining production fmt.Errorf / stdlib errors.New


SaidAltury-ibm (approved):

PR #1868, opened 2026-09-22
fix(fabric): initialize per-channel map in AddChannelProcessor

### LINKED ISSUE #1867: fix(fabric): AddChannelProcessor panics with "assignment to entry in nil map"
### Description

`processorManager.AddChannelProcessor` in
`platform/fabric/core/generic/rwset/processor.go:117` writes into
`r.channelProcessors[channel][ns]` without ever creating the inner map.
`NewProcessorManager` initializes only the outer
`map[string]map[string]driver.Processor`, so the first call for any channel
panics with `assignment to entry in nil map`.

Current state: the panic is reachable through the exported
`platform/fabric.ProcessorManager.AddChannelProcessor` wrapper
(`platform/fabric/processor.go:40`), which implements
`driver.ProcessorManager` (`platform/fabric/driver/processor.go:37`). No in-repo
caller passes through it today, and the existing unit test hides the bug by
pre-seeding `pm.channelProcessors["ch1"]` by hand before calling the method.

Target state: `AddChannelProcessor` creates the per-channel map on first use for
a channel, and the unit test exercises the real path instead of seeding it.

SaidAltury-ibm:
processorManager.AddChannelProcessor wrote into
channelProcessors[channel][ns] without creating the inner map. NewProcessorManager initializes only the outer map, so the first call for any channel panicked with "assignment to entry in nil map". The method is reachable through the exported platform/fabric.ProcessorManager wrapper, which implements driver.ProcessorManager.

Create the per-channel map on first use for a channel.

The unit test previously seeded channelProcessors["ch1"] by hand, which masked the panic. It now calls AddChannelProcessor directly and adds a second namespace on the same channel to confirm the existing entries survive.

Fixes #1867


mbrandenburger (approved):
Thank you @SaidAltury-ibm for providing this fix.

PR #1870, opened 2026-09-22
fix(metrics): guard the no-separator case in GetPackageName

### LINKED ISSUE #1869: fix(metrics): guard the no-separator case in prometheus GetPackageName
`GetPackageName` in `platform/view/services/metrics/prometheus/provider.go:147` slices on the result of `strings.LastIndex(fullFuncName, "/")` without checking for `-1`: `lastSlash := strings.LastIndex(fullFuncName, "/")`; `dotAfterSlash := strings.Index(fullFuncName[lastSlash:], ".")`; `return fullFuncName[:lastSlash+dotAfterSlash]`.

When the resolved caller frame belongs to a package whose fully qualified function name contains no path separator, `lastSlash` is `-1` and `fullFuncName[lastSlash:]` panics with `runtime error: slice bounds out of range [-1:]`.

Go's `main` package has no separator in its fully qualified function name (`main.main`), so any FSC-based binary that creates a metric directly from `main` crashes. Reproduced with a module that calls `prometheus.NewCounter` from `main`: without the guard it panics with `slice bounds out of range [-1:]`, with the guard it returns normally. This reaches `GetPackageName` through the ordinary `NewCounter` → `applyNamespaceSubsystem` chain at the intended `callerSkipFrames = 3` depth — no unusual call depth required.

`runtime` frames (`runtime.main`, `runtime.goexit`) are slash-free for the same reason, so a direct call to the exported `GetPackageName`, which skips the three frames of that chain, also panics.

Current state: of the three copies of this function in the repo, two guard the `-1` case and one does not.

| Location | Guarded |
|---|---|
| `platform/view/services/tracing/config.go:108` | yes — returns `fullFuncName` |
| `platform/common/services/logging/logger.go:121` | yes — returns an error (added in #1844) |
| `platform/view/services/metrics/prometheus/provider.go:147` | **no** |

Target state: the prometheus copy behaves like its `tracing` sibling and returns the unqualified `fullFuncName` when there is no separator.

Note on the parent checklist: #1861 lists this item as `logging/logger.go:111 and metrics/prometheus/provider.go:147`. The `logging/logger.go` half was already fixed by #1844, which merged before the audit was filed, so only the prometheus copy remains.

SaidAltury-ibm:
GetPackageName sliced on strings.LastIndex(fullFuncName, "/") without checking for -1. Go's main package has no separator in its fully qualified function name, so a binary that creates a metric directly from main resolves to main.main and crashes with "slice bounds out of range [-1:]". The call reaches GetPackageName through the ordinary NewCounter -> applyNamespaceSubsystem chain at the intended callerSkipFrames depth, so no unusual call depth is required. Runtime frames such as runtime.main are slash-free for the same reason.

Return the unqualified name when there is no separator, mirroring the guard in platform/view/services/tracing/config.go.

Fixes #1869


platform/view/services/metrics/prometheus/provider.go
```diff
@@ -145,6 +145,9 @@ func GetPackageName() string {
 	}
 	fullFuncName := fn.Name()
 	lastSlash := strings.LastIndex(fullFuncName, "/")
+	if lastSlash == -1 {
+		return fullFuncName
+	}
```
mbrandenburger, line 159: should we use the chance to update the godoc of `GetPackageName()`

mbrandenburger (approved):
LGTM

PR #1873, opened 2026-09-22
fix(runner): guard the output cardinality in the serial implementations

### LINKED ISSUE #1871: fix(runner): guard the output cardinality in the serial runner and executor
### Description
`pkg/runner/serial.go` indexes the result of its `ExecuteFunc` at `[0]` without checking the length, in both implementations: `func (r *serialRunner[V]) Run(val V) error { return r.executor([]V{val})[0] }` (serial.go:20) and `func (r *serialExecutor[I, O]) Execute(input I) (O, error) { res := r.executor([]I{input})[0]; return res.Val, res.Err }` (serial.go:35).

An `ExecuteFunc` that returns an empty or `nil` slice panics with `runtime error: index out of range [0] with length 0`.

The batched sibling validates exactly this contract before pairing outputs to inputs, at `pkg/runner/batch.go:97`: `if len(inputs) != len(outs) { panic(errors.Errorf("expected %d outputs, but got %d", len(inputs), len(outs))) }`.

So the two implementations of the same interfaces disagree on whether the one-output-per-input contract is checked. `NewSerialRunner` and `NewSerialExecutor` are exported and accept a caller-supplied `ExecuteFunc`, so the contract is only as good as the caller.

The single in-repo caller is safe: `Vault.commitTXs` (`platform/common/core/generic/vault/vault.go:172`, wired at `vault.go:109`) returns `collections.Repeat(..., len(txs))` on every path, so it always returns exactly one output per input. The guard protects downstream implementations of the exported `ExecuteFunc`, which is why the parent issue files this under "no in-repo caller today".

Target state: the serial implementations detect a cardinality mismatch and report it. Both `Run` and `Execute` already return an `error`, so they return one rather than panicking -- unlike `batch.go`, which runs on a background goroutine with no error channel back to the caller and therefore has to panic.

### Implementation Steps

- [ ] In `Run`, return an error when the result length is not 1 instead of
      indexing `[0]`
- [ ] In `Execute`, return the zero `O` and an error when the result length is
      not 1
- [ ] Document the one-output-per-input requirement on `ExecuteFunc` in
      `pkg/runner/runner.go`, since it is the contract both implementations rely on
- [ ] Cover empty and over-long results for both implementations in
      `pkg/runner/serial_test.go`
- [ ] Verify: `go test -count=1 -race ./pkg/runner/... ./platform/common/core/generic/vault/...`
      and `make lint`

### Additional Information

_No response_

SaidAltury-ibm:
serialRunner.Run and serialExecutor.Execute indexed their ExecuteFunc result at [0] without checking its length, so an executor returning an empty or nil slice panicked with "index out of range [0] with length 0".

The batched sibling validates this same one-output-per-input contract at batch.go:97 before pairing outputs to inputs, so the two implementations of BatchRunner and BatchExecutor disagreed on whether it is checked. Both serial methods already return an error, so they report the mismatch rather than panicking; the batcher runs on a background goroutine with no error channel back to the caller and has to panic instead.

Document the requirement on ExecuteFunc, since both implementations rely on it and it constrains caller-supplied executors.

Fixes #1871


mbrandenburger (approved):
Thank you @SaidAltury-ibm. LGTM

PR #1875, opened 2026-09-22
fix(fabric): treat a block with no header as a query failure

### LINKED ISSUE #1874: fix(fabric): treat a block with no header as a query failure
### Description

`Ledger.GetBlockNumberByTxID` at `platform/fabric/core/generic/ledger/ledger.go:84` dereferences `block.Header.Number` with no nil check on `Header`:

```go
block := &common.Block{}
if err := proto.Unmarshal(raw, block); err != nil {
	return 0, errors.Wrap(err, "unmarshal failed")
}
return block.Header.Number, nil
```

`raw` is the response from a `qscc` `GetBlockByTxID` query against a remote peer (`queryChaincode`, `ledger.go:100`). `Header` is an optional protobuf message, so a block carrying no header is a valid encoding: it unmarshals without error, leaves `Header` nil, and the dereference panics with `invalid memory address or nil pointer dereference`. A buggy or compromised peer therefore crashes the client.

The fabricx ledger already handles this exact case at `platform/fabricx/core/ledger/ledger.go:95`, using the nil-safe `GetHeader()` getter and returning an error. `Block.DataAt` and `Block.ProcessedTransaction` in this same file (`ledger.go:128`, `ledger.go:136`) likewise guard nil `Data` and `Metadata` against malformed remote blocks. This one method is the outlier.

Target state: `GetBlockNumberByTxID` uses `block.GetHeader()` and returns an error when it is nil, matching the fabricx sibling.

SaidAltury-ibm:
GetBlockNumberByTxID dereferenced block.Header.Number with no nil check. The bytes come from a qscc GetBlockByTxID query against a remote peer, and Header is an optional protobuf message, so a block carrying no header unmarshals without error, leaves Header nil, and panics the client with a nil pointer dereference.

Use the nil-safe GetHeader() getter and report a missing header as a query failure, matching platform/fabricx/core/ledger and the nil Data and Metadata guards in Block.DataAt and Block.ProcessedTransaction.

Fixes #1874


mbrandenburger (approved):
LGTM

PR #1876, opened 2026-09-22
test(view): increase coverage of platform/view/services/storage/driver/sql/sqlite

### LINKED ISSUE #1865: test(view): increase coverage of platform/view/services/storage/driver/sql/sqlite
### Description

Coverage for `platform/view/services/storage/driver/sql/sqlite` sits at 38.4% with `scripts/filter-coverage.sh` applied. The existing tests build stores by hand and pass them to the shared suites in `sql/common`, so the package's public entry points the driver, its store constructors and the config path are never exercised.

**Uncovered**

|---|---|---|
| `NewDriver`, `NewNamedDriver`, `NewDriverWithDbProvider` | `driver.go` | 0% |
| `NewKVS`, `NewBinding`, `NewSignerInfo`, `NewAuditInfo` | `driver.go` | 0% |
| `NewPersistenceWithOpts` | `driver.go` | 0% |
| `NewConfigProvider`, `GetOpts` | `config.go` | 0% |
| `NewSignerInfoStore`, `buildSignerInfoStore` | `signerinfo.go` | 0% |
| `NewAuditInfoStore`, `buildAuditInfoStore` | `auditinfo.go` | 0% |
| `NewBindingStore` | `binding.go` | 0% |
| `NewDbProvider`, `key` | `persistence.go` | 0% |
| `NewRetryWriteDB`, `Exec`, `ExecContext` | `retrywritedb.go` | 0% |
| `NewSanitizer`, `Encode`, `Decode` | `sanitizer.go` | 0% |
| `Map` | `isolationlevels.go` | 0% |
| `WrapError` | `errormapper.go` | 62.5% |
| `sqlOpen`, `openDB`, `getDir`, `open` | `persistence.go` | 67–89% |

The driver uses `modernc.org/sqlite`, which runs in-process, so every store can be opened against `t.TempDir()` with no external database. All four stores the driver returns go through `NewPersistenceWithOpts`, so building a `Driver` with `NewDriver` and requesting each store covers the config path, the DB provider and the constructors together. `MockConfig` in `storage/driver/common/testing` supplies the configuration.

The KV and binding stores have shared behavioural suites in `sql/common` - `TestCases`, `TestPutBindingsMultipleEphemeralsCommon` and `TestManyManyPutBindingsCommon` which can run against the stores the driver returns. Signer-info and audit-info have no shared suite and need a direct round-trip.

### Implementation Steps

- [ ] Build a `Driver` with `NewDriver` against a temp-dir config, and run the existing KV and binding suites on the stores it returns.
- [ ] Round-trip a value through the signer-info and audit-info stores obtained from the driver.
- [ ] Cover `WrapError`'s remaining mappings by triggering the underlying SQLite errors, e.g. a unique-constraint violation.
- [ ] Cover `Map`, the sanitizer and `RetryWriteDB`.
- [ ] Close the remaining branches in `persistence.go`. Any that depend on `sql.Open` failing which the in-process driver doesn't do on a valid path are named in the PR rather than left silent.

### Additional Information

- **Target package:** `platform/view/services/storage/driver/sql/sqlite`
- **Parent:** #1566
- **Note:** 38.4% is the filtered figure, after `scripts/filter-coverage.sh`.

sid200727:
Fixes #1865

Coverage for `platform/view/services/storage/driver/sql/sqlite` goes from 38.4% to 91.2% (filtered, via `scripts/filter-coverage.sh`).

The existing tests build stores by hand and pass them to the shared suites in `sql/common`, so the package's entry points were never exercised. These tests go through the front door instead:

-> **`driver_test.go`** - builds a `Driver` with `NewDriver` against a temp-dir database and runs the shared key-value and binding suites on the stores it returns; checks `NewNamedDriver` and that a missing data source is reported.
-> **`signerinfo_test.go`, `auditinfo_test.go`** - round-trip through driver-built stores (no shared suite exists for these).
-> **`config_test.go`** - `GetOpts` defaults, explicit idle settings, and both error paths.
-> **`errormapper_test.go`** - `WrapError` on a non-sqlite error, a unique violation, and an unmapped code.
-> **`isolationlevels_test.go`, `sanitizer_test.go`, `retrywritedb_test.go`, `persistence_test.go`** - the remaining small pieces, including `open` with pragmas skipped.

The config is supplied by `MockConfig`, the same way the Postgres tests do it. The key-value suite builds a fresh driver per store, again mirroring Postgres, because the DB provider caches the connection by data source and each case closes its store. The in-memory driver's tests do call `sqlite.NewDbProvider()`, but coverage is counted per package, so that use didn't register here.

**Left uncovered**
- `RetryWriteDB`'s `SQLITE_BUSY` retry: it needs two connections contending with no busy timeout, which would be flaky in CI.
- Open-time failures a valid temp-dir database doesn't produce: `sql.Open` (only fails for an unregistered driver), the pragma `Exec`, and the write-side open after the read side succeeded.

**Found along the way** — happy to raise separately if preferred:
- `isolationLevelMapper.Map` formats the zero value from the failed lookup, not the requested level, so every unsupported level is reported as `isolation level [0]`. The test asserts the error without asserting the number.

Plain testify, no Ginkgo.


platform/view/services/storage/driver/sql/sqlite/isolationlevels_test.go
```diff
@@ -0,0 +1,34 @@
...
+	il, err := IsolationLevels.Map(cdriver.LevelDefault)
+	require.NoError(t, err)
+	assert.Equal(t, sql.LevelDefault, il)
+
+	il, err = IsolationLevels.Map(cdriver.LevelSerializable)
+	require.NoError(t, err)
+	assert.Equal(t, sql.LevelSerializable, il)
+
+	_, err = IsolationLevels.Map(cdriver.IsolationLevel(99))
+	require.Error(t, err)
```
mbrandenburger, line 33: I think we should indeed fix the error message and return to correct level and test here accordingly! Good catch

platform/view/services/storage/driver/sql/sqlite/config_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Parallel()
+
+	conns := 7
+	idle := 3 * time.Second
+	cp := NewConfigProvider(testing2.MockConfig(Config{DataSource: "file:test", MaxIdleConns: &conns, MaxIdleTime: &idle}))
+
+	o, err := cp.GetOpts("")
+	require.NoError(t, err)
+	assert.Equal(t, 7, *o.MaxIdleConns)
+	assert.Equal(t, 3*time.Second, *o.MaxIdleTime)
```
mbrandenburger, line 53: should we first `require.NotNil(t, o.MaxIdleConns)` and `require.NotNil(t, o. MaxIdleTime)` before asserting the actual value? WDYT?

platform/view/services/storage/driver/sql/sqlite/driver_test.go
```diff
@@ -0,0 +1,106 @@
...
+// the config path, the DB provider and NewPersistenceWithOpts are exercised end to end.
+func TestDriverKVS(t *testing.T) {
+	t.Parallel()
+
+	ds := tempDataSource(t, "kvs")
+	o := Opts{DataSource: ds}
+
+	common.TestCases(t, func(string) (driver.KeyValueStore, error) {
+		// A fresh driver per store, as the Postgres test does: the DB provider caches the
+		// connection by data source, and each case closes its store when done.
```
mbrandenburger: I am not sure if this comment is 100% correct - do we really use caching in these test cases?

platform/view/services/storage/driver/sql/sqlite/driver_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common/mock"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/common"
+)
+
+// tempDataSource returns a data source for a fresh database file under t's temp dir.
+func tempDataSource(t *testing.T, name string) string {
+	t.Helper()
+
+	return fmt.Sprintf("file:%s.sqlite?_pragma=busy_timeout(1000)", path.Join(t.TempDir(), name))
+}
```
mbrandenburger, line 28: I believe some existing tests in the same package have this as inline code, for example, `sql_test.go`. Would it make sense to adopt this helper function here to be consistent?

mbrandenburger (changes requested):
Thank you @sid200727 for working on this. Please have a look at my comments below.

platform/view/services/storage/driver/sql/sqlite/isolationlevels_test.go
sid200727: Fixed, it now reports level rather than the zero value, and the test asserts the number. 
thank you !

platform/view/services/storage/driver/sql/sqlite/config_test.go
sid200727: Added for both

platform/view/services/storage/driver/sql/sqlite/driver_test.go
sid200727: You're right, that overstated it. Reworded to just say each case gets its own connection since the suite closes every store it's given.

mbrandenburger (approved):

PR #1879, opened 2026-09-23
fix(view): report a missing endpoint resolver instead of dereferencing it

### LINKED ISSUE #1878: fix(view): report a missing endpoint resolver instead of dereferencing it
### Description

`Context.newSession` at `platform/view/services/view/context.go:417` calls `resolver.GetName()` on the resolver returned by `c.resolver.Resolver(...)` without checking it for nil. Its sibling `newSessionByID` (`context.go:430`) performs exactly that check before using the same value, so the two paths disagree.

`c.resolver` is the `EndpointService` interface declared at `context.go:40`, not a concrete type, so an implementation is free to return a nil resolver together with a nil error — which is what the `newSessionByID` guard already anticipates. The in-tree implementation, `endpoint.Service.Resolver`, does not: it propagates `lookupResolver`, whose only nil return carries `ErrNotFound`. The exposure is through the interface, including the generated `mock.EndpointService`.

A third call site shares the flaw: `Context.resolve` at `context.go:468` returns `resolver.GetId()` with no nil check. `GetSession` reaches it through the lazy identity iterator at `context.go:306` while looking for an existing session, so it panics *before* `newSession` is ever called. Fixing only the line named in #1861 would leave the earlier crash in place.

Confirmed with a nil-returning `mock.EndpointService`: `GetSession` panics with `invalid memory address or nil pointer dereference` at `context.go:468`, inside `(*endpoint.Resolver).GetId`.

Target state: both unguarded call sites report a missing resolver as an error, matching `newSessionByID`.

SaidAltury-ibm:
### Description
newSession called resolver.GetName() and resolve returned resolver.GetId() on the value returned by EndpointService.Resolver without checking it for nil, while the sibling newSessionByID guards that same value. Since EndpointService is an interface, an implementation may return a nil resolver alongside a nil error, which is what the existing guard anticipates.

GetSession consults resolve through its identity iterator while looking for an existing session, so that call site panics before newSession is reached. Guard both: newSession reports the party it could not resolve, and resolve wraps ErrInvalidIdentity, matching the IsNone case above it.

Fixes #1878


platform/view/services/view/context.go
```diff
@@ -414,6 +414,9 @@ func (c *Context) newSession(view view.View, contextID string, party view.Identi
 	if err != nil {
 		return nil, err
 	}
+	if resolver == nil {
+		return nil, errors.Errorf("no endpoint resolver found for party [%s]", party)
+	}
```
mbrandenburger, line 419: I believe this is the right approach here. Note that `newSessionByID ` deals with `resolver == nil` still differently. Shouldn't it deal with that case in the same way and return an error instead of creating a new session with an empty endpoint address `ep`. WDYT?

mbrandenburger (commented):
Thank you @SaidAltury-ibm. Please see my comment below.

mbrandenburger (approved):

PR #1881, opened 2026-09-23
fix(state): bound-check the stream At methods

### LINKED ISSUE #1880: fix(state): bound-check the stream At methods
### Description

The three `At` methods in `platform/fabric/services/state/stream.go` index their backing slice with no bounds check: `outputStream.At` (line 164), `inputStream.At` (line 217) and `commandStream.At` (line 252). Each body is a bare `return o.outputs[index]`, so an out-of-range position panics with `index out of range`.

These are reachable, not theoretical. Streams are built from transaction content, and `Filter` returns an empty stream when nothing matches, so the common `Filter(...).At(0)` idiom panics on no match. The integration views use exactly that shape against data received from a remote party — `integration/fabric/atsa/views/transfer.go:65` calls `tx2.Inputs().Filter(...).At(0)` on the transaction returned by `SendAndReceiveTransaction`, and `tx.Commands().At(0)` appears in the iou, atsa and fabricx views without a preceding `Count()` check.

All three methods return a pointer, so an out-of-range position can report itself as nil, the same way `Block.DataAt` in `platform/fabric/core/generic/ledger/ledger.go:128` already does for malformed remote blocks.

Returning nil alone only moves the panic, because callers chain straight into the element's methods: `output.State` dereferences `o.namespace`, so `At(99).State(x)` would nil-panic one frame later with a less obvious message. The layer below already reports a bad index as a wrapped error (`Namespace.GetOutputAt`, `namespace.go:258`), so the element methods that return an error should too.

Target state: the three `At` methods return nil for an out-of-range position, and the error-returning element methods report a nil receiver instead of dereferencing it.

`output.ID`, `output.IsDelete` and `input.ID` return a bare value with no error to carry, so they keep dereferencing their receiver: returning a zero `ID` for a nil element would hide the mistake at the call site rather than report it. `At` documents that callers check its result, or the position against `Count`, before calling those.

SaidAltury-ibm:
outputStream.At, inputStream.At and commandStream.At indexed their backing slice with no bounds check, so an out-of-range position panicked. Filter returns an empty stream when nothing matches, and the integration views index the result of a Filter against a transaction received from a remote party, so the position is not always known to be in range.

Return nil out of range, as Block.DataAt already does for a malformed remote block. Since callers chain off the result, output.State, input.State and input.VerifyCertification report a nil receiver as an error, matching Namespace.GetOutputAt, which wraps a bad index.

output.ID, output.IsDelete and input.ID return a bare value with no error to carry and keep dereferencing their receiver: a zero ID would hide the mistake at the call site rather than report it. At documents that, and a test pins it.

Fixes #1880


platform/fabric/services/state/stream.go
```diff
@@ -248,7 +272,11 @@ func (o *commandStream) Count() int {
 	return len(o.commands)
 }
 
-// At returns the command at the passed position
+// At returns the command at the passed position, or nil if it is out of range.
```
mbrandenburger, line 275: This is good stuff. I am wondering if we should also touch the call code in `integration/fabric/iou/views/lender.go` by adding `assert.Equal(1, tx.Commands().Count(), "expected only a single command, got [%d]", tx.Commands().Count())` (or similar) WDYT?

mbrandenburger (changes requested):
Thank you @SaidAltury-ibm for working on this. Could you add the missing unit tests for this?

Moreover, it seems that `platform/fabric/services/endorser/namespace.go` has a similar bounds check problem. Should we fix this in this PR or a follow up?

We could fix it by ... 
```go 
func (k Namespaces) At(i int) string {
      if i < 0 || i >= len(k) {
              return ""
      }
      return k[i]
}
```

mbrandenburger (approved):

